PR #15, opened 2024-09-03
Added ability to view orders paid with removed payment method without breaking

tmewes:
# Description

Maho currently can't handle deleted/removed payment methods in orders in the frontend as well as in the adminhtml.

Example: An external payment module for the payment method `Amazon Pay` had been installed. Some customers used this payment method for their orders in the past. The module was then removed again because a different payment method was to be introduced instead.

A customer who used such removed payment method currrently will get this error when viewing the order in the customer account:
```
There has been an error processing your request

The requested Payment Method is not available.

Trace:
#0 /var/www/html/app/code/core/Mage/Payment/Model/Info.php(101): Mage::throwException('The requested P...')
#1 /var/www/html/app/code/core/Mage/Payment/Helper/Data.php(131): Mage_Payment_Model_Info->getMethodInstance()
#2 /var/www/html/app/code/core/Mage/Sales/Block/Order/Info.php(47): Mage_Payment_Helper_Data->getInfoBlock(Object(Mage_Sales_Model_Order_Payment))
#3 /var/www/html/app/code/core/Mage/Core/Block/Abstract.php(299): Mage_Sales_Block_Order_Info->_prepareLayout()
#4 /var/www/html/app/code/core/Mage/Core/Model/Layout.php(474): Mage_Core_Block_Abstract->setLayout(Object(Mage_Core_Model_Layout))
#5 /var/www/html/app/code/core/Mage/Core/Model/Layout.php(490): Mage_Core_Model_Layout->createBlock('sales/order_inf...', 'sales.order.inf...')
#6 /var/www/html/app/code/core/Mage/Core/Model/Layout.php(234): Mage_Core_Model_Layout->addBlock('sales/order_inf...', 'sales.order.inf...')
#7 /var/www/html/app/code/core/Mage/Core/Model/Layout.php(200): Mage_Core_Model_Layout->_generateBlock(Object(Mage_Core_Model_Layout_Element), Object(Mage_Core_Model_Layout_Element))
#8 /var/www/html/app/code/core/Mage/Core/Model/Layout.php(205): Mage_Core_Model_Layout->generateBlocks(Object(Mage_Core_Model_Layout_Element))
#9 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Action.php(348): Mage_Core_Model_Layout->generateBlocks()
#10 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Action.php(259): Mage_Core_Controller_Varien_Action->generateLayoutBlocks()
#11 /var/www/html/app/code/core/Mage/Sales/Controller/Abstract.php(51): Mage_Core_Controller_Varien_Action->loadLayout()
#12 /var/www/html/app/code/core/Mage/Sales/Controller/Abstract.php(93): Mage_Sales_Controller_Abstract->_viewAction()
#13 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Action.php(424): Mage_Sales_Controller_Abstract->viewAction()
#14 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Router/Standard.php(255): Mage_Core_Controller_Varien_Action->dispatch('view')
#15 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Front.php(181): Mage_Core_Controller_Varien_Router_Standard->match(Object(Mage_Core_Controller_Request_Http))
#16 /var/www/html/app/code/core/Mage/Core/Model/App.php(358): Mage_Core_Controller_Varien_Front->dispatch()
#17 /var/www/html/app/Mage.php(700): Mage_Core_Model_App->run(Array)
#18 /var/www/html/pub/index.php(59): Mage::run('', 'store')
#19 {main}
```

This PR handles removed payment methods in the frontend e.g. on customer account order view pages as well as in the adminhtml e.g. on sales order view pages so they will successfully render again.

# Type of change

Fix

# How can this fix be tested

1. Checkout the main branch (without this fix).
2. Create a customer account.
3. Place an order.
4. View the order in the customer account.
5. The order view page will render successfully and displays, among other things, the payment method selected when ordering correctly.
6. Change the payment method of the order in `sales_flat_order_payment` database table (column `method`) to a non existent payment method (to simulate a removed payment method), e.g. `some_obsolete_payment_method`.
7. Reload the customer account order view page and see the above error message.
8. Checkout the branch of this PR including the fix.
9. Reload the customer account order view page.
10. The page will render successfully again and displays, among other things, at least the payment method code (e.g. `some_obsolete_payment_method`) in the payment information block.
11. A notice is logged so that this case can be tracked if necessary.
12. View the oder in adminhtml.
13. The sales oder view page will render successfully again and displays, among other things, at least the payment method code (e.g. `some_obsolete_payment_method`) in the payment information block.


fballiano:
@tmewes this is beautiful, will do a quick test as soon as I have a break from the language-pack generation work and we'll try to merge it right away.

fballiano:
@tmewes I've changed the "method" on an order but I've 2 problems

## without developer mode

I see this

Screenshot 2024-09-05 alle 23 43 46

## with developer mode

I see this

Screenshot 2024-09-05 alle 23 44 39

weird :-\

tmewes:
Yes, this PR only affects the frontend (which is why I only addressed it in the test steps). I would prepare a separate PR for the adminhtml, as there are a few other problem areas with missing or removed payment methods. I think this can be better tracked and tested with a separate PR, which will follow soon. :)

tmewes:
The code for fixing the frontend is needed for fixing the adminhtml, so we can't merge those two PRs both together.
But no problem, I added fixes for the adminhtml in this PR and adjusted the title, description and test steps in this PR. :)

Could you please test again to see if everything works now?

app/code/core/Mage/Payment/Model/Observer.php
```diff
@@ -32,7 +32,14 @@ public function salesOrderBeforeSave($observer)
...
 
-        if ($order->getPayment() && $order->getPayment()->getMethodInstance()->getCode() != 'free') {
+        if (($payment = $order->getPayment())
+            && ((
+                    Mage::helper('payment')->getMethodModelClassName($payment->getMethod()) !== null
+                    && $payment->getMethodInstance()->getCode() !== 'free'
+                )
+                || ($payment->getMethod() !== 'free')
+            )
+        ) {
```
fballiano, line 42: do we need such a complex if? to me it's extremely difficult to read, but also, in app/code/core/Mage/Adminhtml/Block/Sales/Order/View.php the "if" is so much simpler, shouldn't it work also here? or could we at least make it more readable?

tmewes: Thanks for your feedback! I think I've now found a much better solution that is also easier to read.
This new variant also has the advantage that now even the entire line for an order in `sales_flat_order_payment` can be removed from the database and the pages will still render.
In projects I work with anonymized dev databases (created with [gdpr-dump](https://github.com/Smile-SA/gdpr-dump)) with very few orders, in which the `sales_flat_order_payment` is truncated. With this new variant this is no longer a problem. :)

Could you please give it a try?

app/code/core/Mage/Adminhtml/Block/Sales/Items/Abstract.php
```diff
@@ -449,8 +449,11 @@ public function canEditQty()
         ) {
             return false;
         }
-        if ($this->getOrder()->getPayment()->canCapture()) {
-            return $this->getOrder()->getPayment()->canCapturePartial();
+        if (($payment = $this->getOrder()->getPayment())
```
fballiano, line 452:
```suggestion
        if ($payment = $this->getOrder()->getPayment()
```

I think this would be the same but in general I kinda dont like assignments inside ifs, dunno, it does seem confusing to me

app/code/core/Mage/Sales/Model/Order.php
```diff
@@ -671,6 +671,9 @@ public function canCancel()
      */
     public function canVoidPayment()
     {
+        if ($this->getPayment() === false) {
```
fballiano, line 674: does this need to be this strict?

app/code/core/Mage/Payment/Helper/Data.php
```diff
@@ -120,9 +120,20 @@ public function getMethodFormBlock(Mage_Payment_Model_Method_Abstract $method)
      *
      * @return  Mage_Core_Block_Template|Mage_Core_Block_Abstract
      */
-    public function getInfoBlock(Mage_Payment_Model_Info $info)
+    public function getInfoBlock(Mage_Payment_Model_Info|false $info)
     {
-        $blockType = $info->getMethodInstance()->getInfoBlockType();
+        $method = $info !== false ? $info->getMethod() : $this->__('No Data Found');
```
fballiano, line 125:
```suggestion
        $method = $info !== false ? $info->getMethod() : $this->__('No Data Found');
```

same here, I usually prefer not to use "negative ifs" (like $info !== false) because they're harder to read

also, "No Data Found" would be a new string to add to translation and I'd avoid that, using "N/A" (which is already in translation) or just a "-" or something like that

app/code/core/Mage/Sales/Model/Order.php
```diff
@@ -841,7 +844,10 @@ public function canEdit()
             return false;
         }
 
-        if (!$this->getPayment()->getMethodInstance()->canEdit()) {
+        if (!($payment = $this->getPayment())
+            || Mage::helper('payment')->getMethodModelClassName($payment->getMethod()) === null
+            || !$payment->getMethodInstance()->canEdit()
+        ) {
```
fballiano, line 850: here too i'd prefer multiple simpler ifs instead of a big difficult to read one

fballiano:
sorry for the notes @tmewes :-(((

app/code/core/Mage/Payment/Helper/Data.php
fballiano: also, where is this showing? I can't find it in frontend of backend

app/code/core/Mage/Adminhtml/Block/Sales/Items/Abstract.php
tmewes: Unfortunately, that's not the same. As it was before, the brackets were necessary because of the `&&` chain. But that's not a problem. I've removed the declaration out of the `if`. :)

app/code/core/Mage/Sales/Model/Order.php
tmewes: In my opinion, yes. If you remove the database table row in `sales_flat_order_payment` for an order or truncate the table completely (like e.g. in dev dumps), `$payment` is false. I always check for type equality when possible. If you would like it different, please let me know. But I'm pretty sure about this change here.

app/code/core/Mage/Sales/Model/Order.php
tmewes: No problem. I simplified this part so the code should be easier to read. :)

app/code/core/Mage/Payment/Helper/Data.php
tmewes: The condition is swapped now, so that the `!==` is no longer necessary.

I have specifically chosen a string here that might fit best and already has a translation (see `app/locale/en_US/Mage_Adminhtml.csv:659` for `No Data Found`). When I should use different string, please let me know (and which one exactly). It's not that important to me which one is used here.

To see the string, just place an order, go to `sales_flat_order_payment` table in your database and delete the row for your order. Then you will see it in the frontend as well as in the adminhtml.

tmewes:
Everything is fine, I've adjusted the mentioned parts again and added explanations/answers to your questions from your review. :)

app/code/core/Mage/Payment/Helper/Data.php
fballiano: my fault, sorry! I somehow searched for "no data found" and didn't find it in the translations csv :-\ weird

fballiano:
Retested today, re-read and everything looks ok to me, thank you! ❤️

PR #16, opened 2024-09-05
API: Adjusted WSDL attributes for Maho branding

tmewes:
# Description

This PR adjusts the `name`, `namespace`, `targetNamespace` and `xmlns:typens` attributes in WSDL nodes as well as the `name` for the default WSDL template variables.

# Type of change

Refactor

# How can this fix be tested

1. `find . -type f -iname '*.xml' | xargs -I '{}' xmllint --noout '{}'` should still work.
2. API should still work.

### DIFF
--- a/app/code/core/Mage/Api/Model/Wsdl/Config/Base.php
+++ b/app/code/core/Mage/Api/Model/Wsdl/Config/Base.php
@@ -44,7 +44,7 @@ class Mage_Api_Model_Wsdl_Config_Base extends Varien_Simplexml_Config
         // set up default WSDL template variables
         $this->_wsdlVariables = new Varien_Object(
             [
-                'name' => 'OpenMage',
+                'name' => 'Maho',
                 'url'  => Mage::helper('api')->getServiceUrl('*/*/*', ['_query' => $queryParams], true)
             ]
         );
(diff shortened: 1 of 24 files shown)


fballiano:
I'm puzzled, do we want to go on that route again? What do you think?

tmewes:
Well, we don't have to do that. It was just a suggestion. This change should be listed in the requested migration guides/scripts for M1 and OM later if it will be merged. There are no major changes to the API yet, but that could still happen. Development here is pretty rapid, which I think is great. :)

I just remembered that OM had these changes from M1. With this PR I mainly wanted to point out that this might not be forgotten and also to ensure consistent branding (because in my opinion we are also technically no longer M1/OM at this point due to the many changes to finally strive for a more modern code base). At the end of the day, it is not extremely important to me personally that we do it this way here.

You can also close the PR if you say that this is not the right approach (at the moment). That would be perfectly fine too. :)

fballiano:
no no, I like this PR, since I was the one that did it in OM 😅
I just wanted to get more opinions since that change created a few problems (for us is not the same but it's maybe important to reflect on it for a sec) :-)

tmewes:
I've been thinking about it for a while, but at the end of the day it's just an additional note in the migration guide (or script, as soon as we have one) from M1 or OM. And the change is quick with this knowledge. So at this moment I don't have a really good reason why we shouldn't do it. :)

fballiano:
true, I agree, at this point in time we can do these kind of changes without too much trouble so it's actually better to do them all at once :-)

PR #17, opened 2024-09-05
Adjust Branding in adminhtml.xmls and system.xmls

tmewes:
# Description

This PR adjusts the `Maho` branding in adminhtml `System > Configuration`.

# Type of change

Refactor

# How can this fix be tested

1. Branding should be fine in adminhtml `System > Configuration`.


fballiano:
beautiful, thank you @tmewes!

PR #22, opened 2024-09-17
Fixed merging of javascript files since the new folder structure

tmewes:
# Description

JavaScript merging is currently broken because of the missing `public` directory in the path while collecting the JavaScript files.

Examples of JavaScript errors when visiting the home page:
grafik

# Type of change

Fix

# How can this fix be tested

1. Checkout `main` branch.
2. Visit the adminhtml.
3. Go to `System > Configuration > Advanced > Developer`
4. Set `JavaScript Settings > Merge JavaScript Files` to `Yes` and save.
5. Visit any frontend page, e.g. the home page.
6. You will see errors in the console like in the image shown above.
7. Checkout the branch of this PR.
8. Reload the frontend page.
9. These JavaScript errors are gone.


fballiano:
great, thanks

although I have to say I want to remove the merge feature completely, it always created a mess but it's now unnecessary since http2 and http3, it's probably even making performance worse instead of better :-)

PR #23, opened 2024-09-18
Made MAHO_ROOT_DIR constant available globally

tmewes:
# Description

In the Maho CLI (`maho:12`) `MAHO_ROOT_DIR` is already in use. This PR makes `MAHO_ROOT_DIR` globally available outside the Maho CLI.

The uses of `MAGENTO_ROOT` have been replaced by `MAHO_ROOT_DIR`.

Unfortunately, we can't simply use `__DIR__` in `index.php` and `api.php` (like it has been solved in the Maho CLI), because these files are located in `public`. Therefore, there has been no change in behavior compared to `MAGENTO_ROOT` for the time being.

`MAGENTO_ROOT` has not been removed for compatibility reasons with custom code. However, a `@deprecated` note has been added.

# Type of change

* Feature
* Refactor

# How can this be tested

1. You can use `MAHO_ROOT_DIR` instead of `MAGENTO_ROOT` now.
2. Nothing else should have changed or could break.

### DIFF
--- a/app/code/core/Mage/Core/functions.php
+++ b/app/code/core/Mage/Core/functions.php
@@ -365,8 +365,8 @@ function mahoFindFileInIncludePath(string $relativePath): string|false
 function mahoListDirectories($path)
 {
     list($packages, $packageDirectories) = mahoGetComposerInstallationData();
-    if (!defined('MAGENTO_ROOT')) {
-        Mage::throwException('MAGENTO_ROOT constant is not defined.');
+    if (!defined('MAHO_ROOT_DIR')) {
+        Mage::throwException('MAHO_ROOT_DIR constant is not defined.');
     }
 
     foreach ($packages as $package) {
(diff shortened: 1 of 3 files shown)


fballiano:
@tmewes with this PR MAHO_ROOT_DIR is both the root of the project (for maho cli) and the public folder (for the php scripts in public), right? mmm I don't know if I love this double meaning

fballiano:
I would have liked

```php
define('MAHO_ROOT_DIR', dirname(__DIR__));
/** @deprecated Use MAHO_ROOT_DIR instead. */
define('MAGENTO_ROOT', MAHO_ROOT_DIR);
```

better, but it looks weird right?

your implementation is totally fine to me

tmewes:
Thank you for your review and feedback! I thought about that, but I think the current state is a bit easier to get/understand, too.

fballiano:
yeah it is

PR #25, opened 2024-09-19
GUI to Edit EAV Attributes & Sets

justinbeaty:
Cherry-picked commits from https://github.com/OpenMage/magento-lts/pull/2317, https://github.com/OpenMage/magento-lts/pull/2352, and https://github.com/OpenMage/magento-lts/pull/2353.

A couple of notes:

- The files `js/varien/js.js`, `skin/frontend/base/default/js/opcheckout.js`, and `skin/frontend/rwd/default/scss/module/_eav.scss` had changes in https://github.com/OpenMage/magento-lts/pull/2352, but I'm not sure where those changes go in Maho.

- The last two WIP commits were part of an experimental feature to automatically add fields into various customer forms in the frontend and backend. I had an experimental branch that did more work here, but I think I may have lost it.

- I didn't pull any of the short array syntax, phpcs, or merge commits. Instead I rebased on top of Maho/main.


fballiano:
@justinbeaty first of all thank you for this PR, for sure questions will arise, first one being...

Do you think that customer attributes should have this kind of scope?
Screenshot 2024-09-19 alle 18 53 01

justinbeaty:
> @justinbeaty first of all thank you for this PR, for sure questions will arise, first one being...
> 
> Do you think that customer attributes should have this kind of scope?

My initial thought is that it should support scopes. An example might be a wholesaler website that asks for tax info. Similarly an EU store view might ask for a VAT id, but a USA one would not.

We also have the "Share Customer Accounts" option in the backend that might influence the decision, but I am positive that would cause any issues.

image

I have to admit I have very limited experience in running more than one Website / Store / Store View.

justinbeaty:
> 2\. If I "open customer attributes", filter something on the grid, then go to "customer address attributes" the grid is already filtered (I think the grids share the same ID or it's because they're actually the same class), is it just a matter of ID or should we implement a custom class for each one of them (instead of the observer setting the properties)?

I am pretty sure this is because of the same ID. What if you try to change this line:

https://github.com/MahoCommerce/maho/blob/01a5dbb56e6cf5151f7edb1180f972bc4f8807cd/app/code/core/Mage/Eav/Block/Adminhtml/Attribute/Grid/Abstract.php#L24

to:

```php
$this->setId('attributeGrid' . (Mage::registry('entity_type') ?? ''));
```

fballiano:
> > 2. If I "open customer attributes", filter something on the grid, then go to "customer address attributes" the grid is 
> ```
> $this->setId('attributeGrid' . (Mage::registry('entity_type') ?? ''));
> ```

done! I changed a bit just because Mage::registry('entity_type') is an object but thanks! works as expected.

app/code/core/Mage/Eav/Block/Adminhtml/Attribute/Set/Grid.php
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * @category   Mage
+ * @package    Mage_Eav
+ */
+class Mage_Eav_Block_Adminhtml_Attribute_Set_Grid extends Mage_Adminhtml_Block_Widget_Grid
+{
+    public function __construct()
+    {
+        parent::__construct();
+        $this->setId('setGrid');
```
justinbeaty, line 21: We might have the same issue here for the attribute set grids.

fballiano: true, fixed these ones too!

justinbeaty:
> > @justinbeaty first of all thank you for this PR, for sure questions will arise, first one being...
> > Do you think that customer attributes should have this kind of scope?
> 
> My initial thought is that it should support scopes. An example might be a wholesaler website that asks for tax info. Similarly an EU store view might ask for a VAT id, but a USA one would not.
> 
> We also have the "Share Customer Accounts" option in the backend that might influence the decision, but I am positive that would cause any issues.
> 
> image
> 
> I have to admit I have very limited experience in running more than one Website / Store / Store View.

I just remembered that in addition to (or instead of) supporting scopes, we have the ability to set a customer attribute set per customer group.

justinbeaty:
Looking into the scope issue again, and it turns out `is_global` is actually a column on the table `catalog_eav_attribute` and not on the base table `eav_attribute`, nor `customer_eav_attribute`. So I'll actually remove the scope dropdown from the base EAV code.

However, Magento actually has something more complicated for customer attributes with the table `customer_eav_attribute_website` which is a website specific override for values in `customer_eav_attribute`.

This is actually already functional by setting various values in System Config -> Customer Configuration -> Name and Address Options. For example, select "Default Config" and set Show Prefix to Required and it's saved in `customer_eav_attribute`, but change to "Main Website" and it is saved in `customer_eav_attribute_website`. I knew the former, but never noticed the latter.

I'm still investigating the best way to wrangle this all...

justinbeaty:
Test comment 2

fballiano:
Tested the category attributes/set
-> it's amazing, it would have been needed 10 years ago hehehehe
-> attribute sets are not really useful for normal people (only the attributes in the default set are manageable via backend, correctly) but it could still be very powerful if used via programming. @justinbeaty do you think there could be a way of switching attribute sets in the category management page? would it be a GUI/UX disaster to do so? IMHO it's not really necessary

Should we add more "return types" instead of docblock? At least where it won't break the inheritance?

Should we add default labels (with install scripts) to attributes that don't have one? This looks kinda weird:
Screenshot 2024-09-28 alle 11 26 28

Regarding customer attributes:
- are there some that we should forbid to edit in any way? Like "default billing address"? I see that it's not possible to delete them and it's not super necessary to forbid them to be edited, it's just a thought
- it seems all the functionality work correctly, except for the "use in form" right? which works for default attributes but not for custom ones?

justinbeaty:
> it seems all the functionality work correctly, except for the "use in form" right? which works for default attributes but not for custom ones?

I accidentally broke this with the commit yesterday; just pushed a fix. It does work, see screenshots below. This actually has *a lot* more potential, but is also where I got stuck two years ago trying to generalize it too much. My idea was to use the customer attribute sets to be groups of fields in the forms, and to remove the hardcoded fields in `template/customer/form/edit.phtml`. However, let's leave that possibility for another PR.

Will respond to your other points in next message...

https://github.com/MahoCommerce/maho/blob/45572a0ac62c837ae90021554a9f85cb2f0fa7bf/app/design/frontend/rwd/default/template/customer/form/edit.phtml#L23-L51

image
image
image

justinbeaty:
> are there some that we should forbid to edit in any way? Like "default billing address"? I see that it's not possible to delete them and it's not super necessary to forbid them to be edited, it's just a thought

> Should we add default labels (with install scripts) to attributes that don't have one? This looks kinda weird:

I think we can hide some of these fields, such as default billing address, and all of those ones that have no label. I actually found out yesterday Magento has a way to disable certain fields when editing an attribute. The first example below is from core Magento, and the second example is added in this PR to make it so you cannot change the required setting of middlename, since it's supposed to always be optional. I think we can expand this and make a `<editable>0</editable>` node which would hide it from the grid.

https://github.com/MahoCommerce/maho/blob/45572a0ac62c837ae90021554a9f85cb2f0fa7bf/app/code/core/Mage/Catalog/etc/config.xml#L605-L637
https://github.com/MahoCommerce/maho/blob/45572a0ac62c837ae90021554a9f85cb2f0fa7bf/app/code/core/Mage/Customer/etc/config.xml#L449-L466

justinbeaty:
> I've added this customer attribute to the customer create form
> then also added it to the default attribute set
> then cleaned the cache but it doesn't show: 

 Hm, it seems to be loading template `persistent/customer/form/register.phtml`, which I didn't modify to include the extra fields (yet). But I have persistent cart disabled. Is there a bug in `ifconfig`?
```xml
<action method="setTemplate" ifconfig="persistent/options/enabled"><template>persistent/customer/form/register.phtml</template></action>
```

app/code/core/Mage/Eav/Model/Config/Source/Form.php
```diff
@@ -0,0 +1,49 @@
...
+ * @package    Mage_Eav
+ */
+class Mage_Eav_Model_Config_Source_Form
+{
+    protected const XML_PATH_EAV_FORMS = 'global/eav_forms';
+
+    public function toOptionArray(string $entity = ''): array
+    {
+        $node = Mage::getConfig()->getNode(self::XML_PATH_EAV_FORMS . '/' . $entity);
+        if ($node === false) {
```
fballiano, line 25: can $node be false? it seems ->getNode() never returns false

justinbeaty: Yes, if an empty string is passed for `$entity`:
```php
var_dump(Mage::getConfig()->getNode('global/eav_forms/')); 
// Outputs: bool(false)
```

fballiano: right, then the docblock for the getNode() is wrong, let's see

justinbeaty:
> attribute sets are not really useful for normal people (only the attributes in the default set are manageable via backend, correctly) but it could still be very powerful if used via programming. @justinbeaty do you think there could be a way of switching attribute sets in the category management page? would it be a GUI/UX disaster to do so? IMHO it's not really necessary

We talked a bit about this here: https://github.com/OpenMage/magento-lts/pull/2317#issuecomment-1193391236 but I think we can leave it for another time...

fballiano:
> We talked a bit about this here: [OpenMage/magento-lts#2317 (comment)](https://github.com/OpenMage/magento-lts/pull/2317#issuecomment-1193391236) but I think we can leave it for another time...

yep, makes total sense

fballiano:
I thought the conflicts were a bit more difficult but they were actually super easy 😂

fballiano:
weird, when I go to the checkout I see this:

Screenshot 2024-10-01 alle 22 41 57

justinbeaty:
> I thought the conflicts were a bit more difficult but they were actually super easy 😂

Yes I knew there would be conflicts. I'll look at the errors tomorrow.

justinbeaty:
The error has nothing to do with the merge, it simply happens when checking out as guest. Will fix work on it later today, have some offline things to do right now.

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tab/Attributes.php
```diff
@@ -161,8 +161,7 @@ protected function _getAdditionalElementTypes()
     {
         return [
             'image' => Mage::getConfig()->getBlockClassName('adminhtml/catalog_category_helper_image'),
-            'textarea' => Mage::getConfig()->getBlockClassName('adminhtml/catalog_helper_form_wysiwyg'),
-            'boolean' => Mage::getConfig()->getBlockClassName('eav/adminhtml_helper_form_boolean'),
```
fballiano: boolean doesn't exist anymore?

justinbeaty: I moved it to `lib/Varien/Data/Form/Element/Boolean.php`, along with adding a `Number.php` type there too. 

Since they're such basic types, It seemed to make more sense than to add those as additional element types manually. It also seemed to make more sense than creating a new folder `Mage/Eav/Block/Adminhtml/Helper` to put these basic types when they're better suited in the Varien lib.

The line I deleted was added in previously in this PR, so it's not breaking anything.

justinbeaty:
@fballiano What do you think about this? 

Here is the frontend customer account edit form template file from main branch: (click the file if you want to see the whole thing)

https://github.com/MahoCommerce/maho/blob/8152a23d2451bb0c1419ee89fd6ac4f67976702e/app/design/frontend/rwd/default/template/customer/form/edit.phtml#L18-L43

And from last few commits in this PR:

https://github.com/MahoCommerce/maho/blob/ed1ebaade7b0083fb57f6899f29fe956b2430fe4/app/design/frontend/rwd/default/template/customer/form/edit.phtml#L20-L37

Basically, what it does is remove all the hardcoded input fields, and replaced them with the groups as defined in the Manage Customer Attribute Sets. Here's what it looks like on the frontend:

image

What's really powerful about this is that you can define different attribute sets per customer group, you can re-arrange all the fields how you like (notice I put last name first), and make as many custom field groups as you like. That means that you don't have to edit the phtml file if you want to add new fields.

This is the direction I wanted to go in back in the original OM PR, in several different templates like the create new account form, address edit form, and possibly even the onepagecheckout forms.

I think it gives a serious practical use for the customer attribute editor. We could also implement that in a separate PR if you think it's too much at once.

fballiano:
OMG this is amazing!!!!!!!!!!! I think using the attribute set group is a great idea and gives enough power to the users to organize the attributes as they want.
can I help you with anything here?

fballiano:
> If you've ever used Magento EE, did they have any functionality like this? Back in 2022 I got caught up on the fact that there's all these generic eav_form tables in the db, and figured they were used somehow, like for a similar form builder feature.

I did use EE multiple times but it was years ago and I don't remember very well about this specific part, most of the EE modules were not super well made (and there were not maintained in the last years)

> In the install script, I think it would be good to make a new group called "Internal" or "Hidden" and move a bunch of the attributes there. And to make sure that the ones in General are in the correct order.

is there the possibility to hide a single attribute somehow?

justinbeaty:
> is there the possibility to hide a single attribute somehow?

Do you mean just hide from the Groups drag and drop tree?

On the one hand: if the end user defines their own internal attribute that they need to assign to a set, they'll need to create an Internal / Hidden group anyway.

On the other hand: We also don't want end users removing things like password_hash from the set. I'm actually not sure what would happen there.

fballiano:
> Do you mean just hide from the Groups drag and drop tree?

my first thought from within the attribute edit page

> On the one hand: if the end user defines their own internal attribute that they need to assign to a set, they'll need to create an Internal / Hidden group anyway.

if those attributes are hidden from the frontend, do we need to move it to a different attribute-set?

> On the other hand: We also don't want end users removing things like password_hash from the set. I'm actually not sure what would happen there. But they already have the possibility to do the same with Product attribute sets.

right... well some of them are already not-deletable right? so I guess they should be non-editable-movable or something like that?

justinbeaty:
> > Do you mean just hide from the Groups drag and drop tree?
> 
> my first thought from within the attribute edit page

Ah right, we already did that here: https://github.com/MahoCommerce/maho/pull/25/commits/46cb59b667cc6f6a6aafce6ff5c6619c76f0eb28, so maybe we can use the same process and not have to modify the DB at all.

justinbeaty:
I apologize if you already started working on this, but it seems I was able to implement it in just two lines of code. Bouncing the idea off you helped tons:

image

justinbeaty:
One other thing you could help with:

Is flatpickr loaded on the frontend? It would be really nice to use it instead of this ugly input type:

image

I think you would need to modify these two files:

app/code/core/Mage/Eav/Block/Widget/Date.php

app/design/frontend/base/default/template/eav/widget/date.phtml

fballiano:
> I apologize if you already started working on this, but it seems I was able to implement it in just two lines of code. Bouncing the idea off you helped tons

perfect!!

> Is flatpickr loaded on the frontend? It would be really nice to use it instead of this ugly input type:

no it's only in backend and it would be great to have it on frontend but I think that would be better for a separate PR because:
- it should load only if necessary
- which date format to use?

justinbeaty:
> no it's only in backend and it would be great to have it on frontend but I think that would be better for a separate PR because:

Yeah, good points. We can revisit later.

justinbeaty:
I added a new EAV attribute input type, "Dropdown with Other Option". It stores the values in the table `eav_entity_varchar`, but allows you to define preset options like a select / multiselect. The user can either select from the preset options, or enter their own. I force pushed the changes so you can review it all in one commit: https://github.com/MahoCommerce/maho/pull/25/commits/d66d790ba82a7a0616360b4fef5772c1276290d6

Attribute edit / create form:

image
image

Admin customer edit form:

image

Frontend customer edit form:

image

---

The reason I did this was because I was thinking about the customer prefix, suffix, and gender inputs with the new customizable account edit form. The prefix and suffix fields are a bit odd, because they're text inputs, but Magento allows you to set a list of options in System > Configuration > Customer Configuration and then they behave like selects. I would have to hardcode those field input to use a different renderer, which isn't super great. Also allowing custom input while still suggesting some presets seems better for these fields.

---

If this is all getting to be too much on one PR, I could split things up:

1) EAV GUI code, everything in this PR up until a few days ago.
2) Add the "Dropdown with Other Option" input type.
3) Make the frontend customer forms customizable.

@fballiano what do you think?

fballiano:
I think this is an amazing new feature!

@justinbeaty the only thing we miss is patching the rest of the phtmls for attribute sets right?

justinbeaty:
> I think this is an amazing new feature!
> 
> @justinbeaty the only thing we miss is patching the rest of the phtmls for attribute sets right?

A few additional things:

1) Improve the syncing between System > Configuration > Customer Configuration and the EAV attributes.
    * Magento already does this one way, I.e. if you set "Show prefix" in the config it will actually set the prefix attribute (both customer and customer_address) to set is_visible and is_required.
    * It doesn't work the other way though, if you change it in the EAV editor, it doesn't update core_config_data. (I have TODOs marked in the code where this should happen.)
    * Really that whole config section *could* be removed, as eveything is controllable by EAV attributes now, but I think better to leave it.
2) There's a bug where if an attribute is required, but not visible in the form you get an error because the field is blank.
3) Some install script stuff. Especially if we decide to move prefix, suffix, and gender to the new input type, we need to modify those attributes in the DB.

What date in November do you plan to release 24.11.0? I have a couple of short trips coming up and would like to still land this in that release.

fballiano:
>    * Magento already does this one way, I.e. if you set "Show prefix" in the config it will actually set the prefix attribute (both customer and customer_address) to set is_visible and is_required.
>    * It doesn't work the other way though, if you change it in the EAV editor, it doesn't update core_config_data. (I have TODOs marked in the code where this should happen.)
>    * Really that whole config section _could_ be removed, as eveything is controllable by EAV attributes now, but I think better to leave it.

IMHO we could also remove the config section, this kind of redudancy usually creates more problems that it solves

> 2. There's a bug where if an attribute is required, but not visible in the form you get an error because the field is blank.

I wouldn't even classify this as a bug, if you hide an attribute and make it required then maybe there should be some javascript or other kind of software-based-thing filling the attribute

> 3. Some install script stuff. Especially if we decide to move prefix, suffix, and gender to the new input type, we need to modify those attributes in the DB.

at the moment I'm a bit lost on this specific point but, if possible, I'd probably leave that as it is and decide later if we want

> What date in November do you plan to release 24.11.0? 

I would do around mid november but we don't have any date. for some reason I see many software set a date like 24.11 and then release the last possible day, which is kinda funny to me :-D

:-))

fballiano:
@justinbeaty I wanted to help starting to patch the phtmls, but for customer_address, how should I get the same as customer's getGroupedFields?

justinbeaty:
> @justinbeaty I wanted to help starting to patch the phtmls, but for customer_address, how should I get the same as customer's getGroupedFields?

In app/code/core/Mage/Customer/Block/Address/Edit.php rename the two occurrences of `getExtraFields` to `getGroupedFields` and then in the customer/address/edit.phtml file you can use the code from customer/form/edit.phtml.

Basically, the main `getGroupedFields` method is defined in the helper classes, but each block needs to call it with the correct entity and form code, so that's why I put the method in the block classes too. All of the `getExtraFields` functions will be deleted once we're done.

The address and register templates should be easy, the one page checkout billing will be a bit more tricky.

justinbeaty:
This is pretty funny. I had no idea!

image

fballiano:
Oooops, not only 😉

justinbeaty:
@fballiano Do you think this whole block of code should be deleted? We are already validating these fields with the Mage_Eav_Model_Form class when you save the info from the frontend or backend. I'm not sure if any other 3rd party extensions call the validate method, but if they do and don't rely on the EAV form feature, then they'd be missing out on other customer attributes marked required anyway.

https://github.com/MahoCommerce/maho/blob/1222b6ab31dd94558a84966ec70437bc3127c35b/app/code/core/Mage/Customer/Model/Customer.php#L1075-L1087

fballiano:
mmm I wonder if it would be better to keep the validation in the model...

but at the same time the model should only validate info that's necessary for the model itself, while frontend/backend (form) validations shouldn't be in the model...

so to me it's fine either ways, as you feel it's better

justinbeaty:
This is a possible solution, but it doesn't yet take into account the input validation rules (but neither did the code it replaces.) The values will be validated twice because of the form model, but still it's some validation right in the model.

```php
        $group = Mage::getModel('customer/group')->load($this->getGroupId());

        $attributes = Mage::getResourceModel('customer/attribute_collection')
                        ->setAttributeSetFilter($group->getCustomerAttributeSetId())
                        ->addFieldToFilter('is_visible', 1)
                        ->addFieldToFilter('is_required', 1)
                        ->setNotCodeFilter(['website_id', 'group_id']);

        foreach ($attributes as $attribute) {
            if (trim($this->getData($attribute->getAttributeCode())) === '') {
                $errors[] = Mage::helper('customer')->__('%s is required.', $attribute->getStoreLabel());
            }
        }
```

I'll come back to this later, going to work on the more important items.

justinbeaty:
> > * Magento already does this one way, I.e. if you set "Show prefix" in the config it will actually set the prefix attribute (both customer and customer_address) to set is_visible and is_required.
> > * It doesn't work the other way though, if you change it in the EAV editor, it doesn't update core_config_data. (I have TODOs marked in the code where this should happen.)
> > * Really that whole config section _could_ be removed, as eveything is controllable by EAV attributes now, but I think better to leave it.
> 
> IMHO we could also remove the config section, this kind of redudancy usually creates more problems that it solves

@fballiano One thing I'm not sure about yet is that we have a prefix/suffix/middle name on both customer and customer_address. The old config section would set is_visible, is_required, and the dropdown options on both, but with the editor we have the ability to change each one separately.

Should we somehow sync these between customer and customer_address? Or should users be able to control them separately?

fballiano:
if you want to sync config->eav then I'd sync both customer and customer_address, but to me the config part could be removed

justinbeaty:
> if you want to sync config->eav then I'd sync both customer and customer_address, but to me the config part could be removed

Say we remove it, which I'm fine with, but if you edit customer prefix and add a dropdown option, should that sync to customer_address prefix? Or vice-versa?

I could see the value of making is_required or is_visible separately controlled, but I'm not sure if having to add the dropdown options in both places is a pain for users.

fballiano:
> Say we remove it, which I'm fine with, but if you edit customer prefix and add a dropdown option, should that sync to customer_address prefix? Or vice-versa?

I'd say no, they're 2 separate things and people will control them separately

> I could see the value of making is_required or is_visible separately controlled, but I'm not sure if having to add the dropdown options in both places is a pain for users.

to sync only part of the attribute (IMHO) creates a bit more problems, this things are usually touched when starting the project and never more so I wouldn't think too much about them

justinbeaty:
Hm, okay I removed the old config options here: https://github.com/MahoCommerce/maho/pull/25/commits/cfafda9c9b0dc05f95fd5fe8f27cd8b234351584

Normally I'd try to not change old install scripts, but they're using config options that no longer exist, so it doesn't make sense. I tested from a fresh install, so everything seems to work.

justinbeaty:
Oof PHPStan. I had to add a few lines to the baseline file. We can fix it, but it requires changing a lot of files to add the `$withEmpty` and `$defaultValues` parameters, like here: https://github.com/MahoCommerce/maho/blob/f2029b90c36ac27368bd6d635aa18402ce96aed4/app/code/core/Mage/Eav/Model/Entity/Attribute/Source/Store.php#L24-L25

Will do later in another PR...

justinbeaty:
Okay big update I was working on last few days. I created a `Mage_Eav_Block_Widget_Form` set of classes that make displaying forms super easy. My main motivation was that on a few pages we basically have two forms combined into one phtml. For instance the onepage billing and multishipping register have elements from both the register and address forms.

You can use it something like this:
```php
class Mage_Customer_Block_Form_Edit extends Mage_Customer_Block_Account_Dashboard
{
    /**
     * Create form block for template file
     */
    protected function _beforeToHtml()
    {
        /** @var Mage_Customer_Model_Form $form */
        $form = Mage::getModel('customer/form');
        $form->setFormCode('customer_account_edit')
             ->setEntity($this->getCustomer())
             ->initDefaultValues();

        /** @var Mage_Eav_Block_Widget_Form $block */
        $block = $this->getLayout()->createBlock('eav/widget_form');
        $block->setTranslationHelper($this->helper('customer'));
        $block->setForm($form);

        $groups = array_keys($block->getGroupedAttributes());
        if ($groups[0] === 'General') {
            $block->setDefaultLabel('Account Information');
        }
        $this->setChild('form_customer_account_edit', $block);

        return parent::_beforeToHtml();
    }
}
```

And then in the template file, just call this:

```php
<?= $this->getChildHtml('form_customer_account_edit') ?>
```

It has tons of features like being able to set HTML tags (li, li.wide, div), set default labels, show as a "flat" or "grouped" form, combine or exclude fields from a different forms, and more!

This definitely wasn't just making it more complicated for the sake of it. Those templates where you combine fields was being a pain, and this was the easiest solution. Plus, it feels a lot more extensible.

There is probably some other stuff I am forgetting to mention, but I'm a bit short on time today to explain more. I am still working on the checkout billing step, so that part doesn't fully work yet.

justinbeaty:
I think there is a bug with the PHPStan workflow, specifically the "Restore result cache" part. [This job](https://github.com/MahoCommerce/maho/actions/runs/11583534021/job/32248884284) shouldn't have failed because I hadn't even pulled in [OM PR 4245](https://github.com/OpenMage/magento-lts/pull/4245) yet, but I think it was reading the cached `var/resultCache.php` file from another job.

I think we can probably replicate the bug by making a test PR that deletes a file with an ignoredError.

fballiano:
Hi @justinbeaty, when I go to checkout I get this

Screenshot 2024-11-04 alle 15 15 40

this $groups array is empty
Screenshot 2024-11-04 alle 15 17 11

justinbeaty:
> Hi @justinbeaty, when I go to checkout I get this

I believe I hit that error too, but just haven't gone back to fix it yet. It is for sure just something small.

fballiano:
@justinbeaty I've tried to solve the conflicts, hope I didn't mess things up

justinbeaty:
> @justinbeaty I've tried to solve the conflicts, hope I didn't mess things up

I double check the merge in a bit, still working locally with some commits.

fballiano:
solved minor conflicts

justinbeaty:
Okay, some big updates:

## De-duplicate Code

The generic EAV code was adapted from the product attribute code, but previously the old code was left as-is. I made all of the product attribute blocks extend the generic EAV ones. We went from net +3,624 LOC to net +1,231 LOC, which is absolutely essential for maintainability:

image

I took pretty good care to make sure this change doesn't break things. While the other EAV types (category, customer, address) use observers to change things if needed, it is now also possible to extend the generic block classes for more control. Thus, all of the old `Mage_Adminhtml_Block_Catalog_Product_Attribute_*` classes still exist, they just now extend the `Mage_Eav_Block_Adminhtml_Attribute_*` classes and are minimal or even empty.

One of the main ways this was accomplished was to convert all of the hardcoded `$this->getLayout()->createBlock()` calls into layout XML. This has the nice benefit that for other EAV types you can override just a single block if needed. This was the reason for https://github.com/OpenMage/magento-lts/pull/4337 btw.

## Consolidated logic into EAV helper

Magento had tons of logic spread around probably a dozen files dealing with EAV attribute's backend types, models, display options, etc. It would have been pretty much impossible to define new input types or add logic for other entity types without overriding tons of classes.

Instead, I moved all of this info into the config.xml files, and wrote a bunch of helper methods to return this info. Everything is now super flexible.

## Rewrote JS files

Instead of the inline JS in a few files, I created new files at `public/js/mage/adminhtml/eav/` in vanilla JS, with one exception of using prototype's Template class, which will be easy for us to replace once prototype is finally dropped. Todo: write JSDoc annotations.

## Misc

There's probably some other stuff too, but those are the major changes. It's very possible something broke here as I haven't fully tested, but all can be fixed and this is still a hundred times better for the long-term. One thing to note, you might see few attributes in the manage customer attribute grid. I have to investigate how Magento devs used the `is_visible` column because right now I'm hiding attributes based on that, which is how the product attributes grid works.

justinbeaty:
One more thing, I re-introduced the `setTemplateIfExists` method. The reason being that I deleted most of the templates at `app/design/adminhtml/default/default/template/catalog/product/attribute/`, but it's highly likely someone has made a customization in one of those files. Since these calls are only in the admin panel, I don't think the extra filesystem check is a huge deal. If the re-introduction of this method is okay with you, I can make a separate PR for it because I think I have another idea for an optimization here.

justinbeaty:
Don't worry about conflicts, I'll resolve later...

fballiano:
@justinbeaty how would you like to go about this? the more we keep it in this state the more it will have conflicts :-\ I'd really like to merge it and release it and maybe improve later I don't know

justinbeaty:
> @justinbeaty how would you like to go about this? the more we keep it in this state the more it will have conflicts :-\ I'd really like to merge it and release it and maybe improve later I don't know

Fixing the conflicts now. Then I will open a new PR with smaller changes.

fballiano:
hey @justinbeaty is there any way you see possible to revive this feature?

PR #26, opened 2024-09-19
Replaced CRLF with LF line separators

tmewes:
# Description

When searching for CRLF line separators with `git grep -Il $'\r'` (see https://stackoverflow.com/a/24604608), there is still one file with CRLF line separators left:
`app/code/core/Mage/Usa/etc/wsdl/FedEx/TrackService_v5.wsdl`

This PR replaces the CRLF with LF line separators.

# Type of change

* Fix

# How can this be tested

1. Checkout `main` branch.
2. Run `git grep -Il $'\r'` in the Maho root directory.
3. See `app/code/core/Mage/Usa/etc/wsdl/FedEx/TrackService_v5.wsdl` as result.
4. Checkout the branch of this PR.
5. Run `git grep -Il $'\r'` in the Maho root directory again.
6. There will be no file with CRLF line separators anymore.


fballiano:
nice catch! :-)

I'll wait for the workflows and merge :-)

fballiano:
I think I'd like to have this check in a workflow, to make sure no \r gets ever committed

tmewes:
I would be very much in favour of that. :)

fballiano:
done in https://github.com/MahoCommerce/maho/commit/b6639c9283eade246ed37ae3afe8577233393d41 ;-)

tmewes:
Thank you! :D

PR #31, opened 2024-09-24
Added Maho branded favicons

tmewes:
# Description

This PR adds a favicon in SVG and PNG file format and replaces all old ICO usages.
Sadly some browsers like Safari and some mobile browsers don't support SVG favicons yet (see https://caniuse.com/link-icon-svg), but for those the PNG file is added as a fallback (see https://caniuse.com/link-icon-png).

The methods `getFaviconFile` and `_getFaviconFile` are marked as deprecated, because there are now two files that would need to be maintained individually via settings.
The benefit of setting the favicon in the settings wasn't clear to me beforehand anyway, since (in all cases I know of) in a custom frontend design the `head.phtml` is adjusted/modified as a copy. If necessary, either a different favicon (including fallback) can be set in this template overwrite or the favicon can be changed e.g. via htaccess. Therefore (and to save calls that are not necessary in my opinion) I decided to hardcode the files in all places.

# Type of change

* Fix
* Refactor

# How can this fix be tested

1. Visit the Maho frontend, adminhtml or installer pages.
2. The favicon should be loaded and shown in your browser tab.


fballiano:
Hi @tmewes and thank you for this PR, I've removed the previous favicons because I've always hated that a "composer install" would overwrite my customers' favicon, but I see some problems in this implementation:
- the public/favicon* files would have the same problems as the magento1 management, since they would be overwritten by a composer install (yes, we can modify our composer plugin, but there's another thing to consider)
- multistore projects would likely have multiple favicons and this implementations makes it impossible, this is the reason why these are these skin-based files/methods. and we can't lose the multistore functionality.

am I missing something?

tmewes:
All of this can be implemented using htaccess, caddy, ddev (using `post-start` hooks) or varnish. One of them is enough, of course. :D
None of my local-, test- and production-instances would have any problems with the current implementation and all custom favicons in all stores would be fine, but I understand that it should also be possible using Maho's onboard tools.

But your points only refer to the frontend, right? So the rest of the implementation can stay as it is (adminhtml, installer)? Then I would think of something for the frontend, how it can stay in the settings and the new `favicon.svg` and `favicon.png` are just the fallback.

I haven't yet installed Maho using composer. What exactly do I have to do to reproduce the problem you mentioned? I'm sure we can find a solution for that. :)

fballiano:
> All of this can be implemented using htaccess, caddy, ddev (using `post-start` hooks) or varnish. 

you're right, silly me

> But your points only refer to the frontend, right?

sure, backend IMHO should always have our favicon 

> I haven't yet installed Maho using composer. What exactly do I have to do to reproduce the problem you mentioned? I'm sure we can find a solution for that. :)

it doesn't happen with the current implementation because there's no favicon, let's say that everything in the /public folder gets overwritten every time you do a composer install, normally it's not a problem because you don't need to patch one of the js from the core (just an example) but a favicon for sure. anyway patching our composer plugin to avoid to copy a file if it already exists shouldn't be hard: https://github.com/MahoCommerce/maho-composer-plugin/blob/main/src/MahoComposerPlugin.php

fballiano:
trying to recap, I'd like to have a non-removable favicon for the admin/install, but for that it shouldn't be directly in /public no?

about the frontend, probably your implementation is the right one and we should discuss a little how we can make the composer experience better, because we still need to copy stuff from maho's public to the final-project's public folder.

for final projects (the stores) it should be absolutely avoided to clone the main repo and eveybody should use `composer create-project mahocommerce/maho-starter yourproject` because it's a much cleaner filesystem for the store projects

tmewes:
I've now spent a bit of time testing `composer create-project mahocommerce/maho-starter yourproject`. It worked really well, thank you for creating and pointing it! I'm still very unfamiliar without `git clone` (or having the whole repo incl. core), but I could very well imagine that this will change. :)

I think I understand the problem better now. Then I would suggest that we simply use `$this->getSkinUrl()`. This way we still have a uniform solution everywhere that still doesn't require any setting and in the frontend we can use skin for each store/theme to put custom favicons into if required. The solutions with htaccess, caddy, ddev or Varnish would of course still be possible, but are no longer required.

I just pushed my suggestion. What do you think?

fballiano:
> I've now spent a bit of time testing `composer create-project mahocommerce/maho-starter yourproject`. It worked really well, thank you for creating and pointing to it! I'm still very unfamiliar without `git clone` (or having the whole repo incl. core), but I could very well imagine that this will change. :)

:-) this is almost specifically the base of Maho, having a project for a store which is freaking clean and managed in a modern way, without thousands of files from the core that end-users don't have to touch ;-)

> I just pushed my suggestion. What do you think?

I'll check it out and think about it in these next few days :-) ❤️

fballiano:
@tmewes is it possible to remove the icons from admin/default/openmage and frontend/rwd? they should inherit from admin/default/default and frontend/base right?

I've never really liked (during the past 10 years) having many favicons coming from the core :-)

tmewes:
Yes, you are right. They are removed now. :)

fballiano:
nice, I'd also:
- remove the _getFaviconFile completely
- empty the getFaviconFile() method and replace it's content with a Mage::throwException, this way we don't keep deprecated code (that will end up stay forever) and at the same time gives a clear help during the update

what do you think? I'd also delete the methods completely and move on and write only a line in the release notes... I don't know...

fballiano:
I've to say that's beautiful to see the Maho logo favicon in the browser ehhehe

tmewes:
I think
> delete the methods completely and move on and write only one line in the release notes

would be best. Anything else, as you rightly pointed out, just means that we're dragging along unused old code forever. The methods are now removed.

PR #32, opened 2024-09-25
Improved branding

tmewes:
# Description

This PR improves the Maho branding. Only "safe" places were taken into account. For the xml files, the corresponding CSV files were already adjusted in the past so that these strings now match correctly again.

# Type of change

* Refactor


fballiano:
Thank you so much, amazing PRs!

PR #35, opened 2024-09-30
Replace `is_null` function calls with `=== null`

tmewes:
# Description

Currently, `is_null` and `=== null` are used very inconsistently, even though they are equivalent and `=== null` is a bit faster (see e.g. https://stackoverflow.com/a/8229005).
This PR replaces all `is_null` function calls in `app` with `=== null` (and `!is_null` analogously with `!== null`). 

Perhaps an additional check in a workflow would be good to ensure consistency so that only `=== null` is used in the future.

# Type of change

* Refactor


fballiano:
Hi @tmewes! What's the rector rule for this? cause we have a rector workflow:
- https://github.com/MahoCommerce/maho/blob/main/.github/workflows/rector.yml
- https://github.com/MahoCommerce/maho/blob/main/.rector.php

But I also see that this one breaks PHPStan for some reason

;-)

tmewes:
Unfortunately, I couldn't find a ready-made Rector rule on https://getrector.com/find-rule for that (only for [Exakat](https://exakat.readthedocs.io/en/latest/Reference/Rules/Php/IsnullVsEqualNull.html) - another PHP analyzer tool) and I don't know enough about Rector to be able to build it myself quickly. :/ Should I take a look at it and try to create an own rule for that (including testing and learning about Rector in general) or do you know more about it? Or should we do that later?

Yes, PHPStan has something to note about in three files, but as far as I can see it has rules for both `is_null()` and `=== null` here. It's only checking the modified lines, right? Then these matches affect `is_null()` just as before. I only adjusted and standardized all occurrences in `app` in this PR without changing any behavior (at least that was the plan).
How do we proceed? I would actually rather not make any potential behavior changes to the content of this PR that could potentially cause side effects. If necessary, we could do this in other PRs and then, e.g. remove unnecessary places where null is checked in order to fix the PHPStan messages. What do you think?

Please excuse me for not being very familiar with these tools yet. I will change that in the medium term.

fballiano:
Hi @tmewes, at the beginning I kept this PR on hold because, just 4 days after you created it, https://github.com/OpenMage/magento-lts/pull/4247 was created upstream which is (obviously...) the opposite... but that one didn't get merged yet but obviously we don't know what will be done upstream in the future.

What do you think about it?
Also, now that there are many conflicts with the current `main` branch?

fballiano:
For the time being, for the discrepancies between this PR and https://github.com/OpenMage/magento-lts/pull/4247 and the conflicts, I'll close this one.

Thank you @tmewes for the effort, I'll keep this topic in memory for the future and try to tackle it when time will be right.

PR #36, opened 2024-10-02
Removed Mage_Persistent and improved cookie handling

justinbeaty:
The whole Mage_Persistent module is pretty outdated as a concept, and could be removed entirely. Why have two cookies, where one (`om_frontend`) expires sooner, but the other (`persistent_shopping_cart`) remains keeping you in a semi-logged in state? Can't we just make the "Keep me logged in" box extend the lifetime of the regular cookie? Before I went headless with my store, I recall it always causing my customers problems. (https://github.com/MahoCommerce/maho/pull/34#issuecomment-2383932291)

This PR does a few things:

1) In the first commit `e4db9b8`, we remove all traces of the Mage_Persistent module. However, we add a new block `Mage_Customer_Block_Form_Remember` that renders the remember me checkbox, but in this commit it will never show up since we have no settings to control its functionality yet.
**Note: if you were to checkout this commit, everything should work the same as if Mage_Persistent still existed, but was disabled.**

2) In the second commit `5f13fd5`, we integrate the remember me functionality directly into Mage_Core and Mage_Customer while cleaning up some of the very old code from Magento 1.6 and older. I added new config options and also increased the default cookie to 7 days as the old default of 60 minutes is pretty short.
**Note: Minor BC break, I removed the typo'd constant `Mage_Core_Model_Resource_Session::SEESION_MAX_COOKIE_LIFETIME` as it's no longer needed.**

3) In the third commit `95326a4`, I just improve some of the `adminhtml` session code. For example the lifetime was hardcoded in `Mage_Core_Model_Session_Abstract_Varien` when we have a nice event that can be used instead.

4) There are lots of more commits since first creating this PR. See this comment: https://github.com/MahoCommerce/maho/pull/36#issuecomment-2411474664

## Screenshot of the new config:

image

## Todo:

- [x] Deal with translation csv files; merge strings from `Mage_Persistent.csv` into other files, define new strings, delete obsolete strings.
- [x] Remember me box isn't currently functional in the checkout screens.
- [x] Install script to `drop table persistent_session` and `alter table sales_flat_quote drop column is_persistent`
- [x] More testing, especially with other backends (db, redis)


fballiano:
this is amazing...

fballiano:
@justinbeaty about the install scripts, do you think they fit the mage_core or mage_customer module better? we will have to bump the version number of a module and that eventually will cause conflicts with openmage. Obviously we can manage those conflicts manually when they will happen but, do you have any idea if this situation would be avoidable somehow?

justinbeaty:
> this is amazing...

You just like it because of this 😅 
image

I actually have one more idea for improving session handling, namely the garbage collection routine. The file-based session gc is completely broken: we've probably all had cases of `var/session` filling to infinity, but there's another bug where it can expire sessions early. The db-based session is okay, but there's no need to gc during user requests when we could use cron. Also maybe some basic bot detection like in https://github.com/colinmollenhour/Cm_RedisSession.

Might be easier for testing to add to this PR.

fballiano:
> Might be easier for testing to add to this PR.

mmm I just wouldn't want to pile up too many things, I'll try to test it without all of these other interesting features only because closing small tasks is much easier ;-)

fballiano:
I'll work on the cvs translations right now

justinbeaty:
Great, I'm going to work ono the other tasks so we can get this closed.

fballiano:
(I work on the easy stuff ehehhehe)

fballiano:
@justinbeaty should I merge the translations in Mage_Core or Mage_Customer? In Mage_Customer would be "cleaner" because a lot of the strings I would merge already exist

justinbeaty:
> (I work on the easy stuff ehehhehe)

I'm pretty sure my translations script should still work here. It's especially useful since you could do:

```sh
git diff --name-only main | php -f translations.php -- missing
```

fballiano:
> I'm pretty sure my translations script should still work here. It's especially useful since you could do:
> 
> ```shell
> git diff --name-only main | php -f translations.php -- missing
> ```

it requires the shell_abstract files, I'd need to convert it to maho's CLI :-\

justinbeaty:
> > I'm pretty sure my translations script should still work here. It's especially useful since you could do:
> > ```shell
> > git diff --name-only main | php -f translations.php -- missing
> > ```
> 
> it requires the shell_abstract files, I'd need to convert it to maho's CLI :-\

Ah, right. Well I did plan on converting it to a Maho script eventually because it helps me not forget translation files in PRs.

fballiano:
> > > I'm pretty sure my translations script should still work here. It's especially useful since you could do:
> > > ```shell
> > > git diff --name-only main | php -f translations.php -- missing
> > > ```
> > 
> > 
> > it requires the shell_abstract files, I'd need to convert it to maho's CLI :-\
> 
> Ah, right. Well I did plan on converting it to a Maho script eventually because it helps me not forget translation files in PRs.

I've recreated shell/abstract.php and translations.php and it works but for example it doesn't detect translatable strings in XML so it would be missing a few from this specific porting.

Since it's a small file (<80 lines) I'll do that manually

justinbeaty:
> I've recreated shell/abstract.php and translations.php and it works but for example it doesn't detect translatable strings in XML so it would be missing a few from this specific porting.

Hm, it's supposed to read XML files too, but I haven't used it for quite some time.

fballiano:
done with the translations, some things were part of the mage_checkout, some others migrated to mage_customers, some others didn't exist anymore

fballiano:
hey @justinbeaty I'm slowly coming back to life after conference and travels (to the conference), do you want me to take specific tests on this PR?

justinbeaty:
@fballiano I am just going to do final tests on my end now. Then I have a comment half written about explaining some of the changes. If you can wait 1-2 hours then you may have an easier time testing / understanding the changes.

fballiano:
man I'll wait (or not) how much you'll tell me :-)

justinbeaty:
I tested pretty thoroughly on my end with combinations of:

* Different settings per store, i.e. English store is 30 days / 365 days, while French is 7 days / 30 days. German has Remember me disabled entirely.
* Files / db / redis session handler

For each store the cookie lifetime should be set appropriately, and the expires / TTL should match in the db / redis handler.

Only one thing I found isn't working: Set email confirmation to required and check Remember me during registration. The email confirmation link doesn't know you checked the box. I can fix this, but store owners will need to update their email templates.

There are also a few minor fixes, like getting rid of hardcoded numbers and using constants, fixing the lifetime of the store currency cookie, etc. If you have any questions about a change, please let me know.

---

## About session_start(['use_cookies' => false, 'use_trans_sid' => false]

PHP only sets the session cookie the first time the session was created. In subsequent requests, it doesn't send a new cookie, so it doesn't update the `Expires / Max Age` property. This is why later in `Mage_Core_Model_Session_Abstract_Varien::start()` Magento would renew the cookie if the session id did not change.

There are a few problems with this. First, we have two pieces of code setting session cookies. One for the first request, and another for all subsequent requests. This caused some inconsistencies between the initial cookie, and renewed cookies. Second, we can't change the session cookie parameters after the session has started, which is a problem if we want to use the observer to do that based on data in the session, such as if remember me was checked.

So instead, we tell PHP to not use cookies for sessions. We manually read the cookie to call `session_id()`, and we will manually set the cookie after `session_start()` and `session_regenerate_id()`.

This is probably the biggest change in the code. If the change is too uncomfortable, I can work around the issues. But to illustrate the point, I found a few scenarios where the two different cookie generation code were inconsistent and caused problems.

---


Config: default config
Problem: No SameSite setting is included on the first request. 

om_frontend=dg9a0kg5sogt278kb00d2g3h26; expires=Sun, 13 Oct 2024 16:29:07 GMT; Max-Age=3600; path=/; domain=store.example.com; secure; HttpOnly

om_frontend=dg9a0kg5sogt278kb00d2g3h26; expires=Sun, 13 Oct 2024 16:30:01 GMT; Max-Age=3600; path=/; domain=store.example.com; secure; HttpOnly; SameSite=None

Impact: Potential security problem if you set SameSite=Strict.

---


For the rest of the problems, use split http/https. I.e. `web/unsecure/base_url="http://store.example.com"`, `web/secure/base_url="https://store.example.com"`. Note: Test with Firefox, as Chrome will always redirect to https.


Config: split http/https, everything else default.
Problem: First page load correctly omits the secure flag, but all subsequent page loads do not. The set-cookie header with the secure flag will be ignored.


om_frontend=ejdkb554fpqnbb8p0kujdpn85u; expires=Sun, 13 Oct 2024 21:14:29 GMT; Max-Age=3600; path=/; domain=store.example.com; HttpOnly

om_frontend=ejdkb554fpqnbb8p0kujdpn85u; expires=Sun, 13 Oct 2024 21:14:30 GMT; Max-Age=3600; path=/; domain=store.example.com; secure; HttpOnly; SameSite=None
	
Cookie “om_frontend” has been rejected because a non-HTTPS cookie can’t be set as “secure”.

Impact: While browsing the site on non-secure URLs (i.e. product pages, category pages, shopping cart, etc), the session cookie will not be renewed with each page visit and eventually expire. When that happens, non-logged in users will lose their cart contents.


---

Config: split http/https, everything else default.
Problem: Visit a secured page, the cookie will be marked as secure, and then non-secure pages cannot access or modify the cookie.


om_frontend=ejdkb554fpqnbb8p0kujdpn85u; expires=Sun, 13 Oct 2024 21:15:44 GMT; Max-Age=3600; path=/; domain=store.example.com; secure; HttpOnly; SameSite=None

om_frontend=acbnfrfvva8mj1gljsn0q0qoel; expires=Sun, 13 Oct 2024 21:15:46 GMT; Max-Age=3600; path=/; domain=store.example.com; HttpOnly

Cookie “om_frontend” has been rejected because there is an existing “secure” cookie.

Impact: This causes tons of issues:
1) Magento is now starting a new session for every unsecure page load because the secure cookie isn't set.

2) Login to your account on the frontend, then go the homepage. You will appear to be logged out. Click the login button and you'll be redirected to the account dashboard.

3) As a guest, add something to your cart, visit a secure URL (like the login or checkout page), then return to a non-secure URL (like the homepage, or cart page). Your cart will appear to be empty. Go back to a secure page, and your cart will be there.

4) Once you have visited a secure URL, you will be unable to add anything to the cart. Instead you'll be redirected to the Enable Cookie Page with the message "Please enable cookies in your web browser to continue." This is especially problematic for logged in users, as they visited a secure page during login. Thus, logged in users can never add items to their cart.



---

For the next problem, switch back to Chrome.

Config: split http/https, set `web/cookie/cookie_httponly=No`.
Problem: First page load doesn't include domain in the set-cookie header, but subsequent page loads do. You end up with two cookies stored for the site, and the browser will send both to the server



om_frontend=mi7d2f4fgbtu3rembk32lsua66; expires=Sun, 13 Oct 2024 21:23:06 GMT; Max-Age=3600; path=/

om_frontend=mi7d2f4fgbtu3rembk32lsua66; expires=Sun, 13 Oct 2024 21:23:15 GMT; Max-Age=3600; path=/; domain=store.example.com; secure; SameSite=None

Impact: Magento will always read from the first cookie, but will update the second one. Load the homepage, click login, enter your email + password and press the login button. You'll be redirected back to the login page with no error message.

The problem doesn't seem to happen with Firefox, unsure if other non-Chromium browsers are affected.

---


## About Cm_RedisSession

I added a patch to revert https://github.com/colinmollenhour/Cm_RedisSession/pull/190 and remove the class rewrites. At this point, the redis session option is in OM v19, so really the class rewrite shouldn't exists, but I'm not sure if it's worth making a PR upstream. It's already been changed and changed back. So Maho users must set `<session_save>redis</session_save>` if they want to use redis.

I also have code in `Mage_Core_Model_Session_Abstract_Varien` that will set some default values if they're not set in local.xml. This is because `colinmollenhour/php-redis-session-abstract` has a default max lifetime of 30 days, but we may want to set the lifetime higher, especially if the  user checked remember me.

In the future, we could also have our own general `<session_config>` entry in local.xml that allows you to set min / max lifetime instead of hardcoding them in constants. But that's not very important right now.

**Note:** Perhaps you would like to create `mahocommerce/composer-patches` so we don't have to use my repo.

---

## Breaking changes

Besides the whole removal of the Mage_Persistent classes, there are just two breaking changes I can think of:

1) Removal of the `SEESION_MAX_COOKIE_LIFETIME` constant. It was only used in two places in core, the file it was defined in, and `Mage_Core_Model_Session_Abstract_Varien`. Users hopefully didn't override the latter one, and if they did they need to merge changes to get the new remember me  functionality anyway. But, I can add that constant back if you'd like.
2) The `<session_save>db</session_save>` option no longer using Redis.

Neither of these are too big of a deal, IMO.

---

## Future improvements

While working on this, a few ideas for later PRs

* Make SameSite=Lax the default, and possibly remove SameSite=None entirely.

* Better files and db session garbage collection.

* Better store switcher by not allowing the SID url param. Instead use POST and check the form key. Prevents all sorts of issues.

* Don't allow mixed http / https. Either run site fully secure (recommended), or not at all (certain local development environments). Get rid of unsecure / secure base_url, om_frontend_cid, etc.

justinbeaty:
Sorry for the novel above 😅, I just try to be pretty thorough in testing.

app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
```diff
@@ -89,68 +98,43 @@ public function start($sessionName = null)
...
-            if (!$cookieParams['secure']) {
-                unset($cookieParams['secure']);
-                if (!$cookieParams['domain']) {
-                    unset($cookieParams['domain']);
-                }
+        // Migrate old cookie from 'frontend'
+        if ($sessionName === Mage_Core_Controller_Front_Action::SESSION_NAMESPACE) {
+            if ($cookie->get('frontend') && !$cookie->get($sessionName)) {
+                $_COOKIE[$sessionName] = $cookie->get('frontend');
+                $cookie->delete('frontend');
```
fballiano, line 106: @justinbeaty do we need this code? I mean, is the "frontend" cookie still relevant?

app/design/frontend/rwd/default/template/page/html/header.phtml
```diff
@@ -22,7 +22,7 @@
 
         <?= $this->getChildHtml('currency_switcher') ?>
 
-        <p class="welcome-msg"><?= $this->getChildHtml('welcome') ?> <?= $this->getChildHtml('accountLinks') ?></p>
+        <p class="welcome-msg"><?= $this->getChildHtml('welcome') ?>
```
fballiano, line 25: is it right to remove accountLinks?

composer.json
```diff
@@ -76,6 +76,9 @@
                 "OM-1081 - Not detecting HTTPS behind a proxy": "https://raw.githubusercontent.com/OpenMage/composer-patches/74862e7a9fa29dd62c5c02d585d931ce9e3c5815/patches/OM-1081.patch",
                 "OM-2047 - Pass delimiter char to preg_quote": "https://raw.githubusercontent.com/OpenMage/composer-patches/74862e7a9fa29dd62c5c02d585d931ce9e3c5815/patches/OM-2047.patch",
                 "OM-2050 - Prevent checking known date codes": "https://raw.githubusercontent.com/OpenMage/composer-patches/74862e7a9fa29dd62c5c02d585d931ce9e3c5815/patches/OM-2050.patch"
+            },
+            "colinmollenhour/magento-redis-session": {
+                "Cm_RedisSession-Revert-190": "https://raw.githubusercontent.com/justinbeaty/composer-patches/7c39eba67913d0e97c121618ee0ba9d5c1588b42/patches/Cm_RedisSession-Revert-190.patch"
```
fballiano, line 81: I think we should have a Maho repo for all the patches, included the ones for ZF, would you like the idea?

app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
justinbeaty: If someone upgraded from Magento 1.9 -> Maho, it is still needed. But not if they went Magento 1.9 -> OpenMage -> Maho.

I would have liked to put that code in an Observer anyway. Especially to support if the user wants to change the cookie name themselves. But was just an unnecessary change in this PR.

app/design/frontend/rwd/default/template/page/html/header.phtml
justinbeaty: That block was only defined in `persistent.xml`, which is no more.

https://github.com/MahoCommerce/maho/blob/43669e44deee5fdba09d4854b5026e756e46abb1/app/design/frontend/base/default/layout/persistent.xml#L16-L20

We could re-define it somewhere like in `page.xml`, but maybe better to just replace it with `$this->getAdditionalHtml()` like is done with the base templates:

https://github.com/MahoCommerce/maho/blob/43669e44deee5fdba09d4854b5026e756e46abb1/app/design/frontend/base/default/template/page/html/header.phtml#L25

composer.json
justinbeaty: Probably a good idea yes, just in case someone force pushed or deleted the OM repo.

app/design/frontend/rwd/default/template/page/html/header.phtml
justinbeaty: Disregard last comment, I think actually better to redefine that block in page.xml, will do so now. Thanks for the catch.

composer.json
fballiano: done in https://github.com/MahoCommerce/maho/commit/73014072b0b52d20d7d573d8ce3c9371ccb4fae2

I'll now are your patch too

fballiano:
```suggestion
                "Cm_RedisSession-Revert-190": "https://raw.githubusercontent.com/MahoCommerce/maho-composer-patches/0424e4b13657d2745931b9063020b005428b9334/Cm_RedisSession-Revert-190.patch"
```

app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
fballiano: as first sight I'd say "duck it", I mean, if somebody still has magento1 I think session cookies are the last of their problem and I'd nuke them

app/design/frontend/rwd/default/template/page/html/header.phtml
fballiano: I think we should go with the implementation from "base" instead of the "accountLinks" which, if I'm not mistaken, seems to exist only in the persistent module with an "emulateAccountLinks" method

justinbeaty: Yup, after having waffled back and forth, I agree. Will push change after lunch (30 mins) unless you do sooner.

app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
justinbeaty: I had to re-read to see why the change was introduced in the first place. https://github.com/OpenMage/magento-lts/issues/990

So, we could also consider to use the same code to rename to “maho_frontend”, or maybe something different to make it so we aren’t recognized as OM.

fballiano: I didn't recognize if was for making it recognizable! then sure I'd use something like maho_frontend, maho_fe or something like that

justinbeaty: How about just “maho_session”? I think it’s probably more descriptive to a customer than “maho_frontend”, especially since these days many sites have a cookie policy page that lists the names of all cookies.

The admin cookie would still just be “adminhtml”.

fballiano:
Approved although i dont love the backend cookie name ehhehe

Il giorno lun 14 ott 2024 alle 21:49 Justin Beaty ***@***.***>
ha scritto:

justinbeaty:
> Approved although i dont love the backend cookie name ehhehe 

We can rename that too. “maho_admin_session”?

fballiano:
I think that would be more clear.
I also wouldnt care about migrating the cookie that much 😅

Il giorno lun 14 ott 2024 alle 21:53 Justin Beaty ***@***.***>
ha scritto:

justinbeaty:
Done. I figured we might as well check both `om_frontend` and `frontend` since the check only happens on the first visit. Could have saved some overhead by not using the `$cookie->get()` method and just checking the `$_COOKIE` global var, but I doubt it's adding more than 0.00001 ms per call for crawlers and other UA that don't store cookies.

app/code/core/Mage/Adminhtml/Model/Observer.php
```diff
@@ -66,4 +66,22 @@ public function clearCacheConfigurationFilesAccessLevelVerification()
         Mage::app()->removeCache(Mage_Adminhtml_Block_Notification_Security::VERIFICATION_RESULT_CACHE_KEY);
         return $this;
     }
+
+    /**
+     * Set the admin's session lifetime based on config
+     */
+    public function cookieSetLifetime(Varien_Event_Observer $observer): void
```
fballiano, line 73: @justinbeaty did you choose this name for a specific reason or can we use something like setCookieLifetime?

justinbeaty: `setCookieLifetime` seems better. It's also in Mage/Customer/Model/Observer.php too. If you want to fix feel free, I am on the EAV branch right now.

fballiano: done!

fballiano:
@justinbeaty is it normal that, if I delete all cookies and navigate to adminhtml, I get 2 cookies set (frontend and backend)? On OM I only get one.

justinbeaty:
> @justinbeaty is it normal that, if I delete all cookies and navigate to adminhtml, I get 2 cookies set (frontend and backend)? On OM I only get one.

No it's not normal, but I cannot reproduce. Can you check the network log? Is it making any requests for something like a favicon, image, or other asset that is 404? If so, then that request would have been handled by the frontend that would be setting the cookie. It could be a cache issue too if you have a script or something cached that is trying to load a file not on this branch.

fballiano:
it may be something caused by some chrome extension

fballiano:
removing all extensions, it still happens on chrome

Screenshot 2024-10-15 alle 19 56 39

justinbeaty:
> it may be something caused by some chrome extension

Maybe, but I also have a password manager extension and no problem. On the network tab can you click the first item "admin" and the response headers Set-Cookie headers? Is it setting maho_session? If not then click through each one to see where it's being set.

fballiano:
mmmm there's only maho_admin_session

Screenshot 2024-10-15 alle 20 11 38

and I couldn't find any request setting the frontend one :-\

fballiano:
it also happens on safari but it seems a request cookie... which I don't understand how it happens since I delete them...

Screenshot 2024-10-15 alle 20 23 52

justinbeaty:
I tried even setting my domain to `maho.test` thinking maybe something special with that TLD.

Does this happen on the main branch?

fballiano:
yes it does
Screenshot 2024-10-15 alle 20 36 55

then I guess let's see on https://demo.mahocommerce.com/ after merging?

justinbeaty:
That's really strange... Is it a git clone or starter template?

I would maybe say something with Caddy as that's the only thing I can think of different in our setups as I'm using nginx, but I don't see how that's even possible if none of the resources are returning a set-cookie header with `maho_session`.

fballiano:
> That's really strange... Is it a git clone or starter template?

clone of the main repo

> I would maybe say something with Caddy as that's the only thing I can think of different in our setups as I'm using nginx, but I don't see how that's even possible if none of the resources are returning a set-cookie header with `maho_session`.

it could also be that, no idea

justinbeaty:
What PHP version are you on? I'm on 8.2, but if you're on newer then I can check that too.

You also could deploy this branch at https://demo.mahocommerce.com/ before we merge if we want to check.

fballiano:
> What PHP version are you on? I'm on 8.2, but if you're on newer then I can check that too.

8.3

> You also could deploy this branch at https://demo.mahocommerce.com/ before we merge if we want to check.

mmm that's a strarter-project and it's a bit more complicated

> Edit: or actually if it's not happening on the demo site which is presumably on the main branch, then it must be something local, right?

right! it's not happening on the demo!

justinbeaty:
I'm really unsure what is happening then. I just double checked with a starter project too, and no problems.

My only guess is that something is requesting a page on the frontend, but that request is being hidden from you for some reason. But it doesn't make much sense that it happens on both browsers and with incognito.

fballiano:
for sure it's something with my local setup, I'd merge and test it, I don't see any problem with that.

do you think we're ready to merge?

justinbeaty:
> do you think we're ready to merge?

I do, I've tested it a bunch already and found no errors. If anything were to pop up we can fix.

fballiano:
exactly, I've also tested and I think everything is ready!

justinbeaty:
Sweet! Let's go! 🚀🚀🚀

fballiano:
updated https://demo.mahocommerce.com/ and no double cookie...

fballiano:
@justinbeaty you're doing unbelievable work here 🥹

justinbeaty:
> @justinbeaty you're doing unbelievable work here 🥹

I'm very glad to. I remember the whole persistent thing caused me grief back in the day, so I'm glad it can be resolved. I don't understand why it's still a thing in M2.

fballiano:
ehhehe many things are "still a thing" in M2 ;-)

fballiano:
@justinbeaty unfortunately this PR broke something, when maho is not installed and you open the browser, it should redirect to the web installer. git bisect identified commit 8152a23d2451bb0c1419ee89fd6ac4f67976702e as the source of the problem which should be this PR here. I'll debug.

justinbeaty:
> @justinbeaty unfortunately this PR broke something, when maho is not installed and you open the browser, it should redirect to the web installer but instead it outputs "Invalid website id requested.".
> 
> git bisect identified commit [8152a23](https://github.com/MahoCommerce/maho/commit/8152a23d2451bb0c1419ee89fd6ac4f67976702e) as the source of the problem which should be this PR here. I'll debug.

I am sorry, I thought I fixed the installer in #46. I will take a look also.

fballiano:
@justinbeaty don't worry man,  if I'm not mistaken the reason of the problem should be in Session/Abstract/Varien.php, I'll continue digging

justinbeaty:
> @justinbeaty don't worry man, if I'm not mistaken the reason of the problem should be in Session/Abstract/Varien.php, I'll continue digging

I am tracing it to `app/code/core/Mage/Customer/Model/Observer.php` in the `setCookieLifetime` method.

justinbeaty:
Quick fix is to do this:

```php
public function setCookieLifetime(Varien_Event_Observer $observer)
{
    if (Mage::isInstalled() && $observer->getSessionName() === Mage_Core_Controller_Front_Action::SESSION_NAMESPACE) {
    // ....
```

But let me also check is there's a better way.

fballiano:
true, your quick fix fixes is, it seems super weird that the system doesn't understand if it's installed only because of the cookie lifetime (which, I think, fails because of the getStoreConfig). 

maybe the session->init() is called by a place where it should check for Mage::isInstalled?

fballiano:
Screenshot 2024-11-14 alle 22 53 37


until the _redirect() it works correctly, it knows it has to redirect to /install. but then it calls setRedirectWithCookieCheck() which fails because it does something with the session

justinbeaty:
What I am finding is that:

1) In `Mage_Customer_Model_Session::__construct()`, we are calling `Mage::app()->getStore()->getWebsite()->getCode()`.
2) In `Mage_Core_Model_App::getStore()`, we check if instatlled, and if not we return `$this->_getDefaultStore()`.
3) In `Mage_Core_Model_App::_getDefaultStore()`, we return a `core/store` model with `id` and `code` properties set.
    * First problem, we should also add `->setWebsiteId(self::DISTRO_STORE_ID)` since normal Magento models have both `id` and the primary key set.
4) In `Mage_Core_Model_App::getWebsite()`, we try to load the website from the database, which of course will fail since there's no DB credentials in `local.xml`.

> maybe the session->init() is called by a place where it should check for Mage::isInstalled?

We could do this in `Mage_Customer_Model_Session::__construct()` instead, but there's also a few other places we call `getStore` there too, and I don't think littering that model with `isInstalled` calls feels correct.

> until the _redirect() it works correctly, it knows it has to redirect to /install. but then it calls setRedirectWithCookieCheck() which fails because it does something with the session

I'll investigate this path too.

justinbeaty:
So yes, we can also fix it with the below, because `setRedirectWithCookieCheck` does in fact initialize `core/session`, which dispatches the event causing the error.

```diff
diff --git a/app/code/core/Mage/Core/Controller/Varien/Action.php b/app/code/core/Mage/Core/Controller/Varien/Action.php
index 59469347b9..9e1bf4796a 100644
--- a/app/code/core/Mage/Core/Controller/Varien/Action.php
+++ b/app/code/core/Mage/Core/Controller/Varien/Action.php
@@ -468,7 +468,7 @@ abstract class Mage_Core_Controller_Varien_Action
         if (!$this->getFlag('', self::FLAG_NO_CHECK_INSTALLATION)) {
             if (!Mage::isInstalled()) {
                 $this->setFlag('', self::FLAG_NO_DISPATCH, true);
-                $this->_redirect('install');
+                $this->_redirectUrl(Mage::getUrl('install'));
                 return;
             }
         }
```

As mentioned earlier in this PR, I would actually consider that at some point we remove the SID parameter option entirely, which would deprecate the `setRedirectWithCookieCheck` method and also fix the issue.

fballiano:
I think this fix! and I also like the idea of removing the SID and redirectwithcookie, all ancient concepts (actually, on another topic... secure/unsecure urls are also an ancient concept)

I'll push this fix now :-)

fballiano:
done in https://github.com/MahoCommerce/maho/commit/18be58c902454de56b7a9e494111714f5a9b262a :-) thank you so much!

justinbeaty:
> I think this fix! and I also like the idea of removing the SID and redirectwithcookie, all ancient concepts (actually, on another topic... secure/unsecure urls are also an ancient concept)

For sure, although removing SID will be a bit tricky if we want to allow switching between stores with separate domains. The "best solution" would be for the store owner to submit their sites to Google's [Related Website Sets](https://developers.google.com/privacy-sandbox/cookies/related-website-sets) list which would allow cross domain cookie sharing. 

The next best solution would be something like creating a temporary key that lasts only a few minutes, then POST the key to the other domain for authentication.

At the very least, we just POST the cookie value to the other domain, which is still better than putting it in the URL, but doesn't solve many of the security issues.

---

Here's the list of future improvements from earlier in this PR:

* Make SameSite=Lax the default, and possibly remove SameSite=None entirely.

* Better files and db session garbage collection. Explore use of SessionUpdateTimestampHandlerInterface and session.use_strict_mode.

* Better store switcher by not allowing the SID url param. Instead use POST and check the form key. Prevents all sorts of issues.

* Don't allow mixed http / https. Either run site fully secure (recommended), or not at all (certain local development environments). Get rid of unsecure / secure base_url, om_frontend_cid, etc.

fballiano:
> For sure, although removing SID will be a bit tricky if we want to allow switching between stores with separate domains. The "best solution" would be for the store owner to submit their sites to Google's [Related Website Sets](https://developers.google.com/privacy-sandbox/cookies/related-website-sets) list which would allow cross domain cookie sharing.

exactly... to me this "take your sessionID to another website" is a recipe for a vulnerability

also, who knows if somebody uses the cross domain authentication

> * Make SameSite=Lax the default, and possibly remove SameSite=None entirely.
> * Better files and db session garbage collection. Explore use of SessionUpdateTimestampHandlerInterface and session.use_strict_mode.
> * Better store switcher by not allowing the SID url param. Instead use POST and check the form key. Prevents all sorts of issues.
> * Don't allow mixed http / https. Either run site fully secure (recommended), or not at all (certain local development environments). Get rid of unsecure / secure base_url, om_frontend_cid, etc.

❤️

PR #38, opened 2024-10-02
Added support for Maho-specific install scripts

fballiano:
Specifics in https://github.com/MahoCommerce/maho/discussions/37

This PR addes `maho_version` column to the `core_resource` and all the software to manage install scripts specifically made for Maho.

They're stored in the `sql/maho_setup/` subfolder of any module, and the filenames have to have this naming `maho-24.10.0.php` where `24.10` is the date of when the script it added and `.0` is a sequential number.

From my tests it seems to work for both installs/upgrade.

@justinbeaty what do you think about this?

## Todo:

- [x] Test web installer (needs to execute maho's install scripts too)
- [x] Test CLI installer (needs to execute maho's install scripts too)

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/CacheController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/CacheController.php
@@ -68,6 +68,7 @@ class Mage_Adminhtml_CacheController extends Mage_Adminhtml_Controller_Action
             Mage::app()->cleanCache();
             Mage_Core_Model_Resource_Setup::applyAllUpdates();
             Mage_Core_Model_Resource_Setup::applyAllDataUpdates();
+            Mage_Core_Model_Resource_Setup::applyAllMahoUpdates();
             Mage::app()->getCacheInstance()->unbanUse('config');
             Mage::getConfig()->saveCache();
         } finally {
--- a/app/code/core/Mage/Core/Model/App.php
+++ b/app/code/core/Mage/Core/Model/App.php
@@ -350,6 +350,7 @@ class Mage_Core_Model_App
                 $this->_initCurrentStore($scopeCode, $scopeType);
                 $this->_initRequest();
                 Mage_Core_Model_Resource_Setup::applyAllDataUpdates();
+                Mage_Core_Model_Resource_Setup::applyAllMahoUpdates();
             }
 
             $this->getFrontController()->dispatch();
--- a/app/code/core/Mage/Core/Model/Config.php
+++ b/app/code/core/Mage/Core/Model/Config.php
@@ -1289,6 +1289,7 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
             case 'sql':
                 $dir .= DS . 'sql';
                 break;
+
             case 'data':
                 $dir .= DS . 'data';
                 break;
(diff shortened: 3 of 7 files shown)


justinbeaty:
Yes, looks exactly like what I had imagined!

There's only one possible issue I could see, what if someone defined in their module's `config.xml`:

```xml
<resources>
    <maho_setup>
        <setup>
            <module>My_Module</module>
        </setup>
    </maho_setup>
</resources>
```
Then Magento would try to run Maho scripts as sql or data upgrade/install scripts. Could be fixed with adding in these two places a check:

https://github.com/MahoCommerce/maho/blob/72c2180a509817a6e02db08801125a7da66bd157/app/code/core/Mage/Core/Model/Resource/Setup.php#L209-L212
https://github.com/MahoCommerce/maho/blob/72c2180a509817a6e02db08801125a7da66bd157/app/code/core/Mage/Core/Model/Resource/Setup.php#L240-L243

```php
// Don't run sql or data upgrades for any resource named maho_setup
if ($resName === 'maho_setup') {
    continue;
}
```

I might be overthinking this and would bet nobody has a resource named `maho_setup`.

fballiano:
@justinbeaty hehehe I was ready to start updating the PR :-) thanks for updating the commenti :-)

fballiano:
There's a limitation in this approach, it could happen that:
- today we have a maho script modifying something on the db
- openmage releases a new version with an install script that it's not compatible with the db that was updated by maho

which could be even more complicated in installation-phase since maho scripts run after all openmage install/data scripts.

I think the possibility of this happening is pretty limited but worth mentioning, anyway I don't think there's a completely 100% no-problem-proof for this kind of thing

@justinbeaty if you think this is acceptable I'd merge this so that we can go on with the rest of the tasks?

justinbeaty (approved):

justinbeaty:
Just found one minor bug. If you define two resources in one module, the `maho_setup` files will be executed twice. See https://github.com/justinbeaty/maho/commit/9042865ac6c48ad7a5a2f8e454ea310ec6fec84a

I don't think any core modules do this, but I have seen 3rd party modules do it. Solution could be to put `maho-*.php` files alongside the core ones, but let's wait until #37 is resolved.

justinbeaty:
Another bug, and I'm sorry I didn't catch this before merging. We need a version_compare in the below function, or else the Maho install scripts are run even if they have already been executed. Try putting `Mage::log("test")` in `app/code/core/Mage/Core/sql/maho_setup/maho-24.10.0.php` and you'll see it keeps getting executed. We didn't catch this because the `addColumn()` function checks if the column exists already, so no error.

```php
protected function _getModifySqlFiles($actionType, $fromVersion, $toVersion, $arrFiles)
{
    $arrRes = [];
    switch ($actionType) {

        //...

        case self::TYPE_MAHO:
            uksort($arrFiles, 'version_compare');
            foreach ($arrFiles as $version => $file) {
                if (version_compare($version, $fromVersion) === self::VERSION_COMPARE_GREATER
                    && version_compare($version, $toVersion) !== self::VERSION_COMPARE_GREATER
                ) {
                    $arrRes[] = [
                        'toVersion' => $version,
                        'fileName'  => $file
                    ];
                }
            }
            break;
```

fballiano:
> We need a version_compare in the below function, or else the Maho install scripts are run even if they have already been executed.

I had put a "die" in the scripts and I didn't notice this behavior but I'll do that right now

PR #40, opened 2024-10-04
Improvements to Maho's autoloaders

justinbeaty:
It started with me trying to enable cache under System > Cache Management, and it not working. I tracked the problem down to this line, but the problem only occurs if you're running via git clone, as I am doing for development.
https://github.com/MahoCommerce/maho/blob/737536e70df35c2de88f1546ed31258b32422f22/app/code/core/Mage/Core/Model/Config.php#L350

The reason is that because of the `array_reverse` it loads `app/etc/local.xml` before `app/etc/config.xml` and that's setting `<active>0</active>` under the db connection node. I re-wrote those functions in the two `Config.php` files to not use `array_reverse` so files are loaded in alphabetical order like before, and I can enable cache now.

To make the code a bit cleaner, I defined two new globals, `MAHO_IS_STARTER_KIT` which is true if you required `mahocommerce/maho` with composer, and `MAHO_ROOT_SOURCE_DIR` which always points to the root of maho's code, either in vendor or where you cloned it.

---

I was also debugging something else, and on a hunch I tried changing the order of `include_paths` and the autoloaders. I'm pretty sure I solved https://github.com/OpenMage/magento-lts/discussions/3462. You can override Zend classes  in any of `app/code/core/`, `app/code/local/`, `app/code/community/`, `lib/`, and any of those directories in 3rd party modules.

One thing that threw me for a loop though was the behavior of opcache. If you were to override `Zend/Log/Formatter/Abstract.php`, it wouldn't load because opcache has already cached the file with `require_once 'Zend/Log/Formatter/Abstract.php'`. There is an opcache setting to disable that, and I put that in for developer mode, but on a production system you may need to restart httpd or php-fpm.

---

Also I wanted to say how much I appreciate your idea of the whole autoloading system. It's amazing. Makes the whole project cleaner and more robust. I need to be offline for a bit, but take some time to review and ask any questions.


app/Mage.php
```diff
@@ -12,45 +12,49 @@
...
 define('PS', PATH_SEPARATOR);
-if (file_exists(__DIR__ . '/vendor/mahocommerce/maho')) {
-    define('BP', dirname(getcwd()));
-} elseif (str_contains(__DIR__, '/vendor/mahocommerce/maho')) {
-    define('BP', str_replace('/vendor/mahocommerce/maho/app', '', __DIR__));
+define('BP', str_replace('/vendor/mahocommerce/maho', '', dirname(__DIR__)));
+
+if (file_exists(BP . '/vendor/mahocommerce/maho')) {
+    define('MAHO_IS_STARTER_KIT', true);
+    define('MAHO_ROOT_SOURCE_DIR', BP . '/vendor/mahocommerce/maho');
```
fballiano, line 19: thank you @justinbeaty for all your amazing work, in the past couple of days I was super busy with personal stuff so I'm sorry I can't comment only briefly.

```suggestion
    define('MAHO_IS_CHILD_PROJECT', true);
    define('MAHO_FRAMEWORK_DIR', BP . '/vendor/mahocommerce/maho');
```

the concept of child project or child theme is borrowed from wordpress and it seems to me a bit easier to understand

do you like them?

justinbeaty: Yeah, that seems like better naming. Don't worry if you can't get to review it for a while.

app/Mage.php
```diff
@@ -12,45 +12,49 @@
...
     Mage::setIsDeveloperMode(true);
     ini_set('display_errors', '1');
     ini_set('error_prepend_string', '<pre>');
     ini_set('error_append_string', '</pre>');
+
+    // Fix for overriding zf1-future during development
+    ini_set('opcache.revalidate_path', 1);
 }
 
+require_once MAHO_ROOT_SOURCE_DIR . '/app/code/core/Mage/Core/functions.php';
```
fballiano, line 35: the thing is that having MAHO_ROOT_DIR (which already exists) and MAHO_ROOT_SOURCE_DIR is confusing to me.

fballiano (commented):
sorry @justinbeaty, isn't there a simpler way to solve the problem you had cause this is gonna be very difficult to test

app/Mage.php
justinbeaty: They do point to different locations though. 

`MAHO_ROOT_DIR` is the directory above the public dir.

`MAHO_ROOT_SOURCE_DIR` is the directory of Maho's source.

They're only the same if you git clone. Renaming the constant as you suggest would help I think.

app/code/core/Mage/Core/functions.php
```diff
@@ -312,16 +312,19 @@ function mahoGetComposerInstallationData(): array
...
                 continue;
             }
-            $packages[] = $package;
-            $packageDirectories[] = realpath($version['install_path']);
+
+            $path = realpath($version['install_path']);
+            if ($path !== BP) {
+                $packages[] = $package;
+                $packageDirectories[] = $path;
+            }
```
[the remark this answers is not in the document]
justinbeaty: With a git install, composer lists `mahocommerce/maho` as an installed package, but it's installed path is just where you cloned the repo. The check filters that out.

I'm not sure it's strictly needed, but anytime you use the results of that function, it seems you're expecting the path to be in vendor. I will investigate.

fballiano:
Git install is used only by us and it is forbidden to store projects, we
shouldnt even consider it as a possibility ans it’s never coveres in the
doc and never will 😂

Il giorno ven 4 ott 2024 alle 20:35 Justin Beaty ***@***.***>
ha scritto:

lib/Varien/Autoload.php
```diff
@@ -37,7 +37,7 @@ public static function instance()
      */
     public static function register()
     {
-        spl_autoload_register([self::instance(), 'autoload']);
+        spl_autoload_register([self::instance(), 'autoload'], true, true);
```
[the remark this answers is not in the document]
justinbeaty: I'm pretty certain it is needed if we want to be able to override Zend classes. I know that the composer autoloader does the same thing for Zend classes anyway, so it would only slow down PSR-4 classes, which I don't think we currently use a lot of. I'll also  investigate this more.

justinbeaty:
> Git install is used only by us and it is forbidden to store projects, we shouldnt even consider it as a possibility ans it’s never coveres in the doc and never will 😂

I do agree, but it's not great that things break on a git install, because it makes testing harder since things work differently between the two install types.

> sorry @justinbeaty, isn't there a simpler way to solve the problem you had cause this is gonna be very difficult to test

I could drop the Zend override stuff from this PR if it will make it easier to test.

fballiano:
> I could drop the Zend override stuff from this PR if it will make it easier to test.

I think people should use a composer patch if they need to patch zf1, overrides are always risky when there is an upgrade of the base file. I think this whole zf1 override thing is overrated.

justinbeaty:
One other thought just now, it might be possible to drop the varien autoloader all together, but still allow Zend overrides. Then we get the fast Composer autoloader if available, and fallback to the slow PSR-0 type.

Although I do agree it's not great practice to allow overrides.

justinbeaty:
@fballiano Check it out, no more Varien auto loader! There's only one caveat. Before you could do stuff like:

```php
$this->_frontController = new mage_core_controller_varien_front();
```

Where it was case-insensitive, not you cannot, it must be `Mage_Core_Controller_Varien_Front`.

But I don't think that's a deal breaker, and it doesn't seem like that's always an issue, I think maybe only the first call to the class must be right.

Still working on the PR.

fballiano:
> @fballiano Check it out, no more Varien auto loader!

wow! very interesting!!

> There's only one caveat. Before you could do stuff like:
> 
> ```
> $this->_frontController = new mage_core_controller_varien_front();
> ```
> 
> Where it was case-insensitive, not you cannot, it must be `Mage_Core_Controller_Varien_Front`.

mmm I kinda don't think that's a problem, since everybody uses an IDE (I hope) :-D

justinbeaty:
Also, question:

Do you think it's okay to change the output of `mahoGetComposerInstallationData()` to something like this?

```php
[
    'mahocommerce/maho' => [
        'type' => 'maho-source',
        'path' => '...',
    ],
    'my/module' => [
        'type' => 'magento-module',
        'path' => '...',
    ],
]
```

There might be a few keys to add to each module for an optimization idea I have.

justinbeaty:
> mmm I kinda don't think that's a problem, since everybody uses an IDE (I hope) :-D

I think I've seen some ancient modules that named something like `My_Module_Block_MySuperAwesomeBlock`, but didn't match the filesystem. But I can't recall, and it's worth not supporting that.

fballiano:
> Do you think it's okay to change the output of `mahoGetComposerInstallationData()` to something like this?
> 
> ```
> [
>     'mahocommerce/maho' => [
>         'type' => 'maho-source',
>         'path' => '...',
>     ],
>     'my/module' => [
>         'type' => 'magento-module',
>         'path' => '...',
>     ],
> ]
> ```

only difference is the key of the array right? I don't see any problem with that at this stage of the project.

> I think I've seen some ancient modules that named something like `My_Module_Block_MySuperAwesomeBlock`, but didn't match the filesystem. But I can't recall, and it's worth not supporting that.

exactly, it's the kind of code that should die anyway

justinbeaty:
> only difference is the key of the array right? I don't see any problem with that at this stage of the project.

Right now you return two arrays:
```php
['mahocommerce/maho', 'my/module']
['path1', 'path2']
```

And yeah, I don't think anyone has relied on it yet. 😄

fballiano:
right, change it if you think it's better, no problem

justinbeaty:
It is actually faster to check which dirs exist first in included modules instead of adding all possible options into the path. So I added a check in `mahoGetComposerInstallationData()`.

See this test: https://pastebin.com/z3T7cQ2j, test 1 is the old way, test 2 is new, ignore test 3 for now. While we will not be loading 3,581 classes on a page load, the numbers are still much better.
```
Test 1 took 0.7837700843811
Test 2 took 0.2346019744873
Test 3 took 1.171895980835
```

app/Mage.php
```diff
@@ -12,71 +12,95 @@
...
-Mage::register('original_include_path', get_include_path());
-
 if (!empty($_SERVER['MAGE_IS_DEVELOPER_MODE']) || !empty($_ENV['MAGE_IS_DEVELOPER_MODE'])) {
     Mage::setIsDeveloperMode(true);
     ini_set('display_errors', '1');
     ini_set('error_prepend_string', '<pre>');
     ini_set('error_append_string', '</pre>');
+
+    // Fix for overriding zf1-future during development
+    ini_set('opcache.revalidate_path', 1);
```
fballiano, line 32: do we need this if we don't care about the zf1 override? cause in some cases I prefer to enforce hard opcache settings for maximum performance and I wouldn't like to make it impossible to disable revalidate_path

composer.json
```diff
@@ -29,6 +29,7 @@
         "ext-soap": "*",
         "ext-spl": "*",
         "ext-zlib": "*",
+        "composer-runtime-api": "^2",
```
fballiano, line 32: should we require (if possible) composer ^2.8 instead instead, just to be sure?

app/Mage.php
justinbeaty: I am not sure, it's possible the path cache causes other issues, will have to test. It's basically for any `include` or `require` that uses a relative path, to catch a file that shows up in the path that's different than what was there when the file doing the including was compiled. I don't think it affects Composer's autoloader, but maybe somewhere else in Mage.

In any case, it's absolutely should be enabled in production. I'm not sure how much of an impact it is if you're using developer mode.

fballiano: right! sorry I didn't notice it was only for developer mode! closing this subthread, sorry!

justinbeaty: No problem, I'll still investigate, because it it does cause an issue, then users should be aware that they may need to flush the opcache when they re-deploy. But that wouldn't be needed on every deploy, just in some cases.

fballiano: same thing happens whenever you tweak opcache (which you should do), if you have a revalidate_freq to 0 (which in production is amazing) you have to reset opcache manually

composer.json
justinbeaty: I don't think we need to go that high, my composer version is 2.7.9 and was released exactly one month ago. But it looks like the runtime-api version is different anyway:

```sh
$ composer show -p composer-runtime-api
name     : composer-runtime-api
descrip. : The Composer Runtime API
keywords : 
versions : * 2.2.2
type     : library
homepage : 
source   : []  
dist     : []  
names    : composer-runtime-api
```

fballiano: mmm ok, it's just that the more updated everything is the lest weird combination of situations we have to test and support

app/Mage.php
```diff
@@ -12,71 +12,95 @@
...
-$paths[] = BP . '/app/code/core';
-$allModules = mahoGetComposerInstallationData()[1];
-foreach ($allModules as $module) {
+if (MAHO_IS_CHILD_PROJECT) {
+    $paths[] = BP . '/app/code/core';
+}
+
+$modules = mahoGetComposerInstallationData();
+foreach ($modules as $module => $info) {
     if (str_contains($module, 'mahocommerce/maho')) {
```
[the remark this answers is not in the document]
justinbeaty: caus I know there's stuff like `mahocommerce/maho-phpstan-plugin`, but that already shouldn't be picked up because it's filtered out already since it's not `maho-module` or `magento-module`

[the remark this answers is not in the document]
fballiano: @justinbeaty I think I wanted to remove the core from that foreach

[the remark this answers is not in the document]
justinbeaty: We can do `if ($info['type'] === 'maho-source') {` now since that's in the array.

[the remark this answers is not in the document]
fballiano: we have no control if somebody creates a module and uses type=maho-source so I'd hardcode the project name, seems safer no?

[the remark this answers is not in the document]
justinbeaty: If they did do that, then we would just not load it as a module though, just about to push a commit and I'll explain in a comment.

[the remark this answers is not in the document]
fballiano: mmmm it doesn't seem that we have to limit that in that way when we can check for the project name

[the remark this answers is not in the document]
justinbeaty: okay, no problem, so then let's just change to `===` instead of `str_contains` ?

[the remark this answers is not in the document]
fballiano: absolutely

justinbeaty:
Let me explain commit [6435ca1](https://github.com/MahoCommerce/maho/pull/40/commits/6435ca179fff6ce8f25ec0e98063b710d0068110):

There were two possible issues I could imagine:

1) It could be possible that the order in which we build our include path for 3rd party modules is not alphabetical. But glob does return files in alphabetical order. So I was thinking there could be a bug where we're preferring one module's xml files, but autoloading a different module's classes. This wouldn't be too farfetched, because a lot of module developers had something like a "Company_All" module that they bundled with all their extensions, sometimes not keeping them all the same version.
But, `\Composer\InstalledVersions::getAllRawData()` does also return packages in alphabetical order, so for now it's actually not a problem. It would be nice if we could get the ordering of packages in `composer.json`, but that's not important now.

2) It could be possible that non magento modules store XML files in a directory named `app/etc/`? I doubt it but you never know...

In any case, I tested and it's actually about 10x faster to use glob in a loop as opposed to globbing the whole vendor dir. I thought it would have been a performance hit.

justinbeaty:
One more small performance improvement. It's quicker to use `stream_resolve_include_path` than loop and use `file_exists`. With three modules installed it brought the time executed in that function on a frontend page load from 1.1ms -> 0.7ms. So not a lot, but still...

justinbeaty:
I am confused about phpstan. How is it possible `MAHO_IS_CHILD_PROJECT` isn't defined,  but `BP` is?

justinbeaty:
I am going to test this thoroughly now. I am very sincere in hoping you know that I am not trying to invalidate any of your work. Your whole autoloader / composer package idea is absolutely genius, and I would have never come up with the idea or thought how to implement it.

I mainly just wanted to fix any gotchas that would pop up once people start building or porting projects over to Maho. Especially since I make heavy use of small internal modules in my magento install. I count 23 modules in my live site.

I hope you're not too overwhelmed with the changes. We can talk through everything together. I'll try to figure out the phpstan issue too.

justinbeaty:
I spent the evening creating some [test modules](https://github.com/justinbeaty/maho-test-modules/tree/main/src) to try out different scenarios. Here are the results:

| Test | OM 21-beta2 | Main Git | Main Starter | PR 40 Git | PR 40 Starter |
| :--- | :---: | :---: | :---: | :---: | :---: |
| Enable Cache | :heavy_check_mark: | :x: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| Conflicting Config | "B" | "A" | "A" | "B" | "B" |
| AW Core | :heavy_check_mark: | :x: | :x: |  :heavy_check_mark: | :heavy_check_mark: |
| Basic Mod A | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| Basic Mod B | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :x: | :x: |
| Core Override | :heavy_check_mark: | :x: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| Zend Override Require Mod | :x: | :x: | :x: | :heavy_check_mark: | :heavy_check_mark: |
| Zend Override Autoload Mod | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| Zend Override Require Local | :x: | :x: | :x: | :heavy_check_mark: | :heavy_check_mark: |
| Zend Override Autoload Local | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| Xyz Template | :heavy_check_mark: | :heavy_multiplication_x: | :x: | :heavy_multiplication_x: | :heavy_check_mark: |
| Xyz Theme | :heavy_check_mark: | :x: | :x: | :heavy_check_mark: | :heavy_check_mark: |


## How to test

#### **If you are using the Maho starter template:**

```sh
# If you want to test out with this PR:
composer config repositories.maho-pr-40 vcs https://github.com/justinbeaty/maho.git
composer require 'mahocommerce/maho:dev-topic-autoloader'
composer config use-include-path true
composer dump
```

#### **If you are using git:**

* Just checkout this branch, or leave as main

## Installing the test modules

```sh
composer config repositories.maho-test-modules composer https://raw.githubusercontent.com/justinbeaty/maho-test-modules/refs/heads/main

composer require aw/core basic/mod config/a config/b core/override test/page xyz/templates xyz/theme myzend/override
```

## Each test explained

After installing the test modules, there exists a page at https://store.example.com/testpage that shows the result of most of these tests.

**Enable Cache:** Can you enable the cache in the backend?

**Conflicting Config:** If two modules, Module_A and Module_B define the same thing in their config.xml, which value will we use?

**AW Core:** Can we load a module that puts their code at `app/code/local/AW/Core`? (I'm looking at you AheadWorks...)

**Basic Mod A:** Nothing special, just check that `Mage::getModel('basic_mod/a')` works.

**Basic Mod B:** Can we load a class with `basic_mod_model_b` instead of `Basic_Mod_Model_B`?

**Core Override:** Can a module override a core Mage class in the local or community folder? Bad practice, I know...

**Zend Override Require Mod:** Can a module override a Zend class that is included with `require_once()` in zf1-future's source code?

**Zend Override Autoload Mod:** Can a module override a Zend class and have it be auto loaded?

**Zend Override Require Local:** Same as above, but by putting the class in `$webdir/lib`.

**Zend Override Autoload Local:** Same as above, but by putting the class in `$webdir/lib`.

**Xyz Template:** Can a module with a name starting with a letter after "m" override a template in rwd? Note: With Maho git, the expected answer is no, because the file exists in `app/design/frontend/rwd`, and it takes priority. If it is working, you'll see:
image


**Xyz Theme:** Change the design package from `rwd` to `xyz_theme` in the backend, load the homepage, and check to see if it works. If it does, you'll see:
image



## Include paths:  

Below is a list of all the directories we put in the include path:


<details><summary>OM 21-beta2, count = 4</summary>
<p>
app/code/local<br>
app/code/community<br>
app/code/core<br>
lib
</p>
</details> 

<details><summary>Main Git, count = 43</summary>
<p>
vendor/shardj/zf1-future/library<br>
app/code/local<br>
app/code/community<br>
app/code/core<br>
vendor/basic/mod/app/code/local<br>
vendor/basic/mod/app/code/community<br>
vendor/basic/mod/app/code/core<br>
vendor/config/a/app/code/local<br>
vendor/config/a/app/code/community<br>
vendor/config/a/app/code/core<br>
vendor/config/b/app/code/local<br>
vendor/config/b/app/code/community<br>
vendor/config/b/app/code/core<br>
vendor/core/override/app/code/local<br>
vendor/core/override/app/code/community<br>
vendor/core/override/app/code/core<br>
app/code/local<br>
app/code/community<br>
app/code/core<br>
vendor/test/page/app/code/local<br>
vendor/test/page/app/code/community<br>
vendor/test/page/app/code/core<br>
vendor/xyz/templates/app/code/local<br>
vendor/xyz/templates/app/code/community<br>
vendor/xyz/templates/app/code/core<br>
vendor/xyz/theme/app/code/local<br>
vendor/xyz/theme/app/code/community<br>
vendor/xyz/theme/app/code/core<br>
vendor/zend/override/app/code/local<br>
vendor/zend/override/app/code/community<br>
vendor/zend/override/app/code/core<br>
vendor/mahocommerce/maho/app/code/core<br>
lib<br>
vendor/basic/mod/lib<br>
vendor/config/a/lib<br>
vendor/config/b/lib<br>
vendor/core/override/lib<br>
lib<br>
vendor/test/page/lib<br>
vendor/xyz/templates/lib<br>
vendor/xyz/theme/lib<br>
vendor/zend/override/lib<br>
vendor/mahocommerce/maho/lib
</p>
</details> 

<details><summary>Main Starter, count = 39</summary>
<p>
vendor/shardj/zf1-future/library<br>
app/code/local<br>
app/code/community<br>
app/code/core<br>
vendor/basic/mod/app/code/local<br>
vendor/basic/mod/app/code/community<br>
vendor/basic/mod/app/code/core<br>
vendor/config/a/app/code/local<br>
vendor/config/a/app/code/community<br>
vendor/config/a/app/code/core<br>
vendor/config/b/app/code/local<br>
vendor/config/b/app/code/community<br>
vendor/config/b/app/code/core<br>
vendor/core/override/app/code/local<br>
vendor/core/override/app/code/community<br>
vendor/core/override/app/code/core<br>
vendor/test/page/app/code/local<br>
vendor/test/page/app/code/community<br>
vendor/test/page/app/code/core<br>
vendor/xyz/templates/app/code/local<br>
vendor/xyz/templates/app/code/community<br>
vendor/xyz/templates/app/code/core<br>
vendor/xyz/theme/app/code/local<br>
vendor/xyz/theme/app/code/community<br>
vendor/xyz/theme/app/code/core<br>
vendor/zend/override/app/code/local<br>
vendor/zend/override/app/code/community<br>
vendor/zend/override/app/code/core<br>
vendor/mahocommerce/maho/app/code/core<br>
lib<br>
vendor/basic/mod/lib<br>
vendor/config/a/lib<br>
vendor/config/b/lib<br>
vendor/core/override/lib<br>
vendor/test/page/lib<br>
vendor/xyz/templates/lib<br>
vendor/xyz/theme/lib<br>
vendor/zend/override/lib<br>
vendor/mahocommerce/maho/lib
</p>
</details> 

<details><summary>PR 40 Git, count = 12</summary>
<p>
app/code/local<br>
app/code/community<br>
vendor/aw/core/app/code/local<br>
vendor/basic/mod/app/code/local<br>
vendor/config/a/app/code/local<br>
vendor/config/b/app/code/local<br>
vendor/core/override/app/code/local<br>
vendor/test/page/app/code/local<br>
app/code/core<br>
vendor/zend/override/lib<br>
lib<br>
vendor/shardj/zf1-future/library
</p>
</details> 

<details><summary>PR 40 Starter, count = 14</summary>
<p>
app/code/local<br>
app/code/community<br>
app/code/core<br>
vendor/aw/core/app/code/local<br>
vendor/basic/mod/app/code/local<br>
vendor/config/a/app/code/local<br>
vendor/config/b/app/code/local<br>
vendor/core/override/app/code/local<br>
vendor/test/page/app/code/local<br>
vendor/mahocommerce/maho/app/code/core<br>
lib<br>
vendor/zend/override/lib<br>
vendor/mahocommerce/maho/lib<br>
vendor/shardj/zf1-future/library
</p>
</details> 


## Page load test

Place the following file at `app/etc/includes/benchmark.php`, enable cache, and load some frontend pages.
```php
<?php

$start_time = microtime(true);
register_shutdown_function(function() use ($start_time) {
    $time = number_format(microtime(true) - $start_time, 4);
    $path = $_SERVER['REQUEST_URI'];
    $msg = "took $time seconds for $path";
    Mage::log($msg, null, 'performance.log');
});
```

OM 21-beta2
```
2024-10-06T01:51:08+00:00 DEBUG (7): took 0.0277 seconds for /
2024-10-06T01:51:10+00:00 DEBUG (7): took 0.0447 seconds for /men.html
2024-10-06T01:51:11+00:00 DEBUG (7): took 0.1062 seconds for /men/new-arrivals.html
2024-10-06T01:51:12+00:00 DEBUG (7): took 0.1073 seconds for /men/new-arrivals/linen-blazer.html
```

Main Git (cache not enabled)
```
2024-10-06T01:59:58+00:00 DEBUG (7): took 0.2605 seconds for /
2024-10-06T01:59:58+00:00 DEBUG (7): took 0.1789 seconds for /men.html
2024-10-06T02:00:00+00:00 DEBUG (7): took 0.2210 seconds for /men/new-arrivals.html
2024-10-06T02:00:01+00:00 DEBUG (7): took 0.2505 seconds for /men/new-arrivals/linen-blazer.html
```

Main Starter
```
2024-10-06T02:23:06+00:00 DEBUG (7): took 0.0455 seconds for /
2024-10-06T02:23:09+00:00 DEBUG (7): took 0.0662 seconds for /men.html
2024-10-06T02:23:10+00:00 DEBUG (7): took 0.1085 seconds for /men/new-arrivals.html
2024-10-06T02:23:10+00:00 DEBUG (7): took 0.1287 seconds for /men/new-arrivals/linen-blazer.html
```

PR 40 Git
```
2024-10-06T02:42:33+00:00 DEBUG (7): took 0.0325 seconds for /
2024-10-06T02:42:35+00:00 DEBUG (7): took 0.0559 seconds for /men.html
2024-10-06T02:42:36+00:00 DEBUG (7): took 0.0881 seconds for /men/new-arrivals.html
2024-10-06T02:42:37+00:00 DEBUG (7): took 0.1101 seconds for /men/new-arrivals/linen-blazer.html
```

PR 40 Starter
```
2024-10-06T03:09:14+00:00 DEBUG (7): took 0.0348 seconds for /
2024-10-06T03:09:15+00:00 DEBUG (7): took 0.0533 seconds for /men.html
2024-10-06T03:09:16+00:00 DEBUG (7): took 0.0931 seconds for /men/new-arrivals.html
2024-10-06T03:09:17+00:00 DEBUG (7): took 0.1161 seconds for /men/new-arrivals/linen-blazer.html
```

## Notes about Zend overrides

#### **Overriding Zend classes in a maho-module:**

Modules that wish to override Zend classes should use a name that comes alphabetically before "shardj/zf1-future". So "myzend/override" works, but "zend/override" does not.

If you've already ran `composer install` (and so zf1-future exists in vendor), and then run `composer install myzend/override`, the second Zend test will fail. That is because when there are conflicting namespaces, composer uses the one installed first. However this is only a temporary issue. If you install your project in a fresh directory (or just delete vendor), and composer install, it will work. 

#### **Overriding Zend classes locally:**

The above problems don't exist if you're just wanting to override in your starter project. To test, first copy the files from the test module locally, and remove myzend/override.

```sh 
mkdir -p lib
cp -R vendor/myzend/override/lib/* lib/
composer remove myzend/override
```
Add to your composer.json:
```
"autoload": {
    "psr-0": {
        "Zend_": "lib/"
    }
}
```

And run: `composer dump`

## Further testing

The test modules are just cases where I thought something might be broken. We should still probably just click around the frontend / backend and make sure all is good.

fballiano:
OMG @justinbeaty you're the man 🥹

> I am very sincere in hoping you know that I am not trying to invalidate any of your work. Your whole autoloader / composer package idea is absolutely genius, and I would have never come up with the idea or thought how to implement it.

man, don't even think that, the fact that we can take an idea to the next level it's incredible and makes me super proud!
 
> I hope you're not too overwhelmed with the changes. We can talk through everything together. I'll try to figure out the phpstan issue too.

you did an incredible job with the testing, I'll do some in the next few days (i've been preparing and I'll be traveling for the MagentiamoDay so it will be a busy week) but I'll try to close this PR so that we can merge this one, the persistent one and then finally go for the EAV :-D

thank you for being this active ❤️

justinbeaty:
> you did an incredible job with the testing, I'll do some in the next few days (i've been preparing and I'll be traveling for the MagentiamoDay so it will be a busy week) but I'll try to close this PR so that we can merge this one, the persistent one and then finally go for the EAV :-D

Have fun at the conference. I'm sure you'll have a lot to talk about there!

This PR isn't blocking anything, so take your time. I plan to get back to the other two PRs this week, and make sure they land in 24.11.0.

composer.json
```diff
@@ -7,6 +7,9 @@
         "AFL-3.0"
     ],
     "type": "maho-source",
+    "repositories": [
+        { "type": "vcs", "url": "https://github.com/justinbeaty/maho-phpstan-plugin.git" }
```
justinbeaty, line 11: Just a note to make sure we remember to change this.

fballiano: maho-phpstan-plugin 1.2.0 is released ;-)

composer.json
```diff
@@ -29,6 +29,7 @@
         "ext-soap": "*",
         "ext-spl": "*",
         "ext-zlib": "*",
+        "composer-runtime-api": "^2",
         "cweagans/composer-patches": "^1.7",
         "ezyang/htmlpurifier": "^4.17",
         "mahocommerce/maho-composer-plugin": "dev-main",
```
fballiano, line 35:
```suggestion
        "mahocommerce/maho-composer-plugin": "^1.2",
```
would this be better?

fballiano:
hey @justinbeaty, I pulled this branch but I get

Screenshot 2024-10-07 alle 12 33 39

justinbeaty:
@fballiano I get that error only when I didn't run:
```
composer config use-include-path true
composer dump
```

I think I could actually check if that is enabled and display a better error message.

justinbeaty:
Although the error should really only affect people who already used `composer create-project` (since we should update the starter template), we could do a few things in Mage.php:

```php
/**                                                                                                                                
 * Require Composer autoloader                                                                                                     
 */
$loader = require_once BP . '/vendor/autoload.php';

// Option 1, force set                                                                                                             
$loader->setUseIncludePath(true);

// Option 2, check and error                                                                                                       
if ($loader->getUseIncludePath() === false) {
    die('Run: composer config use-include-path true; composer dump;');
}

// Option 3, fallback to a second autoloader                                                                                       
if ($loader->getUseIncludePath() === false) {
    spl_autoload_register(function ($class) {
        if ($file = stream_resolve_include_path(strtr($class, '_', DS) . '.php')) {
            return include $file;
        }
    });
}
```

fballiano:
On a plane now, i ha e this error but im not on the starter project, ill try to drop the vendor folder abd recreate it 😊

justinbeaty:
> On a plane now, i ha e this error but im not on the starter project, ill try to drop the vendor folder abd recreate it 😊

Probably just need `composer dump`

fballiano:
oki obviously it was my fault 😂 let's see the starter project after we merge ;-)

justinbeaty:
One other note for the future. If someone had a 3rd party module that fails the "Basic Mod B" test, and either couldn't fix it because it was encrypted or just didn't know how, they could fix it by adding a file in their project:

```php
<?php
# app/etc/includes/myautoloader.php

spl_autoload_register(function ($class) {
    $class = ucwords($class, '_'); // Fix for classes with incorrect capitalization
    if ($file = stream_resolve_include_path(strtr($class, '_', DS) . '.php')) {
        return include $file;
    }
});
```

Although I still wouldn't recommend it, because a module that does that probably has other very dubious code and probably not updated for a long time.

justinbeaty:
Should be the last fixes:

1) I used the option 2 of checking `$loader->getUseIncludePath()`, just because it could be possible someone doesn't use the starter kit and requires `mahocommerce/maho` manually, or maybe they remove that line from their `composer.json`. Idk, let me know if you want to change or drop this.

2) There is an option `<disable_local_modules>` in `local.xml`, and setting that rewrites the include paths, breaking everything. I made it only drop `BP . '/app/code/local'`, not sure if it should also drop the local code pool in 3rd party modules too. I'm really not sure who uses this feature.

3) Get rid of `Mage::register('original_include_path', get_include_path());` so nobody else misuses it.

fballiano:
> 1. I used the option 2 of checking `$loader->getUseIncludePath()`, just because it could be possible someone doesn't use the starter kit and requires `mahocommerce/maho` manually, or maybe they remove that line from their `composer.json`. 

I think it makes sense to have the check, although it's kind of an "overprotection" it doesn't bring any harm to have it

> 2. There is an option `<disable_local_modules>` in `local.xml`, and setting that rewrites the include paths, breaking everything. I made it only drop `BP . '/app/code/local'`, not sure if it should also drop the local code pool in 3rd party modules too. I'm really not sure who uses this feature.

I think that's fine to disable it only in the current project and not for 3rd party modules.

I'll take one more look to this before merging but thank you for this incredible work.

justinbeaty:
omg... I just found one other optimization. :weary:

It's nearly 10 ms faster if we scan the local/community/core folders and add to composer's list of PSR-0 prefixes. I didn't push to this PR, but [here's the commit](https://github.com/justinbeaty/maho/commit/3d614b544349661976ea4d699f1bc6a3159ad1e6). We still need include paths for the `/lib` folders though...

Maybe save for another PR later. I could even imagine having some sort of `./maho optimize` command that dumps the results to a file so we don't even have to scan during boot. But maybe that's too close to the old Magento Compiler though.

fballiano:
@justinbeaty sure you don't want to push this last commit? :-))))

justinbeaty:
> @justinbeaty sure you don't want to push this last commit? :-))))

I'm cooking something now. Hang tight. :wink:

fballiano:
ahhahaha I knew it :-)

justinbeaty:
@fballiano Check out [this diff](https://github.com/justinbeaty/maho/compare/topic-autoloader...topic-autoloader-no-include-paths) and changes to the [composer plugin](https://github.com/justinbeaty/maho-composer-plugin/compare/main...topic-autoloader).

It hooks into the composer dump event and loops through our paths adding them to composer's PSR-0 prefixes, which will be written to `vendor/composer/autoload_static.php`. Meaning in production we don't have to loop through anything during runtime, and we no longer require `use-include-paths`!

There's also the ability to use `composer dump -o` to build an optimized version that uses Composer's classmap feature to map every single class to the relevant file. 

Note: in developer mode, we do have to rebuild the PSR-0 paths since we might be adding new files since the autoloader was last dumped. The optimize flag should not be used in development, else new files will also not be picked up. I have a commented out test in Mage.php that checks this, but not sure that's the best way.

What do you think? I still need to test this idea more, make sure it all works. I would maybe consider either merging this PR as it exists first then new changes in another PR, or to close this PR and create a new one with both changes stacked. Just in case we need to go back to the original implementation in this PR.

Some speed tests:

```
// Current PR
2024-10-06T02:42:33+00:00 DEBUG (7): took 0.0325 seconds for /
2024-10-06T02:42:35+00:00 DEBUG (7): took 0.0559 seconds for /men.html
2024-10-06T02:42:36+00:00 DEBUG (7): took 0.0881 seconds for /men/new-arrivals.html
2024-10-06T02:42:37+00:00 DEBUG (7): took 0.1101 seconds for /men/new-arrivals/linen-blazer.html

// New changes (non-optimized)
2024-10-08T22:22:03+00:00 DEBUG (7): took 0.0273 seconds for /
2024-10-08T22:22:04+00:00 DEBUG (7): took 0.0442 seconds for /men.html
2024-10-08T22:22:04+00:00 DEBUG (7): took 0.0780 seconds for /men/new-arrivals.html
2024-10-08T22:22:05+00:00 DEBUG (7): took 0.0977 seconds for /men/new-arrivals/linen-blazer.html

// New changes (optimized)
2024-10-08T22:23:24+00:00 DEBUG (7): took 0.0250 seconds for /
2024-10-08T22:23:25+00:00 DEBUG (7): took 0.0407 seconds for /men.html
2024-10-08T22:23:26+00:00 DEBUG (7): took 0.0765 seconds for /men/new-arrivals.html
2024-10-08T22:23:26+00:00 DEBUG (7): took 0.0949 seconds for /men/new-arrivals/linen-blazer.html

```

justinbeaty:
> The thing that I'm still thinking about is the MAHO_ROOT_DIR and MAHO_FRAMEWORK_DIR, to me it's confusing to have those two constants and the naming is not really clear to me. I created MAHO_ROOT_DIR because it was the same as MAGENTO_ROOT, but now I don't think these 2 make sense together.

Okay, I greatly simplified the constants:
1) Create MAHO_PUBLIC_DIR
2) Just set BP = MAHO_ROOT_DIR, since that is always the case.
3) Get rid of MAHO_FRAMEWORK_DIR and MAHO_IS_CHILD_PROJECT. Instead, we use info from `$modules['mahocommerce/maho']`.

Also because I made sure to put `$modules['mahocommerce/maho']` as the first item in the array, I was able to simplify the code in the two `Config.php` files.

justinbeaty:
> I would probably merge this one as it is and bring the new version in the next release?

The only weird thing would be introducing constants in one release, then removing them the next. And requiring include paths, then again removing them.

Maybe I can test the new way today and see first if we have any problems. It's not totally undoing all the work already in this PR, just moving some of the code from app/Mage.php to the composer plugin.

fballiano:
at this point let's add this last work to this PR? makes sense to me now :-)

fballiano:
> The only weird thing would be introducing constants in one release, then removing them the next. And requiring include paths, then again removing them.

literally nobody is using maho at the moment, we can do that.
it's also a super-core feature that nobody should even touch.

fballiano:
@justinbeaty do you want to push the last version? I'm got an hosting for a demo and this fix with the cache will make a difference in the performance ;-)

app/code/core/Mage/Api/Model/Wsdl/Config.php
```diff
@@ -121,20 +123,22 @@ public function init()
              * Exclude Mage_Api wsdl xml file because it used for previous version
              * of API wsdl declaration
              */
-            $mergeWsdl->addLoadedFile('app/code/core/Mage/Api/etc/wsi.xml');
+            $mergeWsdl->addLoadedFile(mahoFindFileInIncludePath("$moduleDir/wsi.xml"));
+
+            // Base wsi file
+            $this->loadFile(mahoFindFileInIncludePath("$moduleDir/wsi.xml"));
```
fballiano, line 129: mmm :-\

app/locale/en_US/Mage_Adminhtml.csv
```diff
@@ -692,6 +693,7 @@
 "Only attributes with scope ""Global"", input type ""Dropdown"" and Use To Create Configurable Product ""Yes"" are available.","Only attributes with scope ""Global"", input type ""Dropdown"" and Use To Create Configurable Product ""Yes"" are available."
 "Only mapped fields","Only mapped fields"
 "Maho Project","Maho Project"
+"Optimized autoloader detected in developer mode.","Optimized autoloader detected in developer mode."
```
fballiano, line 696:
```suggestion
```

I think we can avoid developer-related messages from translations

app/locale/en_US/Mage_Adminhtml.csv
```diff
@@ -185,6 +185,7 @@
 "Block Name is required field.","Block Name is required field."
 "Block has been deleted.","Block has been deleted."
 "Blocks","Blocks"
+"Bootup warning:","Bootup warning:"
```
fballiano, line 188:
```suggestion
```

I think we can avoid developer-related messages from translations

app/code/core/Mage/Api/Model/Wsdl/Config.php
justinbeaty: I know you changed that previously, but it introduced a bug.
Notice that the `app/code/core/Mage/Api/etc/wsdl.xml` file is included in the main branch.

```
# OM
/home/www-data/work/openmage/app/code/core/Mage/Api/etc/wsdl2.xml
/home/www-data/work/openmage/app/code/core/Mage/Core/etc/wsdl.xml
/home/www-data/work/openmage/app/code/core/Mage/Directory/etc/wsdl.xml
/home/www-data/work/openmage/app/code/core/Mage/Customer/etc/wsdl.xml
/home/www-data/work/openmage/app/code/core/Mage/Catalog/etc/wsdl.xml
/home/www-data/work/openmage/app/code/core/Mage/Sales/etc/wsdl.xml
/home/www-data/work/openmage/app/code/core/Mage/CatalogInventory/etc/wsdl.xml
/home/www-data/work/openmage/app/code/core/Mage/Checkout/etc/wsdl.xml
/home/www-data/work/openmage/app/code/core/Mage/Tag/etc/wsdl.xml
/home/www-data/work/openmage/app/code/core/Mage/GiftMessage/etc/wsdl.xml
/home/www-data/work/openmage/app/code/core/Mage/Downloadable/etc/wsdl.xml

# Maho Main
app/code/core/Mage/Api/etc/wsdl2.xml
/home/www-data/work/maho/git/app/code/core/Mage/Core/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Directory/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Customer/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Catalog/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Sales/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/CatalogInventory/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Checkout/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Tag/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/GiftMessage/etc/wsdl.xml
----> /home/www-data/work/maho/git/app/code/core/Mage/Api/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Downloadable/etc/wsdl.xml

# Maho PR-40
/home/www-data/work/maho/git/app/code/core/Mage/Api/etc/wsdl2.xml
/home/www-data/work/maho/git/app/code/core/Mage/Core/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Directory/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Customer/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Catalog/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Sales/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/CatalogInventory/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Checkout/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Tag/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/GiftMessage/etc/wsdl.xml
/home/www-data/work/maho/git/app/code/core/Mage/Downloadable/etc/wsdl.xml
```

justinbeaty: It is because in `app/code/core/Mage/Api/Model/Wsdl/Config/Base.php` we keep track of files already added, and by not using `mahoFindFileInIncludePath()` then then this line:
```php
$mergeWsdl->addLoadedFile('app/code/core/Mage/Api/etc/wsi.xml');
```

has no effect when we try to load the full path in `loadModulesConfiguration()`

fballiano: I didn't understand the problem of the last commit 😅 but I'll check it out with a bit more time, is it possible that there's code out there that calls the `loadFile`? because it would stop working right? I know we're breaking a lot of things anyway but I liked that the new loading magic would search files in a compatible way with previous code

justinbeaty: It shouldn't break anything. I thought pretty long about it. Because here:

https://github.com/MahoCommerce/maho/blob/9f57164453a1adb409d627875620a4e12f207848/lib/Varien/Simplexml/Config.php#L425-L435

If the 3rd party module didn't use `mahoFindFileInIncludePath()` (which they likely didn't) then the file won't be found, because we are trying to load from `/htdocs/public` and no XML files are there. So we will use `mahoFindFileInIncludePath()` then.

It just saves double path lookups because most of the time in the core code we are passing absolute paths, so we already ran `mahoFindFileInIncludePath()`

app/code/core/Mage/Core/functions.php
```diff
@@ -303,83 +296,49 @@ function is_dir_writeable($dir)
 
 function mahoGetComposerInstallationData(): array
 {
-    static $composerInstallationData = null;
-    if ($composerInstallationData !== null) {
-        return $composerInstallationData;
-    }
+    return \Maho\MahoAutoload::getInstalledModules(BP);
+}
```
fballiano, line 300: at this point I'd probably remove mahoGetComposerInstallationData function completely

lib/MahoCLI/Commands/HealthCheck.php
```diff
@@ -23,11 +23,43 @@
...
+        $result = Command::SUCCESS;
+        $loader = require MAHO_ROOT_DIR . '/vendor/autoload.php';
+
+        // if ($loader->getUseIncludePath() !== true) {
+        //     $result = Command::FAILURE;
+        //     $output->writeln('');
+        //     $output->writeln('<comment>Warning: Use include paths not enabled.</comment>');
+        //     $output->writeln('Run: composer config use-include-path true; composer dump;');
+        // }
+
```
fballiano, line 37:
```suggestion
```

can we remove these?

composer.json
```diff
@@ -7,6 +7,9 @@
         "AFL-3.0"
     ],
     "type": "maho-source",
+    "repositories": [
+        { "type": "vcs", "url": "https://github.com/justinbeaty/maho-composer-plugin.git" }
+    ],
```
fballiano, line 12:
```suggestion
```

I'd proceed removing this and publishing to the real one

composer.json
```diff
@@ -29,9 +32,10 @@
         "ext-soap": "*",
         "ext-spl": "*",
         "ext-zlib": "*",
+        "composer-runtime-api": "^2",
         "cweagans/composer-patches": "^1.7",
         "ezyang/htmlpurifier": "^4.17",
-        "mahocommerce/maho-composer-plugin": "dev-main",
+        "mahocommerce/maho-composer-plugin": "dev-topic-autoloader",
```
fballiano, line 38:
```suggestion
        "mahocommerce/maho-composer-plugin": "dev-main",
```

lib/MahoCLI/Commands/HealthCheck.php
justinbeaty: Absolutely, just left until I was 100% positive we don't need use-include-paths anymore. I'm 99% sure now, as composer will always lookup the classes with PSR-0 prefixes, or via the classmap.

composer.json
justinbeaty: Sure, I just had one more change in the plugin for disable_local_modules support with the new PSR-0 support. I'll push a PR to the composer plugin now and we can review there.

fballiano:
HAHAHHA I like the title of your last commit

justinbeaty:
Wait, maybe we should have left the `_canUseLocalModules` variable and method? In case someone overrides the class?

fballiano:
let me see

justinbeaty:
> mmm method and variable are "protected" and somebody should have overridden/extended app/code/core/Mage/Core/Model/Config.php which should be quite unusual, would you want to leave them in a disabled state and deprecate them?

Not sure. One the other hand, if someone *did* extend Mage_Core_Model_Config`, and for some reasons checked if local modules were disabled, they might be better off knowing that the function is removed by getting an error.

justinbeaty:
We could *also* check to see if the user has it enabled, and log an exception there too. But I'm almost in favor of just having it completely removed.

fballiano:
nah, the thing is that if we log an exception not a lot of people check the exception log... if we throw an exception at that point it's the same as the code breaking so... I'd say let's go if you don't want to think about it for a bit

fballiano:
let's think about it for a couple of hours and then merge, it's a minor thing anyway I think and again I think in production nobody will want to have all local modules disabled, otherwise remove them altogether at that point

justinbeaty:
Okay, let me get back to testing. I'm pretty sure no more changes unless a problem. Will explain some of previous changes soon.

justinbeaty:
Okay, everything is passing with my test modules. Let me overview some changes. I know you understand most of this anyway, but just for completeness sake.

## Building our paths with `\MahoAutoload::generatePaths()`

This is basically the same code we had to build the include paths that would have been used by Varien_Autoload. One fix/improvement is that instead of:
`mod_a/app/code/local`, `mod_a/app/code/community`, `mod_a/app/code/core`, `mod_b/app/code/local`, `mod_b/app/code/community`, `mod_b/app/code/core`

We do all local, all community, all core, which seems more correct:
`mod_a/app/code/local`, `mod_b/app/code/local`, `mod_a/app/code/community`, `mod_b/app/code/community`, `mod_a/app/code/core`, `mod_b/app/code/core`


## Generating PSR-0 prefixes with `\MahoAutoload::generatePsr0()`

Using the paths from above, we let composer know these paths contain PSR-0 classes. Since Varien_Autoload would have scanned those same paths in the same order and found the same files, there should be no breaking change. However, it does save composer from having to look in every module's path for `Mage_` classes if we know there are none there.

## Composer `use-include-path`

Because we have generated all of the PSR-0 prefixes, we have no need to let composer look in the include paths for files. If it wasn't found in our PSR-0 scan, it wouldn't have been found by Composer. We do still use `set_include_path()` in Mage.php so that modules can `include` or `require` files that aren't PSR-0 compatible.

## Controller Classmaps

I believe the only type of file in Magento core that doesn't use PSR-0 classes are the controllers (but not `Controller` which do). So, we also scan those files and add them directly to Composer's classmap definition.

This has two benefits: (1) Varien_Router is faster because we don't have to loop all our paths to find the class. (2) We don't have to put `require` statements in controller classes that extend other controllers. Composer will find the class you're extending. 

Regarding point 2, it doesn't break if you do have require statements in controller files, they're just unnecessary so I removed them. The only place I could see something breaking is if a module includes a controller with a relative path like `include './AbstractController.php'`, and another module overrides the abstract class. Don't know if I explained that well, but it's so niche I wouldn't worry about it, and I'm pretty sure it would have been broken without this PR anyway.

## Zend overrides

There are now no caveats in overriding Zend, or any other class in lib. Both myzend/override and zend/override work. You also don't have to explicitly define PSR-0 classes in your composer.json. Just put the file in your module's or starter kit's lib folder and it will work.
---

I think that explains everything. Will take a lunch break now, but feel free to do a final review if needed.

justinbeaty:
Sweet! 🚀🚀🚀

fballiano:
@justinbeaty merged!! thank you so much for such amazing work! I'll create a page on the website with the info from your last commit.

I've published the new version to https://demo.mahocommerce.com and check how freaking fast it is (there's not even an FPC module nor redis yet!)!

justinbeaty:
> @justinbeaty merged!! thank you so much for such amazing work! I'll create a page on the website with the info from your last commit.
> 
> I've published the new version to https://demo.mahocommerce.com and check how freaking fast it is (there's not even an FPC module nor redis yet!)!

My absolute pleasure. When I get around to it I am going to try and spin up my store on Maho locally. I have a totally janky process for building the codebase that tries to work around some of the shortcomings of the openmage composer plugins.

fballiano:
boom https://mahocommerce.com/autoloaders/

justinbeaty:
> boom https://mahocommerce.com/autoloaders/

Nice! 😃

And good luck again to you for the conference. Looks like you'll be speaking in about 10 hours from now.

fballiano:
Thank mate, I should definitely go to sleep now ehehehe I'm reviewing the slides right now :-D

PR #45, opened 2024-10-17
API: added `maho_version` to deprecated `magentoInfo` method, also added `mahoInfo` method

justinbeaty:
Trying to make sure the new "Dropdown with Other Option" input type is working with the API. I noticed the v2 SOAP API doesn't include `maho_version`.

Just for my own notes, here is how I am connecting to each API type. I had a heck of a time trying to get WS-I compliance mode to work. I saw some related OM issues, and a [sole answer](https://magento.stackexchange.com/a/284379) on Magento Stack Exchange that helped.

```php
$base_url = 'https://store.example.com/api';
$username = 'test';
$password = 'password1';

ini_set("soap.wsdl_cache_enabled", "0");

// Soap V1
$client = new SoapClient("$base_url/soap/?wsdl");
$session = $client->login('test', 'password1');
$result = $client->call($session, 'magento.info');
$client->endSession($session);

// Soap V2
$client = new SoapClient("$base_url/v2_soap/?wsdl");
$session = $client->login('test', 'password1');
$result = $client->magentoInfo($session);
$client->endSession($session);

// Soap V2 WS-I
$client = new SoapClient("$base_url/v2_soap/?wsdl");
$session = $client->login(['username' => $username, 'apiKey' => $password]);
$args = ['sessionId' => $session->result];
$result = $client->magentoInfo([...$args]);
$client->endSession([...$args]);
```

### DIFF
--- a/app/code/core/Mage/Core/Model/Magento/Api.php
+++ b/app/code/core/Mage/Core/Model/Magento/Api.php
@@ -25,8 +25,6 @@ class Mage_Core_Model_Magento_Api extends Mage_Api_Model_Resource_Abstract
     public function info()
     {
         $result = [];
-        $result['magento_edition'] = Mage::getEdition();
-        $result['magento_version'] = Mage::getVersion();
         $result['maho_version'] = Mage::getMahoVersion();
 
         return $result;
--- a/app/code/core/Mage/Core/Model/Magento/Api.php
+++ b/app/code/core/Mage/Core/Model/Magento/Api.php
@@ -24,6 +24,7 @@ class Mage_Core_Model_Magento_Api extends Mage_Core_Model_Maho_Api
      *
      * @return array
      */
+    #[\Override]
     public function info()
     {
         Mage::log('Deprecated API call to magentoInfo, use mahoInfo instead');
--- a/app/code/core/Mage/Core/Model/Magento/Api.php
+++ b/app/code/core/Mage/Core/Model/Magento/Api.php
@@ -19,4 +19,19 @@
  */
 class Mage_Core_Model_Magento_Api extends Mage_Core_Model_Maho_Api
 {
+    /**
+     * Retrieve information about the current Maho installation
+     *
+     * @return array
+     */
+    public function info()
+    {
+        Mage::log('Deprecated API call to magentoInfo, use mahoInfo instead');
+
+        $result = parent::info();
+        $result['magento_version'] = Mage::getVersion();
+        $result['magento_edition'] = Mage::getEdition();
+
+        return $result;
+    }
 }
--- a/app/code/core/Mage/Core/Model/Magento/Api/V2.php
+++ b/app/code/core/Mage/Core/Model/Magento/Api/V2.php
@@ -17,6 +17,6 @@
  * @package    Mage_Core
  * @deprecated
  */
-class Mage_Core_Model_Magento_Api_V2 extends Mage_Core_Model_Maho_Api_V2
+class Mage_Core_Model_Magento_Api_V2 extends Mage_Core_Model_Magento_Api
 {
 }
--- a/app/code/core/Mage/Core/Model/Magento/Api/V2.php
+++ b/app/code/core/Mage/Core/Model/Magento/Api/V2.php
@@ -6,6 +6,7 @@
  * @package    Mage_Core
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright  Copyright (c) 2022-2023 The OpenMage Contributors (https://openmage.org)
+ * @copyright  Copyright (c) 2024 Maho (https://mahocommerce.com)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
@@ -14,7 +15,8 @@
  *
  * @category   Mage
  * @package    Mage_Core
+ * @deprecated
  */
-class Mage_Core_Model_Magento_Api_V2 extends Mage_Core_Model_Magento_Api
+class Mage_Core_Model_Magento_Api_V2 extends Mage_Core_Model_Maho_Api_V2
 {
 }
--- a/app/code/core/Mage/Core/Model/Maho/Api.php
+++ b/app/code/core/Mage/Core/Model/Maho/Api.php
@@ -5,7 +5,6 @@
  * @category   Mage
  * @package    Mage_Core
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
- * @copyright  Copyright (c) 2021-2023 The OpenMage Contributors (https://openmage.org)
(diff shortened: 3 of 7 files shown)


justinbeaty:
Also, one small patch for an API error caused by #36. Will make a PR tomorrow, but in case you try to use the API on main and get an error.

```patch
diff --git a/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php b/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
index 27d1e28aec..ed3af8ac40 100644
--- a/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
+++ b/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
@@ -122,7 +122,7 @@ class Mage_Core_Model_Session_Abstract_Varien extends Varien_Object
 
         // If we still do not have a session id, then read from the cookie value
         // Otherwise, we will be starting a new session.
-        if (empty($this->getSessionId()) && $cookie->get($sessionName)) {
+        if (empty($this->getSessionId()) && is_string($cookie->get($sessionName))) {
             $this->setSessionId($cookie->get($sessionName));
         }

```

fballiano:
should we just replace magento_version and remove magento_edition instead?

justinbeaty:
> should we just replace magento_version and remove magento_edition instead?

I'm not really sure. I think the reason magento_version is there was so that services could know that we're compatible with Magento 1.9.4.5. But at this point are we really still compatible?

justinbeaty:
> @justinbeaty we are but we also don't and that will not guarantee anything in the future... so to me it could also be misleading to say that's magento 1.9.4.5, it really isn't

I would lean towards removing them, but I don't rely on the API so I don't know if it would break integrations. But we're also not promising a drop-in replacement for Magento / OM.

fballiano:
> Edit: also maybe you were saying just make magento_version = 24.11.0, which would not break anything I don't think. Either way I think it's fine.

I think that would be pretty much the same as renaming the whole node to maho_version, cause none is expecting a version like 24.11.0 😅

fballiano:
mmm I see, but at the same time "magento_version" is the same since 10 years... who still needs it when also trying to support maho?

justinbeaty:
> mmm I see, but at the same time "magento_version" is the same since 10 years... who still needs it when also trying to support maho?

This is definitely one of those "ghosts from the past," so I'd say it's fine to just remove the old magento keys. If you want to edit this PR or just close and make the commit directly that's fine with me.

fballiano:
since we have this PR, let's keep it in the PR so that you will have credits.

I've changed much more... let me know what you think

justinbeaty:
> since we have this PR, let's keep it in the PR so that you will have credits.
> 
> I've changed much more... let me know what you think

I had to revert the operation name attribute back to `magentoInfo` to make V2 and WSI work using the code from the first comment, but kept the other branding changes. Unless we're considering renaming the method `mahoInfo`, but then would also need to move the class at `Mage/Core/Model/Magento/Api.php` to `Mage/Core/Model/Maho/Api.php`.

I also removed the `magento_version` and `magento_edition` keys from the model, not sure if you intended to keep it for some reason, if so then revert that commit.

fballiano:
> I had to revert the operation name attribute back to `magentoInfo` to make V2 and WSI work using the code from the first comment, but kept the other branding changes. Unless we're considering renaming the method `mahoInfo`, but then would also need to move the class at `Mage/Core/Model/Magento/Api.php` to `Mage/Core/Model/Maho/Api.php`.

mmmm ok maybe I'd like to add a mahoInfo in addition to the magentoInfo (deprecating it), I don't know

> I also removed the `magento_version` and `magento_edition` keys from the model, not sure if you intended to keep it for some reason, if so then revert that commit.

no no you did it right, I missed it

fballiano:
@justinbeaty what do you think about my last commit?

justinbeaty:
@fballiano I like, but the mahoInfo calls weren't working. For SOAP v1 it could have been just adding:

```xml
<resources_alias>
    <store>core_store</store>
    <maho>core_magento</maho>  <!-- This line -->
    <magento>core_magento</magento>
</resources_alias>
```

But V2 needs the new model class. So I added `app/code/core/Mage/Core/Model/Maho/Api.php` and `Api/V2.php`.

Note, it didn't add or change a new ACL path, but we certainly could like this (not tested)

```diff
diff --git a/app/code/core/Mage/Core/etc/api.xml b/app/code/core/Mage/Core/etc/api.xml
index 3c0e595dc8..91a196e5a3 100644
--- a/app/code/core/Mage/Core/etc/api.xml
+++ b/app/code/core/Mage/Core/etc/api.xml
@@ -38,11 +38,11 @@
             <core_maho translate="title" module="core">
                 <model>core/maho_api</model>
                 <title>Maho info API</title>
-                <acl>core/magento</acl>
+                <acl>core/maho</acl>
                 <methods>
                     <info translate="title" module="core">
                         <title>Get info about current Maho installation</title>
-                        <acl>core/magento/info</acl>
+                        <acl>core/maho/info</acl>
                     </info>
                 </methods>
             </core_maho>
@@ -88,11 +88,17 @@
                             <title>List of stores</title>
                         </list>
                     </store>
+                    <maho translate="title" module="core">
+                        <title>Maho info</title>
+                         <info translate="title" module="core">
+                             <title>Retrieve info about current Maho installation</title>
+                        </info>
+                    </maho>
                     <magento translate="title" module="core">
                         <title>Maho info</title>
                         <info translate="title" module="core">
-                            <title>Retrieve info about current Maho installation</title>
+                            <title>(deprecated) Retrieve info about current Maho installation</title>
                         </info>
                     </magento>
                 </core>
             </resources>

```

Finally, since we have two methods now, we could consider having `magento.info` still return the old magento_version and magento_edition keys. Just for backwards compatibility until the method is removed. Maybe even log an exception or system.log message.

justinbeaty:
> it makes a lot of sense

Do you want me to make a commit with the ACL and magentoInfo changes? I'm on the branch now so no problem to do so.

fballiano:
if you can, yep sure 🙏

justinbeaty:
Okay, I've added back `magentoInfo`. The WSI was a PITA, apparently because it sets `use="literal"` instead of `use="encoded"`, so I had to redefine the `magentoInfoRequest` etc elements.

Here are the responses:
```php
// mahoInfo
Array (
    [maho_version] => 24.11.0
);

// magentoInfo
Array (
    [maho_version] => 24.11.0
    [magento_version] => 1.9.4.5
    [magento_edition] => Community
);
```
ACL:
image

Once we want to remove the magentoInfo endpoint, just remove all traces of "magento" in `Mage/Core/etc/{api,wsdl,wsi}.xml` and the classes at `Mage/Core/Model/Magento`. Hopefully we won't let the old endpoint be there forever, but this seems like the best solution for now.

What do you think?

fballiano:
I'm 100% fine with this :-)

should I merge?

justinbeaty:
Yup, I've tested on V1, V2, and WSI, all good.

PR #46, opened 2024-10-19
Session Fixes for #36

justinbeaty:
Includes the patch from https://github.com/MahoCommerce/maho/pull/45#issuecomment-2421029210

I also noticed a new session was being started for every API request, two if Configuration > Maho Core API > Enable WSDL Cache set to "No". So if you call `login` then `magentoInfo`, that's 2-4 new files created at var/session every time. Run the script again and you'll have 4-8 new files. I noticed this after testing the API when my session folder had tons of files. If you call `print_r($client->__getCookies());` you'll see Magento's SOAP server is returning a cookie with the default name PHPSESSID.

The culprit was the Mage_Log_Model_Visitor class, which tries to start a core session object in its constructor. The core session object tries to start a session, but no session namespace is provided, so it creates a cookie named PHPSESSID. This bug also happens in OM.

So I put the following check in `app/code/core/Mage/Core/Model/Session/Abstract/Varien.php`

```php
// Do not start a session if no sessionName was provided
if (empty($sessionName)) {
    return $this;
}
```

I don't think this should cause any problem, because Magento should always call this function with a sessionName. The only place it didn't was on the install page where it was also setting the PHPSESSID cookie. I updated that to use the `maho_admin_session` cookie name.

I also noticed that the customer `setCookieLifetime` observer was causing an issue on the install page, so I just added a new variable to the event. It saves some cycles on a normal page load anyway.

I haven't tested this to the nines yet, but I don't think it will cause issues. If it does, then just the `if (empty($sessionName)) {` check should be reverted, the other changes can stay.

As a side note, Soap clients do support the usage of cookies, so one possible future improvement is to allow this instead of having to pass `$session` to every SOAP call. Of course, we can still support the old way too.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Model/Observer.php
+++ b/app/code/core/Mage/Adminhtml/Model/Observer.php
@@ -72,9 +72,10 @@ class Mage_Adminhtml_Model_Observer
      */
     public function setCookieLifetime(Varien_Event_Observer $observer): void
     {
-        /** @var Mage_Core_Model_Session $session */
-        $session = Mage::getSingleton('adminhtml/session');
-        if ($session->getSessionName() === Mage_Adminhtml_Controller_Action::SESSION_NAMESPACE) {
+        if ($observer->getSessionName() === Mage_Adminhtml_Controller_Action::SESSION_NAMESPACE) {
+            /** @var Mage_Core_Model_Session $session */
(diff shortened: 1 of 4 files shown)


fballiano:
tested the CLI installer just to be safe, everything ok

PR #47, opened 2024-10-23
Added `setTranslationHelper` method to `Mage_Core_Block_Abstract`

justinbeaty:
In the EAV editor, we do this something like this:
```php
$block = $this->getLayout()->createBlock('eav/widget_text');
```
And then in the text.phtml file:
```php
<?= $this->__($_attribute->getStoreLabel()) ?>
```

The problem is that the translate function will read from Mage_Eav.csv first when translating, when it really should be using Mage_Customer.csv. Of course, if the string isn't found in Mage_Eav.csv, Magento will fall back to other CSV files.

---

I looked into the core code to see if the Magento devs handled this case anywhere else. They did with the Adminhtml "Insert Widget..." dialog box. Funnily enough, their implementation was broken. (more on that later)

I took their idea and moved the functionality all the way up to Mage_Core_Block_Abstract. So on any block, you can call `setTranslationHelper()` which will make all uses of `$this->__()` in that block use a different module's csv files.

Example usages:

```php
class My_Module_Block_Test extends Mage_Core_Block_Template
{
    public function _construct()
    {
        // You can set the translation helper in the _construct method
        parent::_construct();
        $this->setTranslationHelper(Mage::helper('customer'));
    }

    public function _prepareLayout()
    {
        // You can set the translation helper on a programmatically created block
        // This is especially helpful for core/template types, since by default it uses Mage_Core.csv
        $block = $this->getLayout()->createBlock('core/template');
        $block->setTemplate('test.phtml');
        $block->setTranslationHelper(Mage::helper('customer'));
        $this->append($block);
    }

    public function someMethod()
    {
        // You can set the translation helper anywhere else too, but it will only affect later calls of __()
        $this->__('Foobar'); // Uses My_Module.csv
        $this->setTranslationHelper(Mage::helper('customer'));
        $this->__('Foobar'); // Uses Mage_Customer.csv
    }
}
```

---

I also fixed the Widget implementation of this. It was just a simple mistake by Magento devs, where they tried to read an XML attribute after calling `$xml->asCanonicalArray()`, which removes all attributes. This was supposed to translate strings in the Add Widget dialog with the module specified in the widget.xml file. I tested and it is fixed now.

---

Minor breaking change:

The Widget implementation used a magic setter for `setTranslationHelper()`, then the `getTranslationHelper()` method reads it with `$this->_getData('translation_helper')`. I think it's better to use an actual class member variable `$this->_translationHelper`, but that means that the magic getter would return null. This isn't an issue, because their implementation always returned the widget helper anyway.

If any 3rd party module does a similar thing (without a broken implementation) by using a magic setter for `setTranslationHelper()`, but defining a `getTranslationHelper()` method, they simply need to remove their `getTranslationHelper()` method or fix it to read `$this->_translationHelper`. If they don't do anything, it would just fall back to whatever helper they had set as the fallback.

I really don't think the change is an issue, because (in non-developer mode) we read from all csv files if the string isn't found in the preferred one. We could check the magic getter in `Mage_Core_Block_Abstract::getTranslationHelper()`, but I don't think that's necessary.

### DIFF
--- a/app/code/core/Mage/Widget/Model/Widget.php
+++ b/app/code/core/Mage/Widget/Model/Widget.php
@@ -89,7 +89,7 @@ class Mage_Widget_Model_Widget extends Varien_Object
         $object->setData($xml->asCanonicalArray());
 
         // Set module for translations etc.
-        $module = $object->getData('@/module');
+        $module = $xml->getAttribute('module');
         if ($module) {
             $object->setModule($module);
         }
--- a/app/code/core/Mage/Core/Block/Abstract.php
+++ b/app/code/core/Mage/Core/Block/Abstract.php
@@ -186,6 +186,13 @@ abstract class Mage_Core_Block_Abstract extends Varien_Object
     protected $_app;
 
     /**
+     * Translation helper instance
+     *
+     * @var Mage_Core_Helper_Abstract
+     */
+    protected $_translationHelper = null;
+
+    /**
      * Initialize factory instance
      */
     public function __construct(array $args = [])
@@ -1166,6 +1173,30 @@ abstract class Mage_Core_Block_Abstract extends Varien_Object
     }
 
     /**
+     * Helper getter for translations
+     *
+     * @return Mage_Core_Helper_Abstract
+     */
+    public function getTranslationHelper()
+    {
+        return $this->_translationHelper ?? $this->helper('core');
+    }
+
+    /**
+     * Helper setter for translations
+     *
+     * @param Mage_Core_Helper_Abstract $helper
+     * @return $this
+     */
+    public function setTranslationHelper($helper)
+    {
+        if ($helper instanceof Mage_Core_Helper_Abstract) {
+            $this->_translationHelper = $helper;
+        }
+        return $this;
+    }
+
+    /**
      * Translate block sentence
      *
      * @return string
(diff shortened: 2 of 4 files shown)


fballiano:
wow I love this, do you need me to do some tests? should I merge it right away?

justinbeaty:
> wow I love this, do you need me to do some tests? should I merge it right away?

You can test if you like, otherwise merge. It's not blocking anything because without this PR calling `setTranslationHelper` doesn't throw any error, but just has no effect.

app/code/core/Mage/Widget/Model/Widget.php
```diff
@@ -89,7 +89,7 @@ public function getConfigAsObject($type)
         $object->setData($xml->asCanonicalArray());
 
         // Set module for translations etc.
-        $module = $object->getData('@/module');
+        $module = $xml->getAttribute('module');
```
fballiano, line 92: why is this line? I see the behavior is the same but it is for performance?

justinbeaty: This is actually why despite the good idea, Magento's whole translation helper implementation was broken.

A few lines above it calls `$xml->asCanonicalArray()`, which calls $xml->_asArray(true)`. Setting the parameter as true doesn't add attributes to the Varien_Object:

https://github.com/MahoCommerce/maho/blob/main/lib/Varien/Simplexml/Element.php#L210-L220

So $module was always null.

justinbeaty: I even went back with git blame on the OM repo to see if it worked at some point but then was broken, but nope it was always broken...

PR #50, opened 2024-10-30
Improved PHPStan GitHub Action

justinbeaty:
I figured out why the PHPStan check was failing in #25. It turns out that `actions/checkout@v4` clones the target branch with the PR merged into it. This makes sense as you would want to make sure PHPStan passes once the PR is merged to main, but is sort of confusing when you run PHPStan locally and it passes.

So I found out how to add both the base branch and the PR branch into the matrix strategy. For PRs, we will now run **six** PHPStan jobs, one for each PHP version with each branch combination. So for instance, in #25, the base branch checks would fail because I hadn't merged changes into it and fixed the baseline, but the PR branch would pass just fine.

You might think this would make the the workflows take a really long time, but I also figured out there was in fact a bug in the restore-keys cache action. Before, it would try to load the newest cache object starting with `phpstan-result-cache-`. The problem is that 2 out of 3 times it was trying to load the cache generated from a different PHP version, which causes PHPStan to discard the cache entirely. It now runs all six jobs in 30-60s compared to the several minutes from before once an initial cache is generated for the main branch, assuming there isn't another reason PHPStan needs to discard the cache.

I also enabled `-vvv` mode on PHPStan, since it helped me notice the cache error. It doesn't seem to be too noisy at all.

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -8,10 +8,14 @@ on:
 
 jobs:
   phpstan:
-    name: Analyze
+    name: Analyze PHP ${{ matrix.php-versions }} (${{ matrix.refs == 'pr-head' && github.head_ref || github.base_ref || github.ref_name }})
     runs-on: [ubuntu-latest]
     strategy:
+      fail-fast: false
       matrix:
+        refs: ['target-branch', 'pr-head']
+        exclude:
+          - refs: ${{ github.event_name != 'pull_request' && 'pr-head' }}
         php-versions: ['8.2', '8.3', '8.4']
 
     steps:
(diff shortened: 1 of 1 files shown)


justinbeaty:
Here's a screenshot of all six checks. The ones titled "(main)" are checking the main branch with this PR merged into it, and the ones titled "(topic-phpstan-workflow)" are checking this branch only.

image

For direct commits to main or another branch, you'll only see the three checks corresponding to that branch. Screenshot:

image

fballiano:
Hi @justinbeaty, super interesting as usual. do you think it's necessary to have (for PRs) phpstan running also on the target branch?

otherwise we could add something like

```
exclude:
  - refs: ${{ github.event_name != 'pull_request' && 'pr-head' || 'target-branch' }}
```

to have it running only on the PR branch

fballiano:
your explanations are always the best

PR #51, opened 2024-11-02
Check only changed files for PHP and XML linting

justinbeaty:
I noticed the PHP Syntax check takes around 5 minutes to complete. The problem is that we are checking 5,000+ files every time, even if only a few files were changed.

Instead, I found a few helpful actions that let us get a list of files changed since the last successful check on the main branch. So if the last commit on main was green, we're only checking the files changed in the PR. If the last commit on main was red, we'll include the files changed in that commit too, so that a PR will check if it's fixing the errors that happened on main.

The `SamhammerAG/last-successful-build-action` action doesn't have a lot of stars on GitHub, but it does what it needs to and I'll likely make some PRs there to improve it further. I had originally used the `nrwl/nx-set-shas` action which is more popular, but it only checks files changed in PRs, not files changed since the last successful workflow. The samhammer action's code is much simpler, too.

I tested this pretty thoroughly, and it all seems fine. There is an issue that could happen if the last 100 runs of a given workflow id don't include any runs on the main branch, but that's something I'll PR upstream to the action's repo. It's unlikely to happen in Maho, anyway. As a side note, PHPStan checks for syntax errors too, so the PHP syntax check is really just backup. However, we might as well keep it because who knows if it can catch something that PHPStan doesn't.

The workflows now complete in under 30s, and I also added support for GitHub annotations. So you'll see the errors in the files diff view:

image

image


fballiano:
❤️

PR #54, opened 2024-11-17
Improved `adminhtml/widget_form_element_dependence` block with ability to use complex conditions

justinbeaty:
## Description
The adminhtml dependence block is useful for conditionally displaying or hiding form elements based on other inputs, however its logic is very limited. This PR adds support for complex conditional expressions.

## Rationale
While working on the EAV code, I wanted to clean up some logic and make it more generic for other entity types. The Magento devs used the dependence block for two fields, `html_allowed_on_front` and `wysiwyg_enabled`, but then stuffed the rest of the logic in a JS file:

https://github.com/MahoCommerce/maho/blob/08c565fe8f981798b2be38c89fcf4799738b2af8/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Attribute/Edit/Tab/Main.php#L230-L238

https://github.com/MahoCommerce/maho/blob/08c565fe8f981798b2be38c89fcf4799738b2af8/app/design/adminhtml/default/default/template/catalog/product/attribute/js.phtml#L68-L133

In fact, the JS logic even overrides some of the dependence block logic in that `html_allowed_on_front` is also shown on input types other than textareas. It's pretty messy!

## Solution

Instead, we add support for complex conditions that can include AND, OR, XOR, and NOT. Conditions can even be nested, like this:
```php
$block->addComplexFieldDependence('result', $block::MODE_OR, [
    'source_1' => '1',
    $block->createCondition($block::MODE_AND, [
        'source_2' => '1',
        'source_3' => '1',
    ]),
]);
```

With a little more effort, I was also able to support complex conditions in `system.xml` files, so a similar condition would look like this:

```xml
<result>
    <label>Result</label>
    <frontend_type>text</frontend_type>
    <depends>
        <condition operator="or">
            <source_1>1</source_1>
            <condition operator="and">
                <source_2>1</source_2>
                <source_3>1</source_3>
            </condition>
        </condition>
    </depends>
    <show_in_default>1</show_in_default>
    <show_in_website>1</show_in_website>
    <show_in_store>1</show_in_store>
</result>
```

To accomplish this, I rewrote the JS `FormElementDependenceController` class (in vanilla JS + JSDoc) and added some new method to the PHP class.

## Breaking changes
I don't believe there are any breaking changes here, however there are two other improvements:

1) You no longer have to call `addFieldMap`, unless you want to refer to the field name by an alias. For example, the below code is unnecessarily complex.
https://github.com/MahoCommerce/maho/blob/61d568216d194f9f9650128836447137b7107e11/app/code/core/Mage/Adminhtml/Block/Tax/Rate/Form.php#L159-L168

2) You can call `addFieldValue` to simulate a field's value that may not show up in the form. This is useful for fields that are hidden in certain scopes in the config section. Previously, the config form would detect hidden fields and evaluate the logic in PHP, but that's too messy to try and do with complex conditions.


fballiano:
hey @justinbeaty, quick thing, should the test module be there?

justinbeaty:
@fballiano yes, forgot to write that part in the description, it will be reverted before merging.

I included it since it's the easiest way to test this. Just go to system > config (likely need to logout / login) and test the various sections.

fballiano:
This PR is a very nice addition to the framework, I'm testing it :-)

it seems the complex text it not working:
Screenshot 2024-11-17 alle 17 04 03

justinbeaty:
> it seems the complex text it not working:

Hm, did you hard refresh the page since the `js/mage/adminhtml/form.js` file changed?

fballiano:
I'm dumb and can't read the instructions of the test... it works

justinbeaty:
> all tests work as expected

Great, I'm just pulling the commit into the EAV branch now, let me test everything there before we merge, then I'll revert the test module.

app/code/local/Test/Config/etc/system.xml
```diff
@@ -0,0 +1,422 @@
...
+                            <source_model>adminhtml/system_config_source_yesno</source_model>
+                            <show_in_default>1</show_in_default>
+                            <show_in_website>1</show_in_website>
+                            <show_in_store>1</show_in_store>
+                        </source_3>
+                        <result>
+                            <label>Result</label>
+                            <frontend_type>text</frontend_type>
+                            <depends>
+                                <condition operator="xor">
```
fballiano, line 199: brainstorm: what if instead of `<condition operator="xor">` or would be `<xor>`? I mean, I don't want to change everything now, it's to evaluate an idea

fballiano:
I'll record a screencast for the release notes ;-)

public/js/mage/adminhtml/form.js
```diff
@@ -436,135 +437,214 @@ SelectUpdater.prototype = {
...
+     * @param {Object.<string, string>} [config.field_values] - key/value pairs of fallback values for fields not present in the form
+     * @param {number} [config.levels_up] - deprecated: the number of ancestor elements to find the parent element to hide
+     * @param {boolean} [config.can_edit_price] - deprecated: prevent enabling price inputs, only use this option if dependence block contains no other elements!
      */
-    initialize : function (elementsMap, config)
-    {
-        if (config) {
-            this._config = config;
+    constructor(elementsMap, config = {}) {
+        console.log(elementsMap, config);
```
fballiano, line 458:
```suggestion
```

justinbeaty: That line is in the test module commit that will be reverted :)

app/code/local/Test/Config/etc/system.xml
justinbeaty: So I originally did have it be `<XOR>` etc, but `<condition operator="xor">` felt more XML-ish. Especially if we ever created any sort of schema validation.

But also, it couldn't be lowercase tags because that has a higher chance of conflicting with an actual fieldname. For example:
```xml
<depends>
    <not>1</not>
</depends>
```

Would take more effort to parse if we're looking at a config field like `general/web/not` or we're trying to evaluate a condition. So originally I thought to just use uppercase tags, but visually it doesn't look great.

Instead, conditions are **only** evaluated if the tagname is "condition" and it has an "operator" attribute, i.e. `<condition operator="not">`. If you wrote just `<condition>` it would look for a field name called "condition".

I figured this was the best option to prevent any backwards compatibility issues.

fballiano: spot-on as usual

fballiano:
I have the video for when you'll need to revert the commit

https://github.com/user-attachments/assets/6495824f-e3e5-4f81-8381-cb9fe2ca6d9e

public/js/mage/adminhtml/form.js
```diff
@@ -598,4 +678,4 @@ function setPostcodeOptional(zipElement, country) {
 
 if (typeof varienGlobalEvents != 'undefined') {
     varienGlobalEvents.attachEventHandler("address_country_changed", onAddressCountryChanged);
-}
\ No newline at end of file
+}
```
fballiano, line 681:
```suggestion
}

```

justinbeaty: Do I not already have a newline in this file? My editor always adds one.

fballiano: I think the same about my phpstorm but sometime it misses, I don't even know why is that important anyway

justinbeaty: Not sure why it's necessary, but I'm so used to it. This is how GitHub shows when a newline was added to the end of the file:
image

fballiano: checked out the file and can confirm it's missing :-)

Screenshot 2024-11-17 alle 20 02 43

justinbeaty: Hm, I think you are on the wrong branch, the file should have 681 lines.

justinbeaty: No problem, I've never had emacs fail to insert the line so I think we're good.

justinbeaty:
This is what we can now replace the hardcoded JS logic with, much cleaner! Plus events can hook into this and add their own dependence if needed.

```php
$block
    ->addFieldDependence('frontend_class', 'frontend_input', ['text', 'customselect'])
    ->addFieldDependence('is_filterable', 'frontend_input', ['select', 'multiselect', 'price'])
    ->addFieldDependence('is_filterable_in_search', 'is_filterable', '1')
    ->addComplexFieldDependence('is_required', $block::MODE_NOT, [
        'frontend_input' => ['media_image'],
    ])
    ->addComplexFieldDependence('is_unique', $block::MODE_NOT, [
        'frontend_input' => ['media_image'],
    ])
    ->addComplexFieldDependence('is_global', $block::MODE_NOT, [
        'frontend_input' => ['price'],
    ])
    ->addComplexFieldDependence('is_configurable', $block::MODE_AND, [
        'is_global' => '1',
        'frontend_input' => ['select'],
    ])
    ->addComplexFieldDependence('position', $block::MODE_NOT, [
        'is_filterable' => '0',
    ])
    ->addComplexFieldDependence('used_for_sort_by', $block::MODE_NOT, [
        'frontend_input' => ['multiselect', 'textarea', 'gallery'],
    ])
    ->addComplexFieldDependence('is_html_allowed_on_front', $block::MODE_AND, [
        'frontend_input' => ['text', 'textarea', 'select', 'multiselect', 'customselect'],
    ])
    ->addComplexFieldDependence('is_wysiwyg_enabled', $block::MODE_AND, [
        'frontend_input' => 'textarea',
        'is_html_allowed_on_front' => '1',
    ]);
```

But I think I have one change left to do...

fballiano:
> This is what we can now replace the hardcoded JS logic with, much cleaner! Plus events can hook into this and add their own dependence if needed.

wow...

justinbeaty:
Now you can configure which event name to use for `dependentEl.addEventListener()`. Maybe you want to trigger on `input` instead of on `change`...

This is also useful if you want to manually trigger the dependencies in case you have some other custom logic. Example:

```php
$block
    ->addConfigOptions(['on_event' => 'mycustomevent'])
    ->addFieldDependence('frontend_class', 'frontend_input', ['text', 'customselect'])
    ->addFieldDependence('is_filterable', 'frontend_input', ['select', 'multiselect', 'price'])
    // ...
```

And then in JS:

```js
// Do some stuff before

const formEl = document.getElementById('edit_form');
formEl.querySelectorAll('input, select, textarea').forEach((el) => {
    el.dispatchEvent(new CustomEvent('mycustomevent'));
});

// Do some stuff after
```

justinbeaty:
Just doing some more testing with the EAV code; I want to make sure there's no other edge-cases... Will wait until tomorrow to revert the test module and merge.

justinbeaty:
Some updates:

## Events
The following event will always trigger a dependency re-calculation, no matter the value of `config.on_event`. This actually solves my problem better than the config option, but it's nice to have both.

```js
 el.dispatchEvent(new FormElementDependenceEvent());
```

## Simple Condition Merging

If you call `addFieldDependence()` with the same target and source fields, the values will be merged. This is helpful in an observer wants to add their own value without overwriting the old values.

Before:
```php
// result will show only if source == 'bar'

$block
    ->addFieldDependence('result', 'source', 'foo')
    ->addFieldDependence('result', 'source', 'bar');
```

After:
```php
// result will show if source == 'foo' || source == 'bar'

$block
    ->addFieldDependence('result', 'source', 'foo')
    ->addFieldDependence('result', 'source', 'bar');

// This is equivalent to:
$block
    ->addFieldDependence('result', 'source', ['foo', 'bar'])
```

## Complex Conditions

I reworked the internal array structure a bit, I didn't like that you couldn't do add multiple conditions of the same type.

Before:
```php
// This wouldn't work, as we can't have multiple array elements with the same key!

$block->addComplexFieldDependence('result', $block::MODE_AND, [
    $block::MODE_OR => [
        'source_2' => 'foo',
        'source_3' => 'bar',
    ],
    $block::MODE_OR => [
        'source_3' => 'baz',
        'source_4' => 'qux',
    ],
]);
```

After:
```php
// This works!

$block->addComplexFieldDependence('result', $block::MODE_AND, [
    $block::createCondition($block::MODE_OR, [
        'source_2' => 'foo',
        'source_3' => 'bar',
    ]),
    $block::createCondition($block::MODE_OR, [
        'source_3' => 'baz',
        'source_4' => 'qux',
    ]),
]);
```

The `$block::createCondition()` method is only necessary if you want to create nested conditions, otherwise all the other examples still work.

## System.xml

Due to limitations in the way Magento parses config XML files, nodes must have unique names. Thus you couldn't have multiple `<condition operator="...">` nodes as siblings. I changed it so the node name only needs to start with "condition" (and have an operator attribute). Example:

```xml
<result>
    <label>Result</label>
    <frontend_type>text</frontend_type>
    <depends>
        <condition operator="or">
            <!-- Note: "condition_1" and "condition_2" could be named "condition_foo" and "condition_bar" -->
            <!-- As long as they're unique and start with with the word "condition" -->
            <condition_1 operator="and">
                <source_1>foo</source_1>
                <source_2>bar</source_2>
            </condition_1>
            <condition_2 operator="and">
                <source_3>baz</source_3>
                <source_4>qux</source_4>
            </condition_2>
        </condition>
    </depends>
    <show_in_default>1</show_in_default>
    <show_in_website>1</show_in_website>
    <show_in_store>1</show_in_store>
</result>
```

## Misc

Added some other methods to the block, like `$block->clearFieldDependence('result')`, `$block->getRawFieldDependence('result')`, and `$block->setRawFieldDependence('result', $condition)`. The latter two would be helpful if an observer or child class needs to change a condition slightly without having to completely recreate it.

I also added a new admin form in the test module for testing. I'll pull these changes back into the EAV code and make sure everything seems fine, otherwise I think it's in pretty good shape.

justinbeaty:
Btw, @fballiano what do you think about the static method / properties:

```php
$block->addComplexFieldDependence('result', $block::MODE_AND, [
    $block::createCondition($block::MODE_OR, [
```

Should we make them all non-static? Like this:

```php
$block->addComplexFieldDependence('result', $block->MODE_AND, [
    $block->createCondition($block->MODE_OR, [
```

app/code/core/Mage/Adminhtml/Block/Widget/Form/Element/Dependence.php
```diff
@@ -47,35 +53,193 @@ class Mage_Adminhtml_Block_Widget_Form_Element_Dependence extends Mage_Adminhtml
...
+    public function addFieldMap($fieldId, $fieldAlias)
+    {
+        $this->_fields[$fieldAlias] = $fieldId;
+        return $this;
+    }
+
+    /**
+     * Add configuration option to the javascript dependencies controller
+     *
+     * @see self::addConfigOptions()
```
fballiano, line 225:
```suggestion
```

app/code/core/Mage/Adminhtml/Block/Widget/Form/Element/Dependence.php
```diff
@@ -47,35 +53,193 @@ class Mage_Adminhtml_Block_Widget_Form_Element_Dependence extends Mage_Adminhtml
...
+    }
+
+    /**
+     * Add field alias to id mapping
+     *
+     * @param string $fieldId element ID in DOM
+     * @param string $fieldAlias alias to refer to this field by
+     * @return $this
+     */
+    public function addFieldMap($fieldId, $fieldAlias)
```
fballiano, line 216:
```suggestion
     */
    public function addFieldMap(string $fieldId, string $fieldAlias): self
```

fballiano: ahhh sorry this is for backward compatibility

app/code/core/Mage/Adminhtml/Block/Widget/Form/Element/Dependence.php
```diff
@@ -94,24 +268,31 @@ protected function _toHtml()
...
-        return Mage::helper('core')->jsonEncode($result);
+        return Mage::helper('core')->jsonEncode($this->_depends);
+    }
+
+    /**
+     * Config options with field map JSON map generator
+     *
+     * @return string
+     */
+    protected function _getConfigJson()
```
fballiano, line 289:
```suggestion
     */
    protected function _getConfigJson(): string
```

fballiano:
beautiful! I've checked and retested, let me know when you want to remove the test module and merge :-)

app/code/core/Mage/Adminhtml/Block/Widget/Form/Element/Dependence.php
justinbeaty: I only put that there because `addConfigOptions` is where the config options are documented, but I don't know how it shows up in PHPStorm, so feel free to commit the suggestion.

justinbeaty:
> beautiful! I've checked and retested, let me know when you want to remove the test module and merge :-)

Probably best to wait until tomorrow, as today had many distractions for me... I don't expect to make any more changes to this PR, but I still wanted to make sure it's working as intended in the EAV code.

justinbeaty:
@fballiano Test module has been reverted.

fballiano:
done!

PR #56, opened 2024-11-21
New feature: Improved jstranslator.xml capabilities

justinbeaty:
## Description
Improved the jstranslator.xml syntax to allow including strings based on the scope, layout handle, and/or script included in `<head>`. Also added support very basic `sprintf` like syntax in the JS Translate class.

## Rationale
I wanted to utilize the jstranslator.xml feature, but it has one very glaring problem: every string is included on every page load on both adminhtml and frontend. So if I add 10 new strings meant only for the admin panel, they'll also be included on the frontend. This could even be a security issue if you include strings that leak information about the server setup or installed modules.

Of course, this only happens if you're on a language pack other than English, so to illustrate the problem comment out the `unset()` line below. This simulates a language pack with all strings defined.

https://github.com/MahoCommerce/maho/blob/fddaaf97b9386183331efef84ee9105b6379d031/app/code/core/Mage/Core/Helper/Js.php#L143-L146

Then view the source of the homepage, and you'll see messages that are only used in the admin panel. I highlighted an example where it could be a security issue.

image

With this PR the string count goes from 70 -> 41, and will prevent future strings from being included unnecessarily.

## Solution

I added some new syntax options to jstranslator.xml, here is what's possible:

```xml
<?xml version="1.0"?>
<jstranslator>

    <!-- Legacy style, backwards compatible with existing XML files -->
    <my-string-1 translate="message" module="core">
        <message>This string is included everywhere</message>
    </my-string-1>

    <!--
         Global scope node, this is equivalent to the legacy style,
         except you only need to define the module="" attribute once.

         Note: the module="" attribute is always optional, defaults to 'core'
    -->
    <global module="core">
        <my-string-2 translate="message">
            <message>This string is also included everywhere</message>
        </my-string-2>
        <my-string-3 translate="message">
            <message>Another global string</message>
        </my-string-3>
        <!-- You can override the parent module attribute if you like -->
        <my-string-4 translate="message" module="catalog">
            <message>Global string translated with another module</message>
        </my-string-4>
    </global>

    <!-- Frontend only -->
    <frontend module="core">
        <my-string-4 translate="message">
            <message>This string is included only on the frontend</message>
        </my-string-4>
    </frontend>

    <!-- Adminhtml only -->
    <adminhtml module="core">
        <my-string-5 translate="message">
            <message>This string is included only on the admin panel</message>
        </my-string-5>
    </adminhtml>

    <!--
         Layout node, include only if the handle="" attribute matches the current layout

         Example handles:
           - STORE_default
           - THEME_frontend_rwd_default
           - catalog_product_view
           - PRODUCT_TYPE_configurable
           - customer_logged_in
           - cms_page
           - cms_index_index
           - etc

         Note: you can specify multiple handles by separating them with a comma
    -->
    <layout handle="cms_index_index" module="core">
        <my-string-6 translate="message">
            <message>This string is included only on the home page</message>
        </my-string-6>
    </layout>

    <!--
        Script node, include only if the script was added to the head block.
        This ONLY detects scripts added to the head block with addJs or addItem, it does not
        detect hardcoded <script src=""></script> tags.

        path="" attribute should match the parameter passed to addJs or addItem
        type="" attribute can be either 'js' or 'skin_js', defaults to 'js' if omitted

        Note: you can specify multiple scripts by separating them with a comma, but they all must be of the same type
    -->
    <script path="validation.js" type="js" module="core">
        <my-string-7 translate="message">
            <message>This string is only included if validation.js is loaded</message>
        </my-string-7>
    </script>

    <script path="js/msrp_rwd.js,js/opcheckout_rwd.js" type="skin_js" module="core">
        <my-string-8 translate="message">
            <message>This string is only included if msrp_rwd.js or opcheckout_rwd.js is loaded</message>
        </my-string-8>
    </script>

    <!--
         Scope attribute, you can combine the legacy style, layout, and script nodes with the
         area="" attribute to further limit loading strings
    -->
    <script path="validation.js" area="adminhtml" module="adminhtml">
        <my-string-9 translate="message">
            <message>This string is only included if validation.js is loaded AND we're on the admin panel</message>
        </my-string-9>
    </script>

</jstranslator>
```

Additionally, you can add messages manually with PHP code. You could do this either in a controller or block class, as long as it's before the head block is rendered (i.e. don't do it in a blocks `_toHtml()` method).

```php
// Add a single string translated with the 'core' module
Mage::helper('core/js')->addTranslateData('Some string'); 

// Add multiple strings translated with the 'adminhtml' module
Mage::helper('core/js')->addTranslateData([
    'My string 1',
    'My string 2',
    'My string 3',
], 'adminhtml');
```

Of course, you can also add messages in phtml files as before:
```php
<script>
    Translator.add('Some string', '<?= $this->__('Some string') ?>');

    Translator.add({
        'My string 1': '<?= $this->__('My string 1') ?>',
        'My string 2': '<?= $this->__('My string 2') ?>',
        'My string 3': '<?= $this->__('My string 3') ?>',
    });
</script>
```

## Basic sprintf compatibility 

I also added a *very* basic sprintf feature to the JS translate class. It only matches `%s` and `%d` tokens, but that's pretty much all that is ever used in Magento translation files. Example:

```js
Translator.translate('Please enter a password with at most %s characters.', 256);
```

## Comments

Most of the still included strings on the frontend are from `validation.js`. I moved some that are clearly admin only to `Mage/Adminhtml/etc/jstranslator.xml`. 

I also included a debug commit here that will be reverted. It simply includes all messages without a translation pack and pretty prints the JSON array in the source code.

### DIFF
--- a/app/code/core/Mage/Core/Helper/Js.php
+++ b/app/code/core/Mage/Core/Helper/Js.php
@@ -51,8 +51,7 @@ class Mage_Core_Helper_Js extends Mage_Core_Helper_Abstract
      */
     public function getTranslateJson()
     {
-        // return Mage::helper('core')->jsonEncode($this->_getTranslateData());
-        return json_encode($this->_getTranslateData(), JSON_PRETTY_PRINT);
+        return Mage::helper('core')->jsonEncode($this->_getTranslateData());
     }
 
     /**
@@ -133,9 +132,9 @@ class Mage_Core_Helper_Js extends Mage_Core_Helper_Abstract
         $messageText = is_array($messageText) ? $messageText : [$messageText];
         foreach ($messageText as $text) {
             $translated = Mage::helper($module)->__($text);
-            // if ($text && $text !== $translated) {
-            $this->_translateData[$text] = $translated;
-            // }
+            if ($text && $text !== $translated) {
+                $this->_translateData[$text] = $translated;
+            }
         }
     }
 
--- a/app/code/core/Mage/Core/Helper/Js.php
+++ b/app/code/core/Mage/Core/Helper/Js.php
@@ -35,12 +35,12 @@ class Mage_Core_Helper_Js extends Mage_Core_Helper_Abstract
      *
      * @var array
      */
-    protected $_translateData = null;
+    protected $_translateData = [];
 
     /**
      * Translate config
      *
-     * @var Varien_Simplexml_Config|null
+     * @var Mage_Core_Model_Translate_Config|null
      */
     protected $_config = null;
 
@@ -73,9 +73,7 @@ class Mage_Core_Helper_Js extends Mage_Core_Helper_Abstract
      */
     public function getScript($script)
     {
-        return '<script type="text/javascript">//<![CDATA[
-        ' . $script . '
-        //]]></script>';
+        return '<script type="text/javascript">' . $script . '</script>' . "\n";
     }
 
     /**
@@ -123,27 +121,83 @@ class Mage_Core_Helper_Js extends Mage_Core_Helper_Abstract
     }
 
     /**
+     * Add messages to the JS translation array
+     *
+     * @param string|array $messageText a single or array of messages to translate
+     * @param ?string $module the helper module to use for translating, defaults to 'core'
+     */
+    public function addTranslateData(string|array $messageText, ?string $module)
+    {
+        $module = $module ?: 'core';
+        $messageText = is_array($messageText) ? $messageText : [$messageText];
+        foreach ($messageText as $text) {
+            $translated = Mage::helper($module)->__($text);
+            if ($text && $text !== $translated) {
+                $this->_translateData[$text] = $translated;
+            }
+        }
+    }
+
+    /**
      * Retrieve JS translation array
      *
      * @return array
      */
     protected function _getTranslateData()
     {
-        if ($this->_translateData === null) {
-            $this->_translateData = [];
-            $messages = $this->_getXmlConfig()->getXpath('*/message');
-            if (!empty($messages)) {
-                foreach ($messages as $message) {
-                    $messageText = (string)$message;
-                    $module = $message->getParent()->getAttribute('module');
-                    $this->_translateData[$messageText] = Mage::helper(empty($module) ? 'core' : $module)->__($messageText);
-                }
-            }
+        // Get current area, i.e. "adminhtml" or "frontend" plus "global"
+        $areas = [
+            Mage_Core_Model_App_Area::AREA_GLOBAL,
+            Mage::app()->getTranslator()->getConfig(Mage_Core_Model_Translate::CONFIG_KEY_AREA)
+        ];
+        // Get current layout handles
+        $handles = $this->getLayout()->getUpdate()->getHandles();
 
-            foreach ($this->_translateData as $key => $value) {
-                if ($key == $value) {
-                    unset($this->_translateData[$key]);
+        // Get currently loaded JS files
+        $headBlock = $this->getLayout()->getBlock('head');
+        $scripts = $headBlock ? array_keys($headBlock->getData('items')) : [];
+
+        foreach ($this->_getXmlConfig()->getNode()->children() as $node) {
+            // Check for <frontend>, <adminhtml>, and <global> nodes
+            if (in_array($node->getName(), $areas) && !isset($node['translate'])) {
+                foreach ($node->children() as $child) {
+                    $module = $child->xpath('ancestor-or-self::*/@module')[0]['module'] ?? null;
+                    $this->addTranslateData((string)$child->message, (string)$module);
                 }
+                continue;
+            }
+            // Allow nodes to define a custom separator for area, handle, and script path attributes
+            $separator = $node['separator'] ?? ',';
+
+            // Check if we have an area attribute for all other node types
+            if (isset($node['area']) && !array_intersect(explode($separator, $node['area']), $areas)) {
+                continue;
+            }
+            // Check for <layout> nodes and if current layout handles match
+            if ($node->getName() === 'layout' && isset($node['handle'])) {
+                if (array_intersect(explode($separator, $node['handle']), $handles)) {
+                    foreach ($node->children() as $child) {
+                        $module = $child->xpath('ancestor-or-self::*/@module')[0]['module'] ?? null;
+                        $this->addTranslateData((string)$child->message, (string)$module);
+                    }
+                }
+                continue;
+            }
+            // Check for <script> nodes and if we have loaded the JS file
+            if ($node->getName() === 'script' && isset($node['path'])) {
+                $type = $node['type'] ?? 'js';
+                $paths = array_map(fn ($path) => "$type/$path", explode($separator, $node['path']));
+                if (array_intersect($paths, $scripts)) {
+                    foreach ($node->children() as $child) {
+                        $module = $child->xpath('ancestor-or-self::*/@module')[0]['module'] ?? null;
+                        $this->addTranslateData((string)$child->message, (string)$module);
+                    }
+                }
+                continue;
+            }
+            // Default to original behavior
+            if (isset($node->message)) {
+                $this->addTranslateData((string)$node->message, (string)$node['module']);
             }
         }
         return $this->_translateData;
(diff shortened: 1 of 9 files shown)


fballiano:
Hey @justinbeaty, the backend/frontend separation is very nice and needed, but I'm not super about the advanced features because if users use a full page cache, it will probably break everything right?

justinbeaty:
Put another way, for each page on the frontend:

- The same JS files are always loaded
- The same layout handles are always loaded (*except* `customer_logged_in` and `customer_logged_out`).

Thus for each page, the same strings are included whether FPC is enabled or not.

fballiano:
if I remember correctly some blocks are cache site wise, but probably you're right and the head block is page-based anyway (eg: because javascripts are different between category and product page).

I was checking [Lesti](https://github.com/fballiano/openmage-lesti-fpc/) but it does seem to only work on a page-basis. Maybe I was remembering stuff from Turpentine or the Magento EE full page cache module that's not relevant anymore.

justinbeaty:
So then I think we can agree the strings must be correct for the `<script path="">` type node.

It's still possible you could break things with layout handles. As a completely dumb example you add a randomly generated handle to each page load dynamically from your controller file, and somehow expect to add strings based on that. But I think the solution is, *just don't that* if you're using FPC. Which isn't unreasonable since there are other similar things you can't do when using FPC.

justinbeaty:
> if I remember correctly some blocks are cache site wise, but probably you're right and the head block is page-based anyway (eg: because javascripts are different between category and product page).

Also I would assume the head block is page-based since the `<title>` tag is in there too.

app/code/core/Mage/Core/Helper/Js.php
```diff
@@ -122,28 +121,84 @@ public function getJsSkinUrl($file)
         return Mage::getDesign()->getSkinUrl($file, []);
     }
 
+    /**
+     * Add messages to the JS translation array
+     *
+     * @param string|array $messageText a single or array of messages to translate
+     * @param ?string $module the helper module to use for translating, defaults to 'core'
+     */
+    public function addTranslateData(string|array $messageText, ?string $module)
```
fballiano, line 130:
```suggestion
    public function addTranslateData(string|array $messageText, ?string $module): string
```

it returns string right?

app/code/core/Mage/Core/Model/Translate/Config.php
```diff
@@ -0,0 +1,33 @@
...
+class Mage_Core_Model_Translate_Config extends Varien_Simplexml_Config
+{
+    /**
+     * Extends current node with xml from $source, appending nodes without merging
+     *
+     * @param Varien_Simplexml_Config $config
+     * @param boolean $overwrite Argument has no effect, included for PHP 7.2 method signature compatibility
+     * @return $this
+     */
+    public function extend(Varien_Simplexml_Config $config, $overwrite = true)
```
fballiano, line 26:
```suggestion
     * @param Varien_Simplexml_Config $config
     * @param boolean $overwrite Argument has no effect, included for PHP 7.2 method signature compatibility
     * @return $this
     */
    public function extend(Varien_Simplexml_Config $config, $overwrite = true)
```

I guess we can't add typing here? but also, we don't support php 7 so I don't know if that changes anything

app/code/core/Mage/Core/Helper/Js.php
justinbeaty: It returns nothing, so `void` would be more appropriate. I was thinking `void` is implied but looks like it's better to be explicit.

app/code/core/Mage/Core/Model/Translate/Config.php
justinbeaty: If I change the method signature to:

```php
public function extend(Varien_Simplexml_Config $config, bool $overwrite = true)
```

I get this PHPStan error:

```
 ------ ------------------------------------------------------------------------------------------------------------------------- 
  Line   app/code/core/Mage/Core/Model/Translate/Config.php
 ------ ------------------------------------------------------------------------------------------------------------------------- 
  22     Parameter #2 $overwrite (bool) of method Mage_Core_Model_Translate_Config::extend() is not contravariant with parameter  
         #2 $overwrite (mixed) of method Varien_Simplexml_Config::extend().
 ------ ------------------------------------------------------------------------------------------------------------------------- 
```

---

Regarding the comment, I guess it should really say "included for PHP 7.2 and above method signature", but I just used the same comment as in some other classes: 

https://github.com/search?q=repo%3AMahoCommerce%2Fmaho%20Argument%20has%20no%20effect%2C%20included%20for%20PHP%207.2%20method%20signature%20compatibility&type=code

justinbeaty:
Reverted the debugging commit, however you can apply it with:

```diff
diff --git a/app/code/core/Mage/Core/Helper/Js.php b/app/code/core/Mage/Core/Helper/Js.php
index 0c23ef6e64..0e50d47ae4 100644
--- a/app/code/core/Mage/Core/Helper/Js.php
+++ b/app/code/core/Mage/Core/Helper/Js.php
@@ -51,7 +51,8 @@ class Mage_Core_Helper_Js extends Mage_Core_Helper_Abstract
      */
     public function getTranslateJson()
     {
-        return Mage::helper('core')->jsonEncode($this->_getTranslateData());
+        // return Mage::helper('core')->jsonEncode($this->_getTranslateData());
+        return json_encode($this->_getTranslateData(), JSON_PRETTY_PRINT);
     }

     /**
@@ -132,8 +133,8 @@ class Mage_Core_Helper_Js extends Mage_Core_Helper_Abstract
         $messageText = is_array($messageText) ? $messageText : [$messageText];
         foreach ($messageText as $text) {
             $translated = Mage::helper($module)->__($text);
-            if ($text && $text !== $translated) {
-                $this->_translateData[$text] = $translated;
-            }
+            // if ($text && $text !== $translated) {
+            $this->_translateData[$text] = $translated;
+            // }
         }
     }
```

app/code/core/Mage/Core/Model/Translate/Config.php
fballiano: maybe we could do an @inheritDoc instead so we don't copy the docblock?

fballiano:
i've added the debug code locally to test and everything works beatifully, I though i've seen something wrong about the configurableswatches string but I was checking the category page instead of the product page, so, again, everything is perfect ❤️

fballiano:
I'll put your documentation in the release notes but I'll also have to create a documentation page on the website for this

justinbeaty:
> I'll put your documentation in the release notes but I'll also have to create a documentation page on the website for this

Sweet! At some point I'll probably help with the docu too.

app/code/core/Mage/Core/Model/Translate/Config.php
justinbeaty: Related question. These both would work, right?

```php
////////////////////
// Option 1:

/** @inheritDoc */
#[\Override]
public function someFunction() {}

////////////////////
// Option 2:

/** 
 * @inheritDoc
 */
#[\Override]
public function someFunction() {}
```

I see the second one in the source  code, but it's so long.... 😅

justinbeaty: Also.... I don't fully understand the point of `@inheritDoc` in the first place. PHPStan does it automatically if there's no docblock at all, but maybe it's some IDEs that need it?

fballiano: yeah it's unnecessary long, and there are a ton of useless docblock anyway like `/** constructor`. I also seem to have noticed that probably @inheritDoc is not even necessary anymore (I noticed it in phpstorm but I didn't really test for real)

fballiano: > Also.... I don't fully understand the point of `@inheritDoc` in the first place. PHPStan does it automatically if there's no docblock at all, but maybe it's some IDEs that need it?

probably it's just an old thing not necessary anymore, i've asked AI and it said:

> No, @inheritdoc and {@inheritDoc} tags are largely considered unnecessary in PHP documentation as of 2024. Modern PHP tooling has evolved to handle inheritance automatically without requiring explicit inheritance tags7.
> Symfony has abandoned the usage of @inheritDoc as of version 6.25
> Modern PHP documentation tools can determine inheritance independently7
> The distinction between @inheritDoc and {@inheritDoc} has been a source of confusion even for experienced developers3

fballiano: so, in phpstorm @inheritdoc seems to be useless.

I've removed a bunch of them and phpstan/phpcs didn't care at all, should we trash them all?

fballiano: I'd say inheritdoc is officially useless, i've used this code

```php
<?php

class test1
{
    /**
     * This method is nice!
     */
    public function method1(string $a, int $b): string
    {
        return $a . (string) $b;
    }
}

class test2 extends test1
{
    public function method1(string $a, int $b): string
    {
        return "inherited" . $a . (string) $b;
    }
}

$test = new test1();
$test->method1("a", 1);

$test2 = new test2();
$test2->method1();
```

and hovering the overridden method, the description of the base one is visible:
Screenshot 2024-11-22 alle 13 46 32

fballiano: removed all in commit 6c05282a79

PR #63, opened 2024-12-05
Composer plugin v3

justinbeaty:
I made a couple more changes to improve the autoloading in Maho, feedback is welcome!

## Cleanup app/Mage.php

First, I migrated the "startup" code over to `app/bootstrap.php`, so that `app/Mage.php` is a pure class file.

I also removed the `libxml_disable_entity_loader` stuff as it's been fixed since 2014. To quote a comment from 2020 in [the linked PHP bug report:](https://bugs.php.net/bug.php?id=62577) *"If you're still using libxml < 2.9.0, better change that."*

Finally, I created a `lib/Maho.php` class instead of using functions defined in `functions.php`. So for example: `Maho::findFile($file)` instead of `mahoFindFileInIncludePath($path)`. 

## Pure autoloading

Everything can now be autoloaded by Composer like any other modern codebase, and you no longer need to require `app/bootstrap.php` and `app/Mage.php`.

This makes `public/index.php` (and any other custom scripts) much simpler. Just require the autoloader, then you can use any class or function. For example:

```php
<?php # public/myscript.php

require '../vendor/autoload.php';

Mage::init();
Mage::helper('core')->uniqHash(); // It works!
```

For legacy scripts you can still do it the old way, but if you require the autoloader AND require `app/Mage.php` you'll get an error: *"Cannot declare class Mage, because the name is already in use"*. Simply changing the `require 'app/Mage.php'` to a `require_once` will fix that, but this scenario should be pretty rare.

## New utility functions

There's some new helper functions for working with installed modules:

```php
# Glob packages
foreach (Maho::globPackages('/app/etc/modules/*.xml') as $file) {
    // $file is an absolute path
}

# Find file (already existed, but improved)
$file = Maho::findFile('app/code/core/Mage/Core/Helper/Data.php'); 
// $file = '/home/www/maho/vendor/mahocommerce/maho/app/code/core/Mage/Core/Helper/Data.php'

# Convert back to relative path
$file = Maho::toRelativePath('/home/www/maho/vendor/mahocommerce/maho/app/code/core/Mage/Core/Helper/Data.php');
// $file = 'app/code/core/Mage/Core/Helper/Data.php';

# See more at lib/Maho.php
```

## Performance improvements

We gain a modest 1-2 ms as we don't need to build paths or scan `app/etc/includes/*.php` at runtime. These are pre-scanned during the `composer dump` command.

I believe there's also a possibly significant performance gain to be had with loading `phtml` files, but that's for another PR... 

## Modman support

I added modman support to the composer plugin. For example assume you install [Cm_Diehard](https://github.com/colinmollenhour/Cm_Diehard/blob/master/modman) via composer.

By default, you'll end up with files such as `vendor/colinmollenhour/cm_diehard/code/etc/config.xml`, but this isn't going to work with Maho.

With modman support, the composer plugin will create symlinks in a special folder `vendor/mahocommerce/maho-modman-symlinks/$packageName`, for example:

```
vendor/mahocommerce/maho-modman-symlinks/colinmollenhour/cm_diehard/app/code/community/Cm/Diehard/etc/config.xml
# is symlinked to ↓
vendor/colinmollenhour/cm_diehard/code/etc/config.xml
```

Then, if the composer plugin detects this folder, we'll use it as the module's install path instead. The feature will avoid creating symlinks if there's no actual difference in the source vs target files. For example modman files created with `generate-modman` don't need symlinks.

It even supports modman's `@import` feature which was never supported in the OM plugin. That means you can define multiple modules in a single repo.

I created some more [test modules](https://justinbeaty.github.io/maho-test-modules/#modman). Add the composer repo, then run: `composer require modman/testpage modman/simple modman/complex modman/nolink modman/composermap`

<details><summary>Here's how the symlinks look -- Click to expand</summary>
<p>
<pre>
$ tree vendor/mahocommerce/maho-modman-symlinks/
vendor/mahocommerce/maho-modman-symlinks/
└── modman
    ├── complex
    │   └── app
    │       ├── code
    │       │   └── local
    │       │       └── Modman
    │       │           ├── ComplexA -> ../../../../../../../../modman/complex/modules/mod_a/code
    │       │           └── ComplexB -> ../../../../../../../../modman/complex/modules/mod_b/code
    │       ├── etc
    │       │   └── modules
    │       │       ├── Modman_ComplexA.xml -> ../../../../../../../modman/complex/modules/mod_a/module.xml
    │       │       └── Modman_ComplexB.xml -> ../../../../../../../modman/complex/modules/mod_b/module.xml
    │       └── locale
    │           ├── en_US
    │           │   ├── Modman_ComplexA.csv -> ../../../../../../../modman/complex/locale/en_US/Modman_ComplexA.csv
    │           │   └── Modman_ComplexB.csv -> ../../../../../../../modman/complex/locale/en_US/Modman_ComplexB.csv
    │           └── it_IT
    │               ├── Modman_ComplexA.csv -> ../../../../../../../modman/complex/locale/it_IT/Modman_ComplexA.csv
    │               └── Modman_ComplexB.csv -> ../../../../../../../modman/complex/locale/it_IT/Modman_ComplexB.csv
    ├── composermap
    │   ├── lib
    │   │   ├── bar.php -> ../../../../../modman/composermap/lib/bar.php
    │   │   └── foo.php -> ../../../../../modman/composermap/lib/foo.php
    │   ├── public
    │   │   └── js
    │   │       ├── bar.js -> ../../../../../../modman/composermap/js/bar/bar.js
    │   │       └── foo.js -> ../../../../../../modman/composermap/js/foo/foo.js
    │   └── README.md -> ../../../../modman/composermap/README.md
    ├── simple
    │   └── app
    │       ├── code
    │       │   └── local
    │       │       └── Modman
    │       │           └── Simple -> ../../../../../../../../modman/simple/code
    │       ├── design
    │       │   └── frontend
    │       │       └── base
    │       │           └── default
    │       │               └── template
    │       │                   └── modman
    │       │                       └── simple -> ../../../../../../../../../../../modman/simple/templates
    │       └── etc
    │           └── modules
    │               └── Modman_Simple.xml -> ../../../../../../../modman/simple/module.xml
    └── testpage
        └── app
            ├── code
            │   └── local
            │       └── Modman
            │           └── Testpage -> ../../../../../../../../modman/testpage/code
            └── etc
                └── modules
                    └── Modman_Testpage.xml -> ../../../../../../../modman/testpage/module.xml

40 directories, 13 files
</pre>
</p>
</details> 

In addition to modman support, it also supports [composer.json `extra.map`](https://github.com/Cotya/magento-composer-installer/blob/main/doc/Mapping.md#mapping-with-composerjson).

I also had written a package.xml parser, but I believe it's useless since I don't think you can define any custom mappings there. In other words, `app/code/local/My/Module` always links to the same location, which already works.

## Bug fixes

* Before if you had simply created a folder like `app/code/core/Mage/Core/sql/core_setup` in a module or starter pack, it would have prevented the files in `mahocommerce/maho` from running. This is fixed now. Technically you could override a single setup script, but of course that's not recommended.

* I'm pretty sure there was a discrepancy in the order that module's `etc/config.xml` files were loaded vs models, blocks, phtml files, etc. I didn't confirm this but it should be fixed now anyway. This would only cause problems if two composer packages define the same module like some 3rd party modules do (i.e `AW_Core`) but one version was outdated and they weren't compatible.

## Other comments

* The new composer plugin would need to be released as v3, as Maho 24.11.0 won't work with it. 

* Besides the possibility of the *"Cannot declare class Mage"* error, and the removal of the `mahoFindFileInIncludePath()`, etc functions, I don't believe there are any other breaking changes.

* I think `app/code/core/Mage/Core/functions.php` really should be at `lib/Mage/functions.php`, but it's not really a big deal where it is now. Maybe eventually we can remove a lot of those functions anyway...

* Instead of `lib/Maho.php` we could put it somewhere like `lib/Maho/Core.php` and have it loaded via PSR-4. So instead of `Maho::findFile()` it would be `\Maho\Core::findFile()`. Might be useful in the long term if we want to split the file up into `\Maho\Core`, `\Maho\ErrorHandler`, `\Maho\Whatever`, etc.


public/api.php
```diff
@@ -10,18 +10,10 @@
...
 define('MAHO_PUBLIC_DIR', __DIR__);
 
-if (file_exists(MAHO_ROOT_DIR . '/app/bootstrap.php')) {
-    require MAHO_ROOT_DIR . '/app/bootstrap.php';
-    require MAHO_ROOT_DIR . '/app/Mage.php';
-} else {
-    require MAHO_ROOT_DIR . '/vendor/mahocommerce/maho/app/bootstrap.php';
-    require MAHO_ROOT_DIR . '/vendor/mahocommerce/maho/app/Mage.php';
-}
+require MAHO_ROOT_DIR . '/vendor/autoload.php';
```
fballiano, line 16: loving this semplification

justinbeaty: Absolutely, it's nice that everything is auto-loadable now.

Technically there's now some similar code in `app/bootstrap.php` to find `vendor/autoload.php`, but it will never be executed in a vanilla project. It's only there for backwards compatibility with custom scripts that don't use composer.

fballiano:
hey @justinbeaty, do we want to merge this PR for this release? Modman support it for sure a good addition.

justinbeaty:
> hey @justinbeaty, do we want to merge this PR for this release? Modman support it for sure a good addition.

I realized we should have merged earlier to make sure we had enough testing during development of other PRs, so I was almost thinking to merge right after this release.

But we could merge it tomorrow (so I can resolve conflicts & do some more quick testing) and release next week?

fballiano:
I'm totally fine with merging tomorrow, I'd like to work on a couple more things for the next release so we definitely have a bit of time :-)

justinbeaty:
> I'm totally fine with merging tomorrow, I'd like to work on a couple more things for the next release so we definitely have a bit of time :-)

Sounds great!

fballiano:
@justinbeaty should I give it a test? for sure we'll have to merge https://github.com/MahoCommerce/maho-composer-plugin/pull/3 first in order to have the right composer.json here :-)

justinbeaty:
Sorry for the delay, just testing it on my end now and fixed one bug. Will update you soon.

justinbeaty:
@fballiano Okay, I did quick re-testing and all seems fine. I installed my [test modules](https://github.com/justinbeaty/maho-test-modules?tab=readme-ov-file) as well as `fballiano/openmage-cssjs-minify` and `fballiano/openmage-defer-javascripts` and I found no problems other than the one I just fixed.

Yes the plugin PR will need to be merged and released as `3.0.0` before we merge this. But you should be able to just pull this branch and composer install to test.

fballiano:
testing!

fballiano:
I can't see any fault :-)

justinbeaty:
Let's merge it then. I had tested a lot when I first opened the PR and couldn't find any problems either. Just that one typo I fixed already in the other PR but I must have been working on a local copy of the plugin and didn't push.

composer.json
```diff
@@ -6,6 +6,9 @@
         "OSL-3.0",
         "AFL-3.0"
     ],
+    "repositories": [
+        { "type": "git", "url": "https://github.com/justinbeaty/maho-composer-plugin.git" }
+    ],
```
fballiano, line 11:
```suggestion
```

composer.json
```diff
@@ -33,7 +36,7 @@
         "composer-runtime-api": "^2",
         "cweagans/composer-patches": "^1.7",
         "ezyang/htmlpurifier": "^4.17",
-        "mahocommerce/maho-composer-plugin": "^2",
+        "mahocommerce/maho-composer-plugin": "dev-topic-v3",
```
fballiano:
```suggestion
        "mahocommerce/maho-composer-plugin": "^3",
```

justinbeaty:
Sweet! This will be nice for the release notes!

fballiano:
I hope this project becomes successful because your amazing work deserves to be seen

PR #64, opened 2024-12-06
Fixed workflows failing when too many files are changed

justinbeaty:
Just realized the PHP syntax workflow fails because of too many files.

I didn't realize you could just output `${{ steps.changed-files.outputs.all_changed_files }}` directly into the shell script instead of an ENV var, but noticed that's how you did it in the copyright check.

Also added the ability to exclude certain files in the copyright check.

Pretty sure this should fix both.


fballiano:
super :-)

PR #66, opened 2024-12-09
PHPStan plugin v3 - raised PHPStan to level 6

justinbeaty:
# Description

The old PHPStan plugin had some problems: it reported errors that didn't actually exist, and ignored errors that were legitimate. Related discussions can be found here:
https://github.com/macopedia/phpstan-magento1/pull/6 ([archive](https://web.archive.org/web/20250313171104/https://github.com/macopedia/phpstan-magento1/pull/6)) 
https://github.com/OpenMage/magento-lts/pull/4411 ([archive](https://web.archive.org/web/20250313171747/https://github.com/OpenMage/magento-lts/pull/4411))
https://github.com/MahoCommerce/maho-phpstan-plugin/pull/3

We're start on the main branch with 1,687 errors in our baseline. Here's a list of commits with how many errors there are:

## Removed unused files (-8 errors)

Various files I came across during this PR that I couldn't find a purpose for. I could break this into a separate PR with smaller commits like `Removed Unused Permission Blocks`, and `Removed Unused lib/Varien/Convert/Validator`.

## Standardized instance getters such as `Mage::getModel()` and `Mage::helper()` (+113 errors)

This is the original https://github.com/OpenMage/magento-lts/pull/4411, but I re-ordered the methods *back to default* order to make review easier.

These are the core changes required to allow the new PHPStan plugin to work correctly. Basically, for each instance type (helper, model, block, etc) we need two methods: (1) a method to turn an alias such as `core/template` to `Mage_Core_Block_Template`, and (2) a method to return the instance.

We were missing the first method type for some instance types. Also there was inconsistency between what they do if it's an invalid class; some methods threw, some returned false. Finally, I tried to clean up docblocks, variable names, etc.

This adds so many errors because before `Mage::getResourceSingleton()` returned `object` instead of `Mage_Core_Model_Resource_Db_Collection_Abstract`, and that hides errors.

## PHPDoc: Deprecated redundant instance getters (+0 errors)

For some reason Mage devs added tons of `getModel`, `_getModel`, `_getHelper` methods to random classes instead of using `Mage::getModel`, etc. Besides potential issues with method signature mismatch, these methods are actually an extremely large problem for PHPStan. For example:

```php
class My_Module_Model_Foobar extends Mage_Core_Model_Abstract
{
    public function getProduct(): Mage_Catalog_Model_Product
    {
        // Error: Cannot call method loadByAttribute() on Mage_Core_Model_Abstract|false.    
        return $this->_getModel('catalog/product')->loadByAttribute(1, 'foo');
    }

    protected function _getModel(string $alias): Mage_Core_Model_Abstract|false
    {
        // But, PHPStan knows the below should be Mage_Catalog_Model_Product...
        return Mage::getModel($alias);
    }
}
```

PHPStan doesn't recursively enter other methods, so when it encounters `$this->_getModel()`, it sees that the method return `Mage_Core_Model_Abstract|false`, which is an error since the `loadByAttribute()` method isn't defined there.

Originally, I had to account for every one of these extra methods in the PHPStan plugin, but that was getting completely out of hand since there were more duplicated methods than there are original ones. I marked every single last one I could find as deprecated.

## PHPDoc: Undeprecated methods which are still in use with no alternative (+0 errors)

These methods were marked deprecated my Mage devs, but were still used in core with no apparent alternative. Not sure what else to do but remove the deprecated tag.

## PHPStan: Upgraded maho-phpstan-plugin and updated PHPStan config

Smallest commit, but I'm going to break this down into chunks:

1\. Install new `maho-phpstan-plugin` (+1  error)
All other settings stayed the same. The +1 error was actually something like -100 false positives, and +101 new legitimate errors. The two nearly canceling each other out was coincidental.
 
2\. Fix `paths` and `excludePaths` (+21 errors)
First, I added  `app/code/core/Maho` to the scan paths (actually just `app`) since we weren't scanning it.
Second, I removed everything I could from `excludePaths` while still allowing PHPStan to complete at this point in the commit history. We'll eventually end up with this section empty.
 
3\. Bump to [Level 5](https://github.com/phpstan/phpstan-src/blob/2.1.x/conf/config.level5.neon) (+1,112 errors)
The biggest change in this level (and why we have so many new errors) is the addition of `checkFunctionArgumentTypes`. Can't actually believe we weren't checking this before... For example, PHPStan has no problem with this code at Level 4:
```php
<?php declare(strict_types = 1);

$search = new stdClass();
$replace = 42;
$subject = true;

echo str_replace($search, $replace, $subject);
```
4\. Bump to [Level 6](https://github.com/phpstan/phpstan-src/blob/2.1.x/conf/config.level6.neon) (+3,767 errors)

Oof, so many errors... The change here is that it enforces types, either native PHP or via PHPDoc. However I think it's worth staying at Level 6 so that we don't add any more of them.

Note, by default this level added +10,361 errors.... So I disabled the check for everything except checking for parameters, returns, and properties, as well as ignore everything in the `lib` folder. Probably about 5,000 of those ignored errors are [No value type specified in iterable type](https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type). As I understand, this is caused by our functions returning things like `array` instead of something like `array<Mage_Catalog_Model_Product>`.

5\. Add `phpstan/phpstan-deprecation-rules` (+207 errors)
This makes PHPStan complain about any call to a deprecated method (or any usage of a deprecated variable). The main reason I installed this is so that the calls to the duplicated instance getters (i.e. `_getModel`) error out.

6\. Add bleedingEdge (+0 errors)
To make the eventual upgrade to PHPStan 3 easier :)

Note: for the next commits I'll stop listing the number of errors added or fixed, and will just do a count at the end.

## PHPStan Baseline: Replaced deprecated instance getters

Now that they're deprecated, we replace with the equivalent `Mage::getXxx` method.

## PHPStan Baseline: Made all helpers extend `Mage_Core_Helper_Abstract`

Some helpers didn't extend the abstract class.

## PHPStan Baseline: Mage_Adminhtml_Block_Store_Switcher

Fix deprecated calls to `getWebsiteCollection()`, `getGroupCollection()`, and `getStoreCollection()`.

## PHPStan Baseline: Mage_Admin OTPHP\TOTP

I guess `\OTPHP\TOTP::create()` was deprecated, changed to new methods.

## PHPStan Baseline: Mage_Admin and Mage_Api ACL

Instead of [removing these files](https://github.com/OpenMage/magento-lts/pull/4405/files#diff-e6e616c4c8db15f31fe6fae68b5ca5ba011b001d7501c1900cbc5ba5f2faac6d), I fixed them. Even though they're not really used, it seems like a nice feature and I'm pretty sure simply removing them would break the API.

## PHPStan Baseline: Mage_Core

All of the above are fixes in different modules. The naming isn't 100% accurate, as commits like the `Mage_Sales` one also contains stuff from `Mage_Checkout`. That's because sometimes changes would need to be fixed across modules.

## PHPStan Baseline: Misc

Other stuff...

## PHPStan Baseline: lib/Varien

Somewhat inspired by the files changed in https://github.com/OpenMage/magento-lts/pull/4405, but better. I'm pretty sure `lib/Varien/Directory/` isn't used in Mage, but some other varien files reference it so can't remove it yet.

## Update baseline

The moment of truth... We started with 1,687 errors in our baseline. Unfortunately we end up with 6,697 errors.

But, if I use the old config with the new plugin and all the fixes from this PR, we end up with 1,777 errors. So effectively only +90 errors, which isn't bad considering it added over +500 errors to OM.

# Comments

1\. I haven't done any testing other than to run PHPStan. I'll take some time to go through each changed class and test things.

2\. There is still one place that I'm not sure if it's possible to create a dynamic return extension for. That's usages of `$this->getLayout->getBlock()`. For example if you write `getBlock('head')`, PHPStan does not know what type of block this is, only that it's `Mage_Core_Block_Abstract|false`.

3\. I gave up trying to get this to work on OM back in December.  The autoloader situation is too different, and improving PHPStan was actually the reason I originally did the Composer v3 plugin, so that we could be 100% autoloaded and not need any PHPStan bootstrap file.


.github/workflows/phpstan.yml
```diff
@@ -8,21 +8,20 @@ on:
...
-    name: Analyze PHP ${{ matrix.php-versions }} (${{ matrix.refs == 'pr-head' && github.head_ref || github.base_ref || github.ref_name }})
+    name: Analyze PHP (${{ matrix.refs == 'pr-head' && github.head_ref || github.base_ref || github.ref_name }})
     runs-on: [ubuntu-latest]
     strategy:
       fail-fast: false
       matrix:
         refs: ['target-branch', 'pr-head']
         exclude:
           - refs: ${{ github.event_name != 'pull_request' && 'pr-head' }}
-        php-versions: ['8.2', '8.3', '8.4']
```
fballiano, line 19: question: why did you remove the older php versions?

justinbeaty: PHPStan v2 now has a [phpVersion min / max setting](https://phpstan.org/config-reference#phpversion).

So we put in `.phpstan.dist.neon`:

```neon
parameters:
    phpVersion:
        min: 80200
        max: 80400
```

Which means PHPStan doesn't need to run so many times, and it should be much quicker.

fballiano: oh right that's very nice to run phpstan only once!

justinbeaty:
FYI, I have to pull back a [few fixes](https://github.com/OpenMage/magento-lts/pull/4411/commits/30d89f7f9b3c60f4b1a4bcef70b8eaebd3db0b18) to this branch. I broke a few things...

.github/workflows/phpstan.yml
justinbeaty: Yup, well only twice now. One for the PR Head and one for the merge commit.

fballiano:
hey @justinbeaty, can we keep the .phpstan* naming as it was? I don't know why they use .dist but imho it's not correct as .dist should be the default (as in every linux software ever)

fballiano:
> > hey @justinbeaty, can we keep the .phpstan* naming as it was? I don't know why they use .dist but imho it's not correct as .dist should be the default (as in every linux software ever)
> 
> It's from the [PHPStan docs](https://phpstan.org/config-reference#config-file):

but the names I used also work

Screenshot 2025-03-13 alle 15 28 07

justinbeaty:
> say using the maho-starter

I'm not sure this applies to maho-starter. They'll need their own phpstan config file that they can name whatever they want. PHPStan won't pick up our `.phpstan[.dist].neon`.

fballiano:
exactly, that's why I didn't think overridability mattered

justinbeaty:
@fballiano I believe this is ready for review whenever you have time. I updated the first post with more info.

I think this PR should be Rebase & Merge for git bisect support since we change so much. But, my commit messages aren't great currently. Do you have any improvements?

```
Removed unused files
Standardized instance getters such as `Mage::getModel()` and `Mage::helper()`
PHPDoc: Deprecated redundant instance getters
PHPDoc: Undeprecated methods which are still in use with no alternative
PHPStan: Upgraded maho-phpstan-plugin and updated PHPStan config
PHPStan Baseline: Replaced deprecated instance getters
PHPStan Baseline: Made all helpers extend `Mage_Core_Helper_Abstract`
PHPStan Baseline: Mage_Adminhtml_Block_Store_Switcher
PHPStan Baseline: Mage_Admin OTPHP\TOTP
PHPStan Baseline: Mage_Admin and Mage_Api ACL
PHPStan Baseline: Mage_Catalog
PHPStan Baseline: Mage_Dataflow
PHPStan Baseline: Mage_Eav
PHPStan Baseline: Mage_Index
PHPStan Baseline: Mage_Sales
PHPStan Baseline: Mage_Wee
PHPStan Baseline: Mage_Wishlist
PHPStan Baseline: Mage_Core
PHPStan Baseline: Misc
PHPStan Baseline: lib/Varien
Update baseline
```

composer.json
```diff
@@ -7,6 +7,9 @@
         "AFL-3.0"
     ],
     "type": "maho-source",
+    "repositories": [
+        { "type": "git", "url": "https://github.com/justinbeaty/maho-phpstan-plugin.git" }
```
justinbeaty, line 11: TODO, remove

composer.json
```diff
@@ -61,9 +64,10 @@
     "require-dev": {
         "dealerdirect/phpcodesniffer-composer-installer": "^1.0.0",
         "friendsofphp/php-cs-fixer": "^3.4",
-        "mahocommerce/maho-phpstan-plugin": "^2",
+        "mahocommerce/maho-phpstan-plugin": "dev-topic-v3",
```
justinbeaty, line 67: TODO, change to "^3"

fballiano:
man where do I even start 😅😅 for sure everything is perfect, but I'll read everything but, question, would you like to merge it before tagging 25.3?

fballiano:
I guess that merging into 25.3 is a bit too early and there's no need to rush but this is amazing improvement in modernizing our whole stack.

The only thing I was thinking is about this: it's great to bump phpstan level to 6, but instead of excluding some checks/folders, why don't we just baseline those? I'm thinking about this in order to have more comprehensive checks on new PRs that we'll implement moving forward and have the most complete set of checks on the new code.

What do you think about that?

fballiano:
> I think this PR should be Rebase & Merge for git bisect support since we change so much.

we'll totally rebase&merge for sure

> my commit messages aren't great currently. Do you have any improvements?

I think they're ok, we'll probably have another commit to add the "v3" version bump and that will complete the circle :-)

justinbeaty:
> I guess that merging into 25.3 is a bit too early and there's no need to rush
Yeah, I'd for sure wait until after the release.

> instead of excluding some checks/folders, why don't we just baseline those?

| Description | Errors |
|--------|--------|
| All paths & errors | 13,334 |
| Cell | Cell |
| Cell | Cell |
| Cell | Cell | 

If I remove the entire `ignoreErrors` section, we go from 6,697 -> 13,334 errors. However, 6,010 of those errors are `missingType.iterableValue` https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type

This error would be solved by adding `/** @var ... */` docblocks at every foreach loop, but I don't want to promote that because there's a better way with generics that can probably solve the majority of these errors in one go.

Otherwise, I agree let's make sure to enforce parameter, return, and property types in `lib/`, see 008a2f5a84b96e35bfad12cb88d5d9313acf8475.

fballiano:
> If I remove the entire ignoreErrors section, we go from 6,697 -> 13,334 errors. However, 6,010 of those errors are missingType.iterableValue https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type

Is that making phpstan slower? because otherwise I still think that would be a good thing to enforce for new PRs

justinbeaty:
> Is that making phpstan slower? because otherwise I still think that would be a good thing to enforce for new PRs

I don't think it affects speed, but I'd like to take a look at using generics to hopefully eliminate the problem at a deeper level. Otherwise we'll be putting tons of useless `/** @var ... */` in the code just to appease PHPStan.

But yes, I would like that error checked, because PHPStan needs to know the type of objects in a loop.

fballiano:
will checkout this branch and test ASAP, sorry for having the "merge main branch" commit, I wanted to re-trigger the workflows.

would you like to re-force-push and then "rebase and merge"?

justinbeaty:
> would you like to re-force-push and then "rebase and merge"?

I actually tested this out yesterday, and the merge commits are no problem. When we rebase & merge those commits will disappear.

Test pull request with merge commit: https://github.com/justinbeaty/test-rebase-merge/pull/1 (also PR 2)

Commit history: https://github.com/justinbeaty/test-rebase-merge/commits/master/

fballiano:
shall I press the green button? (well actually it's gray 😅)

justinbeaty:
> shall I press the green button? (well actually it's gray 😅)

Let me do some final testing today, then we can merge.

justinbeaty:
I went through all the commits, double checked and tested each change. I made a [few fixes](https://github.com/MahoCommerce/maho/compare/46ef4b81825bf6923813758f6e6ff6fa8c304a90..c70693eb84c274a40ac88ff2f546a17f11e033e5). Otherwise, this is ready to be merged from my end.

fballiano:
I'll try to double check tonight and merge :-)

fballiano:
tests results:
- CLI
- normal navigation categories/cms/products
- cart(with actions)
- checkout
- myaccount
- wishlist
- backend (sales, catalog, customer)
- 2fa
- dataflow

I mean, it's difficult to test a large PR like this but I couldn't find anything wrong with it, and if it happens we'll fix it

fballiano:
also updated https://demo.mahocommerce.com

PR #68, opened 2024-12-12
Removed prototypejs from varien/product.js, varien/configurable.js and configurableswatches/swatches-product.js

fballiano:
It seems to me that everything is working correctly but since I had 1000 problems with these files I'd like for somebody else to give them a test.

**EDIT**: product page should work with or without prototypejs

@justinbeaty would you review this? :-)


justinbeaty:
Absolutely, I'll test first thing tomorrow.

Also nice on making a new `Template` class! That was the only prototypejs dependency I had left in part of the EAV code.

fballiano:
heheh good that you need that Template class :-) we can't really avoid having it :-)

public/skin/frontend/rwd/default/js/configurableswatches/swatches-product.js
```diff
@@ -14,84 +14,78 @@ window.addEventListener('load', function() {
...
-    return;
-};
+    resetChildren (element) {
+        return;
+    }
 
-Product.Config.prototype.resetChildren = function (element) {
-    return;
-};
+    onfigureForValues () {
```
justinbeaty, line 33:
```suggestion
    configureForValues () {
```

public/js/varien/product.js
```diff
@@ -5,310 +5,335 @@
...
         this.priceTemplate = new Template(this.config.template);
-        this.prices     = config.prices;
+        this.prices = config.prices;
 
-        this.settings.each(function(element){
-            Event.observe(element, 'change', this.configure.bind(this));
-        }.bind(this));
+        // Add change event listeners to all settings
+        this.settings.forEach(element => {
+            element.addEventListener('change', () => this.configure());
```
justinbeaty, line 27:
```suggestion
            element.addEventListener('change', (event) => this.configure(event));
```

justinbeaty: Also in `swatches-product.js`:
```diff
    onSelectChange(select) {
-        var attr = select;
+        var attr = select.attr;
```

I was getting an JS error when changing the size attribute of "Plaid Cotton Shirt", and the available colors weren't updating for the selected size. I.e. selecting 'XS' should make only Charcoal available.

justinbeaty:
The new Template class is breaking old usages of the class. Maybe we can make it compatible, or name it `MahoTemplate` until all other usages are converted.

image

fballiano:
> The new Template class is breaking old usages of the class. Maybe we can make it compatible, or name it `MahoTemplate` until all other usages are converted.

it should have been compatibile, at least for the product page it is, i've to investigate what happens with the backend

fballiano:
> With "Plaid Cotton Shirt", select a size, but don't select a color, and press "Add to cart".
> 
> image

it's weird cause I see it this way:
Screenshot 2024-12-12 alle 17 00 30

but it's better cause at least we're testing both situations

justinbeaty:
> it's weird cause I see it this way: Screenshot 2024-12-12 alle 17 00 30 
> but it's better cause at least we're testing both situations

I probably changed it at some point in the backend for size to not use a swatch is all.

justinbeaty:
> it should have been compatibile, at least for the product page it is, i've to investigate what happens with the backend

The problem is that majority of usages use the custom syntax: `/(^|.|\r|\n)({{(\w+)}})/` which has three capture groups instead of the default pattern's one capture group.

fballiano:
> > it should have been compatibile, at least for the product page it is, i've to investigate what happens with the backend
> 
> The problem is that majority of usages use the custom syntax: `/(^|.|\r|\n)({{(\w+)}})/` which has three capture groups instead of the default pattern's one capture group.

I'll redo it manually, AI was driving crazy with that conversion and was never working until I forced the single group... let's see

justinbeaty:
> I'll redo it manually, AI was driving crazy with that conversion and was never working until I forced the single group... let's see

I also have an idea for a fix I'm testing now.

justinbeaty:
@fballiano How about something like this? A couple of notes:

1) I changed the static properties to non-private. That's so we can access them like from outside. This will save us from repeating the "handlebars" pattern that is used quite a lot in the codebase. If you don't like that they could be modified from the outside, we could use property getters.

```js
new Template('...', Template.HANDLEBARS_PATTERN)
```

2) To make things easier, I convert all bracket to dot notation before looping through the parts. It works fine to access an array offset with a string, so no need for parseInt. It also means we can support nested arrays, i.e. `#{ foo[1][1] }`.

3) I rewrote `Template.create()` to not use `new Function()` since that's basically `eval`. For example with the old code:
```js
const tpl = Template.create('\` + alert(FORM_KEY) + \`')
tpl({ foo: 'bar' }); // Causes an alert
```

```js
class Template
{
    static DEFAULT_PATTERN = /(^|.|\r|\n)(#{(.*?)})/;
    static JAVASCRIPT_PATTERN = /(^|.|\r|\n)(\${(.*?)})/;
    static HANDLEBARS_PATTERN = /(^|.|\r|\n)({{(.*?)}})/;

    /**
     * Creates a Template object for string interpolation
     * @param {string} template - The template string
     * @param {RegExp} pattern - Optional custom pattern for replaceable symbols
     */
    constructor(template, pattern = Template.DEFAULT_PATTERN) {
        this.template = String(template);
        this.pattern = new RegExp(pattern, 'g');
    }

    /**
     * Evaluates the template with the provided data
     * @param {Record<string, any>} data - Object containing values for interpolation
     * @throws {Error} If data is null or undefined
     * @returns {string} Interpolated string
     */
    evaluate(data = {}) {
        return this.template.replaceAll(this.pattern, function() {
            const before = arguments[1] ?? '';        // The preceding character
            const symbol = arguments[2] ?? '';        // The entire symbol, i.e. "#{ foo }"
            const expr = (arguments[3] ?? '').trim(); // The expression, i.e. "foo"

            // Check if symbol was escaped
            if (before === '\\') {
                return symbol;
            }

            // Check if expression is empty
            if (expr === '') {
                return before;
            }

            // Convert bracket to dot notation
            const parts = expr.replaceAll(/\[(.*?)\]/g, '.$1').split('.');

            // Loop through each part and assign with null-safe property access
            let value = data;
            for (const part of parts) {
                value = value?.[part];
            }
            value ??= '';

            return before + value;
        });
    }

    /**
     * Creates a template using template literals
     * @param {string} template - Template string
     * @returns {Function} Template function
     */
    static create(template) {
        const t = new Template(template, Template.JAVASCRIPT_PATTERN);
        return t.evaluate.bind(t);
    }
}
```

fballiano:
how the heck do you also know all that javascript too? 😅 I've just committed your version, I've tested the frontend (product page with price variation on custom option) and the color attribute page (the one you posted before) and it's working fine.

I'm seeing this warning on the console:
Screenshot 2024-12-12 alle 19 00 36
but it's not really related

justinbeaty:
I've written a lot of Node JS in my lifetime... 😅 

And yes that `{{swatch}}` error is unrelated, I already have a fix for it in the EAV PR.

Continuing to test in a few minutes, but so far everything is looking good!

justinbeaty:
@fballiano I am not seeing any other errors and everything seems to be working!

fballiano:
badaboom let's merge this, I'm already working on checkout!

justinbeaty:
> badaboom let's merge this, I'm already working on checkout!

Nice! I am making really good progress on removing ExtJS too!

fballiano:
can't wait to drop that old #### too hiihhihiih

PR #70, opened 2024-12-14
Revert "Added JsonSerializable interface to Varien_Object (#55)"

### LINKED ISSUE #69: Backend: bundle product don't work
you can't modify a bundle product, when you hit save it shows a broken "bundle products" tab

loading a bundle product backend page shows like this:
Screenshot 2024-12-14 alle 16 05 04

justinbeaty:
Reverts #55, fixes #69.

While serializing nested data is nice, I had not taken into account circular references.

For example:

```php
$x = new Varien_Object();
$y = new Varien_Object();

$x->setY($y);
$y->setX($x);

json_encode($x, JSON_THROW_ON_ERROR);

// PHP Fatal error:  Uncaught JsonException: Recursion detected
```

I am not sure I see any good solution at the moment. While a class could define its own `jsonSerialize()` method to exclude any circular references, that's just not worth it.

### DIFF
--- a/lib/Varien/Object.php
+++ b/lib/Varien/Object.php
@@ -17,7 +17,7 @@
  * @category   Varien
  * @package    Varien_Object
  */
-class Varien_Object implements ArrayAccess, JsonSerializable
+class Varien_Object implements ArrayAccess
 {
     /**
      * Object attributes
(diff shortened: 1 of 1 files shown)


fballiano:
Thanks @justinbeaty, #69 now works perfectly ❤️

PR #71, opened 2024-12-15
Removed prototypejs from opcheckout.js and the whole frontend

fballiano:
This PR rewrites opcheckout.js (I've tested it and, as it is now, seems to be completely working) and **removes prototypejs from the frontend!!**

I can't see any other place where prototypejs is used (in the frontend) and I think it's time to remove it so that if there are some call left, we will see the errors and fix those (I've also checked all the "myaccount" and couldn't find any usage).

@justinbeaty would you take a look? 🙏🙏🙏


justinbeaty:
Sweet! I'll take a look first thing tomorrow.

Btw just so you know, you can also write those fetch statements like this. Just saves some indenting, etc. Browser support should be good.

```js
    async reloadStep(prevStep) {
        try {
            const response = await fetch(this.progressUrl + (prevStep ? `?prevStep=${prevStep}` : ''));
            if (!response.ok) {
                throw new Error('Network response was not ok');
            }

            const html = await response.text();
            const prevStepDiv = document.getElementById(`${prevStep}-progress-opcheckout`);
            if (prevStepDiv) {
                document.getElementById(`${prevStep}-progress-opcheckout`).innerHTML = html;
            }
            this.resetPreviousSteps();

        } catch (error) {
            this.ajaxFailure(error);
    }
```

justinbeaty:
@fballiano regarding CodeQL. How about put this into `public/js/varien/js.js`:

```js
/**
 * Alternative to PrototypeJS's string.stripTags() method
 */

function stripTags(str) {
    const div = document.createElement('div');
    div.innerHTML = str;
    return div.textContent;
}
```

fballiano:
@justinbeaty added the stripTags() function :-))) I'll take a look into the whole await/async which I don't really know 😅

public/skin/frontend/base/default/js/opcheckout.js
```diff
@@ -23,211 +24,256 @@ Checkout.prototype = {
...
-        for (var i = stepIndex; i < this.steps.length; i++) {
-            var nextStep = this.steps[i];
-            var progressDiv = nextStep + '-progress-opcheckout';
-            if ($(progressDiv)) {
-                //Remove the link
-                $(progressDiv).select('.changelink').invoke('remove');
-                $(progressDiv).select('dt').invoke('removeClassName','complete');
-                //Remove the content
-                $(progressDiv).select('dd.complete').invoke('remove');
+    }
```
justinbeaty, line 144: What do you think about merging the overridden method from here? https://github.com/MahoCommerce/maho/blob/main/public/skin/frontend/rwd/default/js/opcheckout_rwd.js

I see no harm in adding the `opc-has-progressed-from-login` class outside of rwd. But would just need to wrap the match media part since `bp` is defined in rwd's `app.js`.

```js
if (bp?.xsmall) {
    // Scroll viewport to top of checkout steps for smaller viewports
    if (window.matchMedia('(max-width: ' + bp.xsmall + 'px)').matches) {
        const checkoutSteps = document.getElementById('checkoutSteps');
        const topPosition = checkoutSteps.getBoundingClientRect().top + window.scrollY;
        window.scrollTo({ top: topPosition, behavior: 'smooth' });
    }
}
```

justinbeaty: And I think we can just use:
```js
checkoutSteps.scrollIntoView()
```

public/skin/frontend/base/default/js/opcheckout.js
```diff
@@ -23,211 +24,256 @@ Checkout.prototype = {
...
+
+    setStepResponse(response) {
         if (response.update_section) {
-            $('checkout-'+response.update_section.name+'-load').update(response.update_section.html);
+            const element = document.getElementById(`checkout-${response.update_section.name}-load`);
+            const range = document.createRange();
+            const fragment = range.createContextualFragment(response.update_section.html);
+            element.innerHTML = ''; // Clear existing content
+            element.appendChild(fragment);
```
justinbeaty, line 267: This is so `<script>` blocks execute, right? I created a helper function while removing ExtJS to do the same thing, but I actually like your version better.

Mine: https://github.com/justinbeaty/maho/blob/df443e0aff8aa46d1119de78aa6d4efecff3cf3a/public/js/varien/js.js#L71-L95

Yours with some slight modification:
```js
function updateElementHTML(targetEl, html, executeExternalScripts = false) {
    const range = document.createRange();
    const fragment = range.createContextualFragment(html);
    if (!executeExternalScripts) {
        fragment.querySelectorAll('script[src]').forEach(script => script.remove());
    }
    targetEl.replaceChildren(fragment);
}
```

The `executeExternalScripts` thing is because prototype by default doesn't execute `<script src="..."></script>` tags, so I wanted to mimic that behavior. Not sure any returned HTML from Magento contains external scripts though...

public/skin/frontend/base/default/js/opcheckout.js
justinbeaty: @fballiano I totally agree with merging everything into base. Ideally `rwd` is maybe just a stylesheet and possibly some extra JS.

fballiano: exactly, I still want to review everything to see if there are some useless classes or something like that to clean, but ideally RWD is base...

public/skin/frontend/base/default/js/opcheckout.js
```diff
@@ -521,298 +608,328 @@ ShippingMethod.prototype = {
...
         this.agreementsForm = agreementsForm;
-        this.onSave = this.nextStep.bindAsEventListener(this);
-        this.onComplete = this.resetLoadWaiting.bindAsEventListener(this);
-    },
+        this.isSuccess = false;
 
-    save: function(){
-        if (checkout.loadWaiting!=false) return;
+        this.nextStep = this.nextStep.bind(this);
+        this.resetLoadWaiting = this.resetLoadWaiting.bind(this);
```
justinbeaty, line 870: Is this supposed to be:
```js
        this.onSave = this.nextStep.bind(this);
        this.onComplete= this.resetLoadWaiting.bind(this);
```

public/skin/frontend/base/default/js/opcheckout.js
```diff
@@ -236,280 +282,321 @@ Checkout.prototype = {
...
+            this.form.addEventListener('submit', (event) => {
+                event.preventDefault();
+                this.save();
+            });
+        }
+
+        // Assuming you have a separate validation library
         this.validator = new Validation(this.form);
-        this.onSave = this.nextStep.bindAsEventListener(this);
-        this.onComplete = this.resetLoadWaiting.bindAsEventListener(this);
```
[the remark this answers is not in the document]
fballiano: mmmm no not really, but if you think they're useless let's remove them

public/skin/frontend/base/default/js/opcheckout.js
fballiano: should I put this new version in `js.js` as we did for the `stripTags`?

justinbeaty: Yeah that's where I put it while working on ExtJS since `js.js` is loaded on both frontend and backend. I have a feeling we will need to use it more once we eventually remove prototype from admin.

justinbeaty:
Regarding those  `onSave` and `onComplete` comments:

In both `class ShippingMethod` and `class Review` you should have:

```js
this.onSave = this.nextStep.bind(this);
this.onComplete = this.resetLoadWaiting.bind(this);
```

And then the fetch statements should be:

```diff
-            .then(this.nextStep)
+            .then(this.onSave)
            .catch(error => checkout.ajaxFailure.call(checkout, error))
-            .finally(this.resetLoadWaiting);
+            .finally(this.onComplete);
```

Basically, the `onSave` and `onComplete` methods don't exist in the class, but are placeholders that you can override. So for example in some other file:

```js
review.onSave = function(evt) {
    // Do some stuff
}
review.onComplete = function(evt) {
    // Do some stuff
}
```

This is done in the [gift message template](https://github.com/MahoCommerce/maho/blob/main/app/design/frontend/base/default/template/giftmessage/inline.phtml) for `billing` and `shipping` which you have correct, but some other modules might hook into `shippingMethod` or `review`.

fballiano:
> Regarding those `onSave` and `onComplete` comments:

done it (seems to work fine) in https://github.com/MahoCommerce/maho/pull/71/commits/3f167c9ba4b6ad2abb710325fcdd33f108a16ed0 :-))

public/skin/frontend/base/default/js/opcheckout.js
fballiano: should be ok now

public/skin/frontend/base/default/js/opcheckout.js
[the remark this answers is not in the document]
fballiano: if I'm not mistaken also this should be ok now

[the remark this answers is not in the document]
justinbeaty: Yup, both things are resolved now.

public/skin/frontend/base/default/js/opcheckout.js
fballiano: > The `executeExternalScripts` thing is because prototype by default doesn't execute `<script src="..."></script>` tags, so I wanted to mimic that behavior. Not sure any returned HTML from Magento contains external scripts though...

in the checkout there are a lot of js returned via ajax (cause they're inside phtml files) so, if that's ok with you, I think that the expected behavior is to execute the js, would you prefer to have it disabled?

fballiano: @justinbeaty done it in https://github.com/MahoCommerce/maho/pull/71/commits/069c5785d2804866fb0886ea9f015d9f8dc6feaf

and probably, for security reasons, it's better to keep the executeScripts as disabled

justinbeaty: What I meant is that prototype's `Ajax.Updater` only executes inline scripts, i.e. `<script>doFoo()</script>` and not external scripts like `<script src="https://..."></script>`.

I just figured instead of outright denying external scripts, we would give a parameter to control it. But to keep compatibility with prototype, you actually want to remove that `true` parameter.

Maybe the function should be named better, like `updateElementHTMLandExecuteScripts`, because if you *don't* want to execute scripts you can just keep using `element.innerHTML = '...'`

Hope that makes sense.

justinbeaty: Ref from http://api.prototypejs.org/ajax/Ajax/Updater/index.html

image

fballiano: @justinbeaty makes perfect sense, done in https://github.com/MahoCommerce/maho/pull/71/commits/6a76ca55f2b547a83fb4d5ac165fc8e50ebed7fc

I've used the updateElementHtmlAndExecuteScripts following this guideline:
Screenshot 2024-12-16 alle 17 48 29

justinbeaty: Yup, that function name is better 😃

public/skin/frontend/base/default/js/opcheckout.js
fballiano: @justinbeaty I've done it in https://github.com/MahoCommerce/maho/pull/71#discussion_r1887071236 but actually I've removed the breakpoint check, it seems to work fine anyway also for desktop (and it kinda makes sense to scrollIntoView also on desktop)

justinbeaty: Nice!

Btw, any idea why those two script tags are in the template file instead of XML? I noticed it in #56 where I had to put the translation messages using `<layout handle="checkout_onepage_index">` instead of referencing the script.

fballiano: yeah I really don't know, I tried not to think about it when I saw that hahaha

let's see if we can fix that

fballiano: @justinbeaty done it in https://github.com/MahoCommerce/maho/pull/71/commits/cdd828247f1092cbe8480225b2a50d22b5469bdc but since opcheckout is not skin-related anymore, shouldn't we remove it from skin?

justinbeaty: @fballiano Yeah, I think it's better in `public/js`. 

Also, you can now change [Mage/Checkout/etc/jstranslator.xml](https://github.com/mahocommerce/maho/blob/main/app/code/core/Mage/Checkout/etc/jstranslator.xml) to:

```xml
<jstranslator>
    <script path="mage/opcheckout.js" module="checkout">
        <!-- ... -->
    </script>
</jstranslator>
```

Of course replace `mage/opcheckout.js` with wherever you put the file (note: there's no leading `js/`)

That way the strings are included whenever `opcheckout.js` is included.

justinbeaty: Looks good to me! And yeah, we should definitely standardize the JS folder's layout at some point...

justinbeaty:
As a side note, this is how the login template looked before and after #34. That "checkout as guest" radio input made no sense if it was the only option, and the template file had stuff like:
```php
        <?php if( $this->getQuote()->isAllowedGuestCheckout() ): ?>
            <?php if( $this->getQuote()->isAllowedGuestCheckout() ): ?>
                <!-- ... -->
            <?php endif ?>
        <?php endif ?>
```

image

fballiano:
I think it's fine :-)

I still have to check the async/await thing

fballiano:
@justinbeaty I've rewritten all fetch with await/async, most of the times the method becomes a bit longer but I think it's easier to read :-)

justinbeaty:
> @justinbeaty I've rewritten all fetch with await/async, most of the times the method becomes a bit longer but I think it's easier to read :-)

Nice. :) 

Also my patch caused an issue with selecting register and checkout. Will fix shortly, just taking a quick lunch break...

justinbeaty:
I forgot that doing just `document.querySelector('input[name=checkout_method]')` would return the first input, even if it was not checked, so it was always choosing guest.

I changed it to this query selector. It gets the input that is either checked or hidden.

```js
const method = document.querySelector('input[name=checkout_method]:is(:checked,[type=hidden])')?.value;
```

Patch also fixes an `addEventListener` error when neither guest nor register methods are allowed.

Download link: [checkout_method_fix.patch](https://github.com/user-attachments/files/18157103/checkout_method_fix.patch)

fballiano:
pushed!

fballiano:
gave it a try and saw no problem

fballiano:
@justinbeaty I've created https://github.com/MahoCommerce/maho/issues/73 just not to forget about it (although I think it's kinda minor because of the importance of giftmessages ehehhe)

Do you think we can merge this one and toast to this milestone?

justinbeaty:
> @justinbeaty I've created #73 just not to forget about it (although I think it's kinda minor because of the importance of giftmessages ehehhe)
> 
> Do you think we can merge this one and toast to this milestone?

A bit more testing, I want to try out the braintree module.

fballiano:
I tried various paypal modules

justinbeaty:
I have one issue I'm tracking down with the braintree module, hang tight :)

justinbeaty:
I fixed a few things:

1\. I was able to cause a bug by modifying `OnepageController.php` like this:
```php
    public function saveMethodAction()
    {
        $this->getResponse()
            ->clearHeaders()
            ->setHeader('HTTP/1.1', '500 Internal Server Error');

        return;
````

Despite the error, I was able to proceed to the billing step. To fix, I ensured there is an `if (!response.ok)` check after each fetch call. 

2\. I made sure `this.ajaxFailure(error)` was always called with an argument because of the `alert()`. Else there could have been empty messages. As a side note, I just moved the `try {` block up a level in `setMethod()` to slightly simplify.

3\. I made sure `onSave` and `onComplete` were always called with an object that looks somewhat similar to Prototype's response object, i.e.:

```js
{
    status: 200,
    responseText: '...',
    responseJson: '...',
}
```

It was the lack of the `responseText` that was causing the braintree bug since the module hooks into the `onSave` event and wants to parse the json from `responseText`. I'm guessing it does that because Prototype only parses json with an `application/json` or similar header, and maybe Magento didn't used to return that header. The braintree module definitely needs updating since I had to re-enable prototype to make it work, but it's an easy change and might break less modules.

4\. Similarly, I made sure each `nextStep` method starts with the following which was on some of the methods but not all. 

```js
let response = transport.responseJSON || JSON.parse(transport.responseText) || {}
```

justinbeaty:
@fballiano let me know what you think, or feel free to change anything.

justinbeaty:
Hm, the braintree module's paypal option is actually broken too, because it does something like this:

```js
var origMethod = Review.prototype.initialize;

Review.prototype.initialize = function(saveUrl, successUrl, agreementsForm) {
    // Add in the PayPal buttons
    // ...

    // Call the original initialize method
    origMethod.apply(this, arguments);
}
```

But we no longer have an initialize function since that's prototype class stuff.

It can be fixed with this, and probably the same for the other steps in case other modules do something similar.

```js
class Review
{
    constructor() {
        this.isSuccess = false;
        this.onSave = this.nextStep.bind(this);
        this.onComplete = this.resetLoadWaiting.bind(this);

        this.initialize(...arguments);
    }

    initialize(saveUrl, successUrl, agreementsForm) {
        this.saveUrl = saveUrl;
        this.successUrl = successUrl;
        this.agreementsForm = agreementsForm;
    }

    // ...
}
```

However I'm not sure where to draw the line on BC support. It's a simple fix, but there must be a better way if you tested other PayPal modules that aren't broken.

I actually only use that Braintree module for the backend code as I have a custom frontend, but feel I should eventually make it compatible with Maho so people have options.

justinbeaty:
Maybe a more clean solution:

```js
class Review
{
    constructor(saveUrl, successUrl, agreementsForm) {
        this.saveUrl = saveUrl;
        this.successUrl = successUrl;
        this.agreementsForm = agreementsForm;
        this.isSuccess = false;
        this.onSave = this.nextStep.bind(this);
        this.onComplete = this.resetLoadWaiting.bind(this);

        this.initialize(...arguments);
    }

    initialize() {
        // Placeholder for 3rd party modules
    }

    // ...
}
```

fballiano:
mmmm yeah so... the paypal modules are in the core so no problem there because I fixed those but for 3rd party it's tricky.

yes we can add initialize() (to all classes or just Review?), that's pretty effortless, could a module override another method that we made incompatible? like the save()?

justinbeaty:
> yes we can add initialize() (to all classes or just Review?), that's pretty effortless, 

I would say all of the classes, since a module may want to hook into any of the steps, or even the main Checkout class. I've added a commit, hope it's okay. I wouldn't say we need to put these placeholder `initialize` methods everywhere we re-write a class in vanilla JS, but for checkout it seems worth for the extra compatibility.

> could a module override another method that we made incompatible? like the save()?

As long as they call the `onSave` and `onComplete` method afterwards, which they should be doing since that would have been required in Magento.

As as example I put this into one of the template files. It uses Prototype and adds a new param to the request, but could also send the request to another controller or do anything else. Of course it would only work with Prototype still enabled, but it could have been written with fetch, or XMLHttpRequest, or whatever else.

```js
Billing.prototype.save = function () {
    if (checkout.loadWaiting!=false) return;

    var validator = new Validation(this.form);
    if (validator.validate()) {
        checkout.setLoadWaiting('billing');

        // Add a new param
        var params = Form.serialize(this.form);
        params += '&myparam=1';

        new Ajax.Request(this.saveUrl, {
            method: 'post',
            onComplete: this.onComplete,
            onSuccess: this.onSave,
            onFailure: checkout.ajaxFailure.bind(checkout),
            parameters: params,
        });
    }
}
```

justinbeaty:
Otherwise, I've tested everything and I think it's good to merge and say goodbye to Prototype on the frontend. You've done absolutely incredible work, this is an incredible milestone!

fballiano:
@justinbeaty we merge? we merge? :-)))

justinbeaty:
> @justinbeaty we merge? we merge? :-)))

Yes, sorry I delayed it by one day, but hope you still celebrated last night!

justinbeaty:
Amazing!

fballiano:
> Yes, sorry I delayed it by one day, but hope you still celebrated last night!

hahahha don't be sorry, this is amazing and thank you for helping me with this awful task hahaahha 

i'm about to cry 🥹

fballiano:
https://demo.mahocommerce.com has been updated!

Issue #78, opened 2024-12-21
Webp format for cached images


empiricompany:
I'm testing the new feature that allows setting and enforcing a format for cached generated images.

this commit: https://github.com/MahoCommerce/maho/commit/36e9e93290d9bc0a40b4db4f0c7a7eb7e53a58a9

I've noticed that if I set the format to WebP, the generated images are fake WebP—they remain in JPG format. Specifically, they have the .webp extension, but if I save the image, it is recognized as JPG.

I tested the same image on the demo site, and when I save the image there, it is correctly recognized as webp. However, on my end, it is identified as a jpg with larger file sizes.

https://demo.mahocommerce.com/accessories/jewelry.html
![demo](https://github.com/user-attachments/assets/f6ee3f79-ece0-45b1-8d1a-801ab23aaeb5)
![sizewebp](https://github.com/user-attachments/assets/79f3aa94-df23-4baf-bd82-a12ddd67b043)

![demo](https://github.com/user-attachments/assets/04e09799-7bca-4246-bada-400605a999ec)
![testwebp](https://github.com/user-attachments/assets/dcd81479-1e8d-457d-8338-6e0772757c32)

fballiano:
weird, the webp support was already in OM since a while, I'll check it out

empiricompany:
Yes, I can successfully handle WebP images, upload them, and modify them, but only if they are natively in WebP format. It seems that no actual conversion to WebP takes place, and only the file extension is changed.

fballiano:
they get converted in the process of creating the thumbnails, but I will retest it

fballiano:
fixed by #79

PR #79, opened 2024-12-22
Issue 78 fix force format for cached images

empiricompany:
I think I have solved issue https://github.com/MahoCommerce/maho/issues/78
The same callback was being called for the output because it was based on the original image's mimetype, for example, imagejpeg. 
I forced the use of the correct gd2 function only for the output, while for the create function, it remains based on the original image's mimetype.

Now the image is saved in the correct format, and the file size is smaller as well.
ls
PS: I think the slight difference from demo version is due to the quality setting.


fballiano:
yes! it works, thanks for saving my Sunday cause I was already working on this 😂

PR #80, opened 2024-12-27
Fixed configurable swatches undefined array key warning

justinbeaty:
Go to https://store.example.com/ellis-flat.html and you'll receive warning in dev mode. Happens on OM main branch too.

The main fix was just to use `$mapping['labels'] ?? []`, but I cleaned up the method to use the splat operator instead of `call_user_func_array`, and simplified the `array_merge` / `array_map`.


fballiano:
weird I always have dev mode enabled, but thanks!

justinbeaty:
> weird I always have dev mode enabled, but thanks!

I think ellis flat just didn't have swatch specific images. I've never seen the error on any other configurable product.

PR #81, opened 2024-12-28
Replaced ExtJS with Maho Tree

justinbeaty:
It is finally here! Took me a bit longer than expected since ExtJS was used on more pages than I expected. I organized the commits for easier review / testing:

### Remove ExtJS, replace html5sortable with Sortable.js

Just library changes, no other code changes: removed ExtJS and html5sortable.js from `public/js`, and added Sortable.min.js.

### Misc library updates

Added some helper JS functions, fixed a bug in `maho-dialog.js` (`dialog.remove;` -> `dialog.remove();`), changed the backdrop style so stacked dialogs don't get darker, etc.

### Misc cleanup, CSS, and translations

Removed ExtJS from `head.phtml`, some layout files, cleaned up some CSS, and added new translations.

### Introduce maho-tree.js

Added the new MahoTree class + CSS. The JS file ended up being 900+ lines of code, but that's still small in comparison, and it does exactly what we need.
```
31.8 KB   maho-tree.js
4.5 KB    maho-tree.min.js.gz

150 KB    infinite-tree.min.js
29.7 KB   infinite-tree.min.js.gz
```

### Convert Roles / Permissions pages

Converts these pages. Be careful on the last one and don't lock yourself out of the admin panel like I did once!

https://store.example.com/index.php/admin/api_role/editrole/rid/1/
https://store.example.com/index.php/admin/api2_role/edit/id/1/
https://store.example.com/index.php/admin/api2_attribute/edit/type/guest/
https://store.example.com/index.php/admin/permissions_role/editrole/rid/1/

### Convert Product Reviews pages

Coverts these pages. They didn't use Ext Tree, but used Ext's AJAX lib.

https://store.example.com/index.php/admin/catalog_product_review/edit/id/16/
https://store.example.com/index.php/admin/catalog_product_review/new/

### Convert Insert Image Dialog

Converts the "Insert Image..." dialog.

https://store.example.com/index.php/admin/cms_page/edit/page_id/1/

### Improve categories tree recursion level

Some preliminary work: this fixes some of the mess in the category tree code. On the main branch if you go to Manage Categories, you'll notice every time you expand a category only one level deeper is loaded. However each time **four** levels deep were loaded from the database. This is pretty wasteful. I think Mage devs did this to work around several off-by-one errors. There were also some other redundant database operations I fixed as well.

**Note:** There are two method signatures I added the `$recursionLevel = null` parameter to, _so this is breaking change_. However I really doubt anyone has overridden these methods, and if they did the fix is to simply add the parameter to their signature.

### Move methods to Abstract block

More preliminary work: in true Mage fashion, there's sometimes not a great distinction between when methods are put in an Abstract class vs the one extending it.

In this case, generic tree methods were split between `Mage_Adminhtml_Block_Catalog_Category_Tree` and `Mage_Adminhtml_Block_Catalog_Category_Abstract`. The former is specifically for the Manage Categories page, but other pages with category trees seemingly chose at random which one to extend. Even worse, the abstract class called methods that didn't exist in it, and I suppose Mage devs just hoped the extending class implemented them. 

This commit makes the Abstract class the real base tree block. The only possible breaking change might be an `#[\Override]` needed where it wasn't before.

### Convert Manage Categories pages

Converts "Catalog > Manage Categories". This took the longest time and was by far the worst code I've ever seen in Magento. It had 800 lines of pure spaghetti JS split between three phtml files, and `CategoryController.php` returned even more JS which was evaluated in an iframe to call functions from the phtml files. It also used way too many session variables to try and keep track of state across AJAX calls. Many things were actually broken but maybe nobody ever noticed.

I rewrote `CategoryController.php` and made almost all actions both AJAX and non-AJAX compatible. So while AJAX is the default, you can set non-AJAX mode in layout xml with `<action method="setUseAjax"><flag>0</flag></action>`. I removed all session state variables and instead use JS's `history.replaceState()` function to update the URL when you change categories. I also improved error handling quite a bit. Then of course the entire JS was re-written into a nice class at `/public/js/mage/adminhtml/catalog/category.js`. 

There's obviously some breaking changes here, but nothing too bad. Mainly just that if someone had overwritten one of the three phtml files that the removal of ExtJS will break things. But since those files are much simpler now migration would be easy.

### Convert Manage Product pages

Converts "Catalog > Manage Products > Edit Product > Categories", and also the "Associated Products" tab for configurable products (replacing html5sortable). With the work done in the `Move methods to Abstract block` commit, I was able to simplify quite a bit here.

Only two possible breaking changes: I removed a controller file deprecated since `1.4.0.0-alpha2`, and removed the property `protected $_categoryIds;` that was totally unused. 

### Convert Url Rewrite Management pages

Converts "Catalog > URL Rewrite Management > Add URL Rewrite". Check both "For Category" and "For Product". The latter had a bug with lazy loaded categories where you can select a category that product is not in, causing an invalid rewrite. They also used grey text for both non active categories, and categories the product is not in, which was super confusing.

### Convert Widgets

Converts several pages:

1\. "CMS > Pages > Content > Insert Widget... > Select (Product | Category)" for both "Catalog Category Link" and "Product Link" types.

2\. "CMS > Widgets > Add Layout Update" when selecting "Display On > Anchor Categories / Non-Anchor Categories".  

3\. "Promotions > Catalog / Shopping Cart Price Rules" when selecting category as a condition.

### Convert Manage Attribute Set pages

Converts "Catalog > Attributes > Manage Attribute Sets > Edit". This took the second longest behind the Manage Categories page, but works really nice now. I also made it so the user is required to type `confirm` when deleting a set since it deletes all products from the set, ref: https://github.com/OpenMage/magento-lts/discussions/4364.

I fixed the `setIsReadOnly()` method so that you can hook into the `adminhtml_catalog_product_attribute_set_main_html_before` event and disable the edit form entirely. You will also be able to do this via layout xml once I integrate these changes to the EAV PR. It would be helpful for some store owners since messing with those sets can cause problems. Of course you could just disable access to the pages via permissions too.

I also removed a few useless files so it's a small breaking change, but again migration should be easy.


justinbeaty:
Question: is `maho-effects.js` still needed? I think I see the only usage of `Effect.Morph` in ExtJs.

fballiano:
This is my Christmas 🥹🥹

fballiano:
> Question: is `maho-effects.js` still needed? I think I see the only usage of `Effect.Morph` in ExtJs.

It was introduced to replace prototype effects or something like that, but only used by ExtJS so you can nuke it :-)

fballiano:
I'm in the middle of some other development but I'll review this ASAP :-)

justinbeaty:
> I'm in the middle of some other development but I'll review this ASAP :-)

I see a new branch was created, so if it's that then awesome!

> It was introduced to replace prototype effects or something like that, but only used by ExtJS so you can nuke it :-)

Will do in another PR as I see a few other stray references, but this PR already removes the loading of it from layout xml files.

fballiano:
> > I'm in the middle of some other development but I'll review this ASAP :-)
> 
> I see a new branch was created, so if it's that then awesome!

yep it is, it seems to be quite easy (at least the basic functionality for the backend), I should be able to commit something today

fballiano:
So, I love when we manage to throw away ancient stuff and make everything lighter, and this PR is exactly that :-)

I've a weird suggesting, why for ACL we don't just remove the icons at all? I never really understood the "document" icon next to the "leaves" and the "folder" ones are really just decorative because you can click the +/- symbols anyway...

Screenshot 2024-12-29 alle 00 38 24

it would be many less DOM nodes and IMHO it's easier on the eyes

justinbeaty:
I changed the expand / collapse buttons to monochrome. I had tried the Maho green and some other colors which didn't look great, but somehow I never considered grey, haha.

I also pushed a change that let's us disable the icons where needed, you can do:

```js
new MahoTree(this.config.treeDiv, {
    noIcons: true,
})

// or per node:

new MahoTreeNode(tree, {
    icon: false,
    // ...
})
```

But I think in many places the icons help, even if the specific icons chosen don't make 100% sense:

### Empty folders

Gives some extra feedback that you did expand the folder, and it's just empty. It's hard to see the difference on a static image, but you can see the folder icon change when clicking it.

image

### Loading icon

This only shows if the network request takes longer than the `Loading Indicator Timeout` admin setting (200 ms), but you can probably see it if you hit the "Expand All" button on manage categories.

image

### Attribute set

We have different icons for system attributes, configurable attributes, and user defined attributes. Since you can only remove the last type, it's helpful to know that before trying to drag over to the unassigned column. If you wanted to put all configurable attributes in one group, you'd have no way of knowing what ones those were without trying to unassign every attribute. While the gear vs wrench isn't super intuitive, users should figure it out once they try to move one of them. I think the folders icons also help here to distinguish from group vs attribute.

image

### Catalog product link widget

Not totally necessary, but it does give a visual distinction between the categories and the "View All Products" icons.

image

### Permissions

I agree the paper icon doesn't make sense, but one case where I think it helps is when a folder is in the "indeterminate" state because not all of its children are checked or unchecked. Without the folder icon, it's feels (to me) not quite clear what the checkbox means.

image

### Future cases?

One other scenario I can think of is if we ever have a tree that combined categories and products, like below, but we don't have this scenario now so I'm not weighing it very heavily.

```
├── Accessories
│   ├── Aviator Sunglasses
│   └── Jackie O Round Sunglasses
└── Men
    └── ...
```

public/js/maho-tree.js
```diff
@@ -0,0 +1,935 @@
...
+
+        this.uniqId = generateRandomString(6);
+
+        this.config = {
+            selectable: false,
+            sortable: false,
+            lazyload: false,
+            rootVisible: true,
+            noLeafNodes: false,
+            noIcons: false,
```
fballiano, line 71:
```suggestion
            leafNodes: true,
            icons: true,
```

could we use positive configurations I find it very confusing to have a "no-something" set to true

maybe we could call them "showLeafNodes" and "showIcons" if icons/leafNodes are not clear enough?

justinbeaty: Yeah, good call. I think the variant with "show" is better. Will also rename `rootVisible` to `showRoot` for consistency.

fballiano:
> I changed the expand / collapse buttons to monochrome. I had tried the Maho green and some other colors which didn't look great, but somehow I never considered grey, haha.

I really like the new +/- icons, actually I was also referring to all other icons as well 😅 (I'm always a bit too drastic I know) because they may look more modern and more adaptable to different color scheme (of the whole backend) but actually what I think it's more important it's that the "folder" icons to me don't look super clear, what I mean is that I can't really distinguish between the closed and opened state because the difference is kinda minor.

> But I think in many places the icons help, even if the specific icons chosen don't make 100% sense:

ok since it's backend I'm not against having icons, although i'm always worried about the million DOM nodes we have (google complains about it) but that's mainly important for the frontend

public/js/maho-tree.js
justinbeaty: Actually, I'm a bit hung up on the leaf node option naming. Is `treatAllNodesAsFolders` too wordy?

fballiano: IMHO if wordy is easier to understand then we should go with wordy :-)

fballiano:
> ### Permissions
> I agree the paper icon doesn't make sense, but one case where I think it helps is when a folder is in the "indeterminate" state because not all of its children are checked or unchecked. Without the folder icon, it's feels (to me) not quite clear what the checkbox means.
> 
> image

maybe in this case there could be folder icons and no leaves icons? dunno if it makes sense to have that distinction but, to me, if the icon doesn't represent the element, it makes my brain think "why is that a document icon? this is not a document, is that an error?" which IMHO is confusing to the user.

but again I wouldn't make a big deal out of it anyway

justinbeaty:
> I really like the new +/- icons, actually I was also referring to all other icons as well 😅 (I'm always a bit too drastic I know) because they may look more modern and more adaptable to different color scheme (of the whole backend) but actually what I think it's more important it's that the "folder" icons to me don't look super clear, what I mean is that I can't really distinguish between the closed and opened state because the difference is kinda minor.

As a test I put `filter: grayscale(1);` on the icons in CSS, and I don't hate it. Does it look okay to you?

image

fballiano:
you're right about disabled categories, although actually I was thinking about something like heroicons instead when talking about black&white icons heehhe

Screenshot 2024-12-29 alle 22 16 51

but again, this is more something to think about but probably, for a first implementation, it's ok to have the colored icons too, as you fell more appropriate :-)

justinbeaty:
The folder icon is actually the only one that has color. Well also the leaf one is very slightly blue.

But if we go full line icons we would have to do stuff like this too:

image





Also one more test, putting greyscale only on disabled ones:

image

justinbeaty:
I'm also really not against monochrome line icons btw, like I will absolutely no qualms if we decide to change it.

fballiano:
maybe I'd leave the colored icons for the moment and think about the whole icon thing in a next release ehhehe

justinbeaty:
> maybe I'd leave the colored icons for the moment and think about the whole icon thing in a next release ehhehe

Yeah, sounds good. It would be nice too if we could control the icon color with a CSS var, but you need inline SVG or an icon font for that I believe.

---

One other thing, I was removing some prototype from some other files, and I realize I didn't take into account when Magento sends an ajax expired response when the user is logged out (from cookie expired or whatever).

Instead of checking for this everywhere, I wonder if a small helper function in `public/js/mage/adminhtml/tools.js` is worthwhile? Something like:

```js
async function mahoFetch(url, options) {
    const response = await fetch(url, options);
    if (!response.ok) {
        throw new Error(Translator.translate('Server returned status %s', response.status));
    }
    const result = await response.json();
    if (result.error) {
        throw new Error(result);
    } else if (result.ajaxExpired && result.ajaxRedirect) {
        return setLocation(response.ajaxRedirect);
    }
    return result;
}
```

fballiano:
yeah probably a mahoFetch would have helped in the frontend too... and if it was the same then it could have been in `js.js` but now I don't know if it's worth converting the frontend...

justinbeaty:
It seems the `ajaxExpired` thing is specifically only on the admin side, but could still find a way to make it generic for both I think.

Probably not worth it now for the frontend anyway like you said, but I'm sure we will always be improving the JS side so in time...

fballiano:
then to me you can put it in js.js and we'll see in the future if we want to bring it to the frontend :-)

justinbeaty:
Added `mahoFetch`. Also a new `MahoError` class that translates messages, although could be called `TranslatedError` or something, but maybe that could conflict with some other library? Naming things is hard... 😅 

https://github.com/MahoCommerce/maho/blob/612d691ad6473fd13531b28ed4e36a822875fd68/public/js/varien/js.js

Will test everything again tomorrow.

fballiano:
In "manage categories" page, when clicking "expand all" I got:

Screenshot 2024-12-30 alle 10 36 07

It happens also when clicking on the "+" icon of a leaf node, hope it helps

app/design/adminhtml/default/default/layout/catalog.xml
```diff
@@ -32,7 +32,8 @@
     <adminhtml_catalog_product_new>
         <update handle="editor"/>
         <reference name="head">
-            <action method="addItem"><type>js</type><name>html5sortable.min.js</name><params/></action>
+            <action method="addJs"><name>Sortable.min.js</name></action>
```
fballiano, line 35: would it be awful to rename it to lowercase? for some reason case-sensitive names for static assets make me uncomfortable 😅 although it should never be a problem, I don't know, a single uppercase name seems like a sore thumb

public/js/mage/adminhtml/tools.js
```diff
@@ -754,3 +755,74 @@ function copyText(event) {
...
+    setMessageDivHtml('');
+}
+
+/**
+ * Set a message in <div id="messages"></div>
+ *
+ * @param {string} message - text value of the message to display
+ * @param {string} type - one of `success|error|notice`
+ */
+function setMessageDiv(message, type = 'success') {
```
fballiano, line 772:
```suggestion
function setMessagesDiv(message, type = 'success') {
```

since the target div is #messages?

public/js/mage/adminhtml/tools.js
```diff
@@ -754,3 +755,74 @@ function copyText(event) {
...
+    setMessageDivHtml(`<ul class="messages"><li class="${type}-msg"><ul><li><span>${message}</span></li></ul></li></ul>`);
+}
+
+/**
+ * Raw function to update <div id="messages"></div>
+ *
+ * @param {string} html
+*/
+function setMessageDivHtml(html) {
```
fballiano, line 781:
```suggestion
function setMessagesDivHtml(html) {
```

public/js/mage/adminhtml/tools.js
```diff
@@ -754,3 +755,74 @@ function copyText(event) {
...
+    return url.toString();
+}
+
+/**
+ * Set query params, i.e. ?id=1
+ *
+ * @param {string} url - the base URL
+ * @param {Object} params - key value pairs to add, update, or remove
+ */
+function setQueryParams(url, params = {}) {
```
fballiano, line 818: would it make sense to you to prefix all new functions with `maho`? The thing is, when looking for a maho-javascript-something in an IDE, if we have a prefix, it's much easier, and we also avoid possible name collisions?

app/design/adminhtml/default/default/layout/catalog.xml
justinbeaty: Fixed. I didn't like the uppercase either.

public/js/mage/adminhtml/tools.js
justinbeaty: Created #84 for us to discuss 😃

fballiano: but then I'd probably merge this PR before moving all the javascripts around, since this is already working?

justinbeaty: Agreed, I just have a few changes to push today.

app/code/core/Mage/Adminhtml/controllers/Catalog/ProductController.php
```diff
@@ -559,6 +559,7 @@ public function validateAction()
             $response->setMessage($this->getLayout()->getMessagesBlock()->getGroupedHtml());
         }
 
+        $this->getResponse()->setHeader('Content-type', 'application/json', true);
         $this->getResponse()->setBody($response->toJson());
```
fballiano, line 563: maybe this could also use the setBodyJson?

justinbeaty: How did you review it so fast, haha. You're right, I missed that one.

justinbeaty:
Okay, couple of final changes:

1\. I updated mahoFetch so that it now adds the form key and `?isAjax=true` automatically. It also shows the loader by default so we don't have to manually show / hide before every ajax request. This makes it work like prototype's Ajax.Request.

2\. Some maho dialog improvements such as prevent body scrolling when active, fixing button colors, and executing scripts in the content html. I think the last thing is okay by default, but actually we could put it behind a config option if you'd like.

3\. I figured I might as well rewrite some of the JS classes without prototype.

4\. Some other bug fixes, some of the issues present in OM.

I tested everything again and all seems to be working fine.

Also, regarding the copyright action, this should work. It was just the extra `\( \)` around the 2024 part.
```
grep -E -q "Copyright \(c\) (2024-)?${CURRENT_YEAR} Maho \(https://mahocommerce\.com\)"
```

PR #83, opened 2024-12-28
Added 2FA optional support for backend

fballiano:
This is the flow from a user perspective

1. from the backend you enter your profile
Screenshot 2024-12-28 alle 18 43 13

2. now you have a 2FA subsection
Screenshot 2024-12-28 alle 18 43 31

3. if you enable it it shows the QR to scan
Screenshot 2024-12-28 alle 18 43 38

4. enabling 2FA and saving then you get the new login
Screenshot 2024-12-28 alle 18 43 53

**NOTES:**
- This PR modifies a lot of public methods, which I don't like, but it seems correct (to me) to complete the `login` method only after login and 2FA are verified, do you have better ideas?


app/code/core/Mage/Adminhtml/Helper/TwoFactorAuthentication.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public function getQRCodeUrl(#[\SensitiveParameter] string $username, #[\SensitiveParameter] string $secret): string
+    {
+        $storeName = Mage::getStoreConfig('general/store_information/name');
+        $otp = \OTPHP\TOTP::create($secret);
+        $otp->setLabel($username . '@' . $storeName);
+        $otp->setIssuer('Maho Admin');
+
+        return $otp->getQrCodeUri(
+            'https://api.qrserver.com/v1/create-qr-code/?color=000000&bgcolor=FFFFFF&data=[DATA]&qzone=2&margin=0&size=300x300&ecc=M',
```
fballiano, line 31: I'm using this online service to generate the QR because I wouldn't want to deal with images to save and then delete etc... is that ok?

justinbeaty:
Neat! Testing now.

One thing I'm not so crazy about is the call to `api.qrserver.com` to generate the code, since that leaks your secret key. OTPHP mentions https://github.com/Bacon/BaconQrCode in their docs, which has only one dependency, and can also generate SVGs.

app/code/core/Mage/Adminhtml/Helper/TwoFactorAuthentication.php
justinbeaty: Just mentioned that point in a comment. I think the SVG route could allow us to not have to save any files.

fballiano:
The amount of dependencies is always a tricky thing, but in this case I think this feature is pretty important nowadays and should be provided by the core so... we have to add them.

PR: just removed qrserver.com and replaced it with an inline SVG which works file ;-)

justinbeaty:
> The amount of dependencies is always a tricky thing, but in this case I think this feature is pretty important nowadays and should be provided by the core so... we have to add them.
> 
> PR: just removed qrserver.com and replaced it with an inline SVG which works file ;-)

Nice, pulling now. Too many deps of course isn't great, but it at least shouldn't get loaded by composer on any other page than this one.

I also forgot to hit the "Flush Cache Storage" button at first and wondered why I got a fatal error, even though I have cache disabled. I forgot there's still some sort of schema cache saved.

justinbeaty:
Here's a patch that uses `adminhtml/widget_form_element_dependence` which lets you get rid of the `system/account/edit/form/js.phtml` file.

I could push the commit, but maybe you're working on something now so didn't want to interfere. :)

```diff
diff --git a/app/code/core/Mage/Adminhtml/Block/System/Account/Edit/Form.php b/app/code/core/Mage/Adminhtml/Block/System/Account/Edit/Form.php
index 91c51057e5..96b4b2c07b 100644
--- a/app/code/core/Mage/Adminhtml/Block/System/Account/Edit/Form.php
+++ b/app/code/core/Mage/Adminhtml/Block/System/Account/Edit/Form.php
@@ -106,20 +106,19 @@ class Mage_Adminhtml_Block_System_Account_Edit_Form extends Mage_Adminhtml_Block
             'label' => Mage::helper('adminhtml')->__('Enable 2FA'),
             'name' => 'twofa_enabled',
             'values' => Mage::getModel('adminhtml/system_config_source_yesno')->toOptionArray(),
-            'onchange' => 'toggleTwoFactorSetup(this.value)',
         ]);
 
         if (!$user->getTwofaEnabled()) {
             $secret = $user->getTwofaSecret();
-            if (!$user->getTwofaSecret()) {
+            if (!$secret) {
                 $secret = $twoFactorAuthenticationHelper->getSecret();
                 $user->setTwofaSecret($secret)->save();
             }
 
             $twoFactorAuthenticationFieldset->addField('twofa_qr', 'note', [
                 'label' => Mage::helper('adminhtml')->__('QR Code'),
-                'text' => $twoFactorAuthenticationHelper->getQRCode($user->getUsername(), $secret) . '<br/><br/>' .
-                    Mage::helper('adminhtml')->__('Scan this QR code with your authenticator app (Google Authenticator, Authy, etc.)'),
+                'text' => $twoFactorAuthenticationHelper->getQRCode($user->getUsername(), $secret),
+                'note' => Mage::helper('adminhtml')->__('Scan this QR code with your authenticator app (Google Authenticator, Authy, etc.)'),
             ]);
 
             $twoFactorAuthenticationFieldset->addField('twofa_verification_code', 'text', [
@@ -140,12 +139,12 @@ class Mage_Adminhtml_Block_System_Account_Edit_Form extends Mage_Adminhtml_Block
 
         $this->setForm($form);
 
-        $this->setChild(
-            'form_after',
-            $this->getLayout()
-            ->createBlock('adminhtml/template')
-            ->setTemplate('system/account/edit/form/js.phtml'),
-        );
+        /** @var Mage_Adminhtml_Block_Widget_Form_Element_Dependence $block */
+        $block = $this->getLayout()->createBlock('adminhtml/widget_form_element_dependence');
+        $block->addFieldDependence('twofa_qr', 'twofa_enabled', '1')
+            ->addFieldDependence('twofa_verification_code', 'twofa_enabled', '1');
+
+        $this->setChild('form_after', $block);
 
         return parent::_prepareForm();
     }
```

app/code/core/Mage/Adminhtml/Helper/TwoFactorAuthentication.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        return \OTPHP\TOTP::create()->getSecret();
+    }
+
+    public function getQRCode(#[\SensitiveParameter] string $username, #[\SensitiveParameter] string $secret): string
+    {
+        $storeName = Mage::getStoreConfig('general/store_information/name');
+        $otp = \OTPHP\TOTP::create($secret);
+        $otp->setLabel($username . '@' . $storeName);
+        $otp->setIssuer('Maho Admin');
```
justinbeaty, line 28: I didn't have a store name set, so it showed as `Maho Admin: admin@`

```suggestion
        $otp = \OTPHP\TOTP::create($secret);
        $otp->setIssuer('Maho Admin');
        if ($storeName = Mage::getStoreConfig('general/store_information/name')) {
            $otp->setLabel($username . '@' . $storeName);
        } else {
            $otp->setLabel($username);
        }
```

justinbeaty: Maybe also an image field? It's non standard, but some apps support it:

```php
$otp->setParameter('image', 'https://placehold.co/400x400.png');
```

Note I couldn't even find a publicly hosted Maho logo in PNG format since everything is SVG. 😅

justinbeaty:
One more patch, encrypts the secret in the db:

```diff
diff --git a/app/code/core/Mage/Admin/Model/User.php b/app/code/core/Mage/Admin/Model/User.php
index 986b847155..78a105371f 100644
--- a/app/code/core/Mage/Admin/Model/User.php
+++ b/app/code/core/Mage/Admin/Model/User.php
@@ -65,6 +65,8 @@
  * @method string getRpTokenCreatedAt()
  * @method $this setRpTokenCreatedAt(string $value)
  * @method $this setUserId(int $value)
+ * @method int getTwofaEnabled()
+ * @method $this setTwofaEnabled(int $value)
  */
 class Mage_Admin_Model_User extends Mage_Core_Model_Abstract
 {
@@ -793,4 +795,28 @@ class Mage_Admin_Model_User extends Mage_Core_Model_Abstract
         $absoluteMinLength = Mage_Core_Model_App::ABSOLUTE_MIN_PASSWORD_LENGTH;
         return ($minLength < $absoluteMinLength) ? $absoluteMinLength : $minLength;
     }
+
+    /**
+     * Retrieve unencrypted value of twofa_secret
+     */
+    public function getTwofaSecret(): ?string
+    {
+        if ($this->hasData('twofa_secret')) {
+            return Mage::helper('core')->getEncryptor()->decrypt($this->_getData('twofa_secret'));
+        }
+    }
+
+    /**
+     * Store encrypted value of twofa_secret
+     */
+    public function setTwofaSecret(#[\SensitiveParameter] ?string $secret): self
+    {
+        if ($secret === null) {
+            $this->setData('twofa_secret', null);
+        } else {
+            $encrypted = Mage::helper('core')->getEncryptor()->encrypt($secret);
+            $this->setData('twofa_secret', $encrypted);
+        }
+        return $this;
+    }
 }
```

app/code/core/Mage/Admin/Model/Session.php
```diff
@@ -134,13 +136,10 @@ protected function logoutIndirect()
 
     /**
      * Try to login user in admin
-     *
-     * @param  string $username
-     * @param  string $password
      * @param  Mage_Core_Controller_Request_Http $request
```
justinbeaty, line 139:
```suggestion
-     * @param  Mage_Core_Controller_Request_Http $request
```

?

justinbeaty:
This is amazing, incredible work!! I've tested and it works great. Let me know if you want me to push changes.

justinbeaty:
Also can remove `app/design/adminhtml/default/default/template/system/account/edit/form/js.phtml` now :)

app/code/core/Mage/Adminhtml/Helper/TwoFactorAuthentication.php
fballiano: @justinbeaty about the image, does it work for you? cause on Authy it's not working so I don't know, I wouldn't add it also because maybe people would like to have their own store logo?

justinbeaty: The image does work on FreeOTP which I downloaded to test out. On Authy I get some weird image, I assume you get the same. Apparently they use a result from Google Image Search. It seems they have a way to ask their support team to use the proper logo, but that has nothing to do with code.

I think even on a customer's store, Authy is still going to google search "Maho Admin" since that's set as the issuer. Maybe if we made it:

```php
if ($storeName = Mage::getStoreConfig('general/store_information/name')) {
    $otp->setIssuer($storeName);
} else {
    $otp->setIssuer('Maho Admin');
}

$otp->setLabel($username);
```

fballiano: done! :-)

justinbeaty:
I can retest everything if you like, otherwise it all looks good! Again this is super awesome!!

fballiano:
> I can retest everything if you like, otherwise it all looks good! Again this is super awesome!!

I've tested it now and it seems to work ok so everything should be safe to merge

The implementation is not super amazing I think... I had to modify quite some public methods which is never nice, also the flow is not perfect... because you have to login 2 times (since the 2FA verification code appears in the login form after the first login) but I wouldn't get the headache of creating a 2fa-only page because it would mean saving the half completed authentication somewhere and it would be such a pain.

This version should be safe from a security point of view, although for example you're not prompted the 2FA code to disable 2FA and this kind of nice-to-have things.

But if you think it looks ok I'm fine merging it ehehe

justinbeaty:
I think it's good enough. Changing the public methods is okay here I think because if someone overwrote the login method they *should* implement the 2fa logic in their method too. So getting the error will warn them. At least you still need a password to disable 2fa, so it's not the most terrible.

Maybe the separate page could be solved with AJAX instead. I.e. on login first we call to see if 2fa is enabled for the user and then hide the username password fields and show the 2fa one.

fballiano:
> Maybe the separate page could be solved with AJAX instead. I.e. on login first we call to see if 2fa is enabled for the user and then hide the username password fields and show the 2fa one. That way all three fields are sent on the actual login, but it looks like two pages.

I thought about something similar but I was very worried to introduce some vulnerability anywhere and this way the 2FA is checked only if the user already provided a correct password instead of just checking if 2FA is enabled for a user when they type just the username. Did I overthink it?

justinbeaty:
> > Maybe the separate page could be solved with AJAX instead. I.e. on login first we call to see if 2fa is enabled for the user and then hide the username password fields and show the 2fa one. That way all three fields are sent on the actual login, but it looks like two pages.
> 
> I thought about something similar but I was very worried to introduce some vulnerability anywhere and this way the 2FA is checked only if the user already provided a correct password instead of just checking if 2FA is enabled for a user when they type just the username. Did I overthink it?

Let me take a quick look.

justinbeaty:
Had to stop for the night, but I think I have a solution for this in the morning.

Will also respond to the comments on the other PRs then too.

justinbeaty:
I pushed a commit for testing, but if it's not okay then we can revert. It definitely makes things more complicated, so let's consider it carefully.

Basically it adds a new prelogin endpoint that we POST to first. The endpoint validates the username + password and only then if 2fa is needed will return the response to go to step 2. If there is any other failure, it will proceed to submit the login form as usual.

justinbeaty:
Force pushed one more fix. Seriously if this makes things too complicated or whatever we can revert.

fballiano:
Hey @justinbeaty, I was testing and I put a wrong 2FA code just to see what would happen and now the login doesn't work anymore, seems like the wrong 2FA is saved in session or something like this, is it possible?

justinbeaty:
> Hey @justinbeaty, I was testing and I put a wrong 2FA code just to see what would happen and now the login doesn't work anymore, seems like the wrong 2FA is saved in session or something like this, is it possible?

Do you have the latest commit? It probably made it more complicated that I force-pushed, but you should be able to do:

```sh
git reset --hard HEAD~1
git pull
```

fballiano:
mmm yes I have it but I still get stuck, not even resetting cookies and removing the form_key from the url :-\

justinbeaty:
> mmm yes I have it but I still get stuck, not even resetting cookies and removing the form_key from the url :-\

Also try hard refresh if not already? There was a bug in varienForm class I had to fix too.

fballiano:
not even from a new incognito window, wth :-\

justinbeaty:
Hm, I can reproduce it now. I had use secret key in url disabled.

justinbeaty:
Wait, I actually can't reproduce it... :/

Does it still work if you go back a commit?

fballiano:
If I go back 1 commit (which is now my last merge) it works :-(

fballiano:
but if I pull and try to login I get:

Screenshot 2024-12-29 alle 16 24 19

which comes from this:
Screenshot 2024-12-29 alle 16 25 44

but if I remove that setMessage() I get:
Screenshot 2024-12-29 alle 16 26 19

justinbeaty:
Hm, seems the "You did not sign in correctly..." message is meant to hide the exact reason why it failed, not sure if that's exactly what we want, but we could fix that to show the "2FA verification code is invalid"  instead.

But I don't understand why it is failing for you. Do you want us to try and debug, or just revert?

fballiano:
but the failed 2FA is from a previous try, dunno why it gets stuck there, the "hiding" of the real message from the exception could be a good thing (it was already there). I'll try to debug it, let's see

justinbeaty:
> but the failed 2FA is from a previous try, dunno why it gets stuck there

The secret is still in your db right? I noticed one line before my commit that I thought could potentially wipe it:

```php
# in Mage_Admin_Model_User

if ($this->getTwofaEnabled() && $this->getTwofaSecret()) {
    if ($twofaVerificationCode) {
        if (!Mage::helper('adminhtml/twoFactorAuthentication')->verifyCode($this->getTwofaSecret(), $twofaVerificationCode)) {
                Mage::throwException(Mage::helper('adminhtml')->__('2FA verification code is invalid.'));
        }
        $this->setTwofaSecret(null); // This line
    }
}
```

> the "hiding" of the real message from the exception could be a good thing (it was already there).

We could only "unhide" the 2fa error. But like you said maybe it's good to hide all errors.

fballiano:
yes the secret is still there

justinbeaty:
> yes the secret is still there

Can you check the final login POST payload? Might need "Preserve log" checked. I'm wondering if it could be auto-fill weirdness with the hidden step 1.

image

fballiano:
oh right!!!! it is a stupid thing!! I press enter instead of clicking on ok!!!!!!!!!!!!

justinbeaty:
> oh right!!!! it is a stupid thing!! I press enter instead of clicking on ok!!!!!!!!!!!!

🤦‍♂️ okay this is varien form being dumb I think... let me fix.

justinbeaty:
Okay, it should be fixed now. I also removed both prototype and varien form from the page, leaving only validation.js

fballiano:
it works perfectly now ❤️

ideally I'd like varienForm to manage this situation but at the same time... does the world need what varienForm does in 2024? most of the features seem revolving around validation anyway...

so to me we can merge this as it is because the 2FA experience is now much better than my implementation (as usual when you step in) 😊

justinbeaty:
I think the varien submit handler wasn't even being called at all. It really was just setting up the validation object. Also I'm not sure why, but we have two varien form classes (one for the frontend and one for backend).

One more change, I just added the `autocomplete=""` attributes, especially since `one-time-code` is a supported option.

I may have improved it, but adding this feature was never even on my radar in the first place. 😄

fballiano:
I'll quickly retest and merge :-)))

justinbeaty:
I am testing again too.

There's one error sort of related, but it happens on OM too. If you fail to sign in then refresh the page you keep getting invalid form key because the page keeps sending the POST request. Browsers used to warn about form resubmission but I guess they don't anymore...?

fballiano:
yep I see that but I wouldn't try to fix it now especially if it happens also in OM.

I retested and everything seems to work fine so let's mergeeeee, weekend project for the win :-)))

fballiano:
hiihih ok then let's add it :-D

app/design/adminhtml/default/default/template/login.phtml
```diff
@@ -35,26 +33,95 @@
...
+                            <input type="submit" class="form-button" value="<?= Mage::helper('core')->quoteEscape(Mage::helper('adminhtml')->__('Login')) ?>" title="<?= Mage::helper('core')->quoteEscape(Mage::helper('adminhtml')->__('Login')) ?>" />
+                        </div>
                     </div>
                 </div>
                 <p class="legal">Powered by Maho</p>
             </form>
             <div class="bottom"></div>
-            <script type="text/javascript">
-                var loginForm = new varienForm('loginForm');
+            <script type="module">
```
fballiano, line 67: question, why type module?

fballiano: it seems to work also with type text/javascript too

app/design/adminhtml/default/default/template/login.phtml
```diff
@@ -35,26 +33,96 @@
...
+                history.replaceState(null, null, window.location.href);
```
fballiano, line 68: this way?

app/design/adminhtml/default/default/template/login.phtml
justinbeaty: Just creates a new scope, that way nothing can overwrite the methods, and we have no possibility of something like `const validator` clashing with another variable defined elsewhere. You probably saw I used that a lot in the ExtJS PR.

app/design/adminhtml/default/default/template/login.phtml
justinbeaty: ya, it seems to be working for me. Now on a refresh it's a regular GET instead of POST.

app/design/adminhtml/default/default/template/login.phtml
fballiano: I realize I don't know anything anymore 😂

fballiano:
badabim bam boom

app/design/adminhtml/default/default/template/login.phtml
justinbeaty: All this stuff that you couldn't use a few years ago is now is widely supported, I had to catch up since I haven't done frontend devel in some time.

Basically my thought was to use module where possible, that way you can do `const form = new VarienForm()` and not worry about whatever other scripts are on the page.

Sometimes we need stuff in the global scope though, cause Magento likes to make it's form buttons use `onclick="myFunction()"`. So I either did `window.myFunction = function() {}` or just used a regular script tag. HTML5 also lets us remove the `type="text/javascript"` part too.

We should probably standardize this all, but for now there's just sooo many inline script tags....

fballiano: > Basically my thought was to use module where possible, that way you can do `const form = new VarienForm()` and not worry about whatever other scripts are on the page.

it's a nice idea for sure

> We should probably standardize this all, but for now there's just sooo many inline script tags....

yeah.....

justinbeaty:
image

Btw, this is how it looks in FreeOTP. 😎

fballiano:
hhiihih yes I tested it yesterday and uploaded that logo :-D

justinbeaty:
> hhiihih yes I tested it yesterday and uploaded that logo :-D

I saw you uploaded, just didn't know if you checked in freeotp. Anyway, this is such a great addition! Pretty sure lots of merchant accounts require it for compliance too.

fballiano:
hehehe yes I tested it with FreeOTP, I was curious :-)

I thought about adding 2FA (but not too the same level of PITA of M2 ahahha) cause it wasn't the first time I heard people saying they wanted it, I'm glad you liked the idea :-)

PR #85, opened 2024-12-31
 Added `$this->getResponse()->setBodyJson()` method

justinbeaty:
In #81, I copied the `_prepareDataJSON` method as defined in OnepageController to some other controllers, but I am finding more places it will be useful.

Also, pretty much every controller that send back JSON isn't setting the content type correctly. Not that it matters much, but it should be correct.

### DIFF
--- a/app/code/core/Mage/Core/Controller/Response/Http.php
+++ b/app/code/core/Mage/Core/Controller/Response/Http.php
@@ -99,4 +99,14 @@ class Mage_Core_Controller_Response_Http extends Zend_Controller_Response_Http
         $this->sendHeaders();
         exit;
     }
+
+    /**
+     * Prepare JSON formatted data for response to client
+     */
+    public function setBodyJson($response): self
+    {
+        $this->setHeader('Content-type', 'application/json', true);
+        $this->setBody(Mage::helper('core')->jsonEncode($response));
+        return $this;
+    }
 }
--- a/app/code/core/Mage/Core/Controller/Response/Http.php
+++ b/app/code/core/Mage/Core/Controller/Response/Http.php
@@ -7,7 +7,7 @@
  * @package    Mage_Core
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright  Copyright (c) 2019-2024 The OpenMage Contributors (https://openmage.org)
- * @copyright  Copyright (c) 2024 Maho (https://mahocommerce.com)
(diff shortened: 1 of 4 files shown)


justinbeaty:
Actually, one other possibility could be to define it here:

```php
abstract class Mage_Core_Controller_Varien_Action
{
    // ...

    protected function _sendJson($response) {
        $this->getResponse()->setHeader('Content-type', 'application/json', true)
            ->setBody(Mage::helper('core')->jsonEncode($result));
    }

    protected function _sendJsonError($message, $extra = []) {
        $this->_sendJson([ 'error' => true, 'message' => $message, ...$extra ]);
    }
}
```

And controllers use:

```php
$this->_sendJson([ 'foo' => 'bar' ]);
$this->_sendJsonError('my error message', [ 'some_extra_data' => 'foobar' ]);
```

I think I like the original implementation better though, especially since we don't have to worry about controllers defining their own implementations.

The only benefit of this solution is that it standardizes that an error object should have `error=true` and a message key.

app/code/core/Mage/Core/Controller/Response/Http.php
```diff
@@ -99,4 +99,14 @@ public function sendHeadersAndExit(): never
         $this->sendHeaders();
         exit;
     }
+
+    /**
+     * Prepare JSON formatted data for response to client
+     */
+    public function setBodyJson($response): self
```
fballiano, line 106:
```suggestion
    public function setBodyJson($response): self
```

can $response have a type?

justinbeaty: You're right, but it should be `mixed` since that's what the core helper method accepts:

https://github.com/MahoCommerce/maho/blob/119aeb38714d3a0fadb6aece9c2c8118ec6dd697/app/code/core/Mage/Core/Helper/Data.php#L685-L693

We could choose a more restrictive type, but it should at least include array, object, varien object, and null. But it's also valid to encode strings, numbers, and probably other types too, so it's basically just mixed anyway.

fballiano: let's go with mixed :-)

justinbeaty: I actually hadn't realized that mixed was a valid type in PHP8 until now. Will update later today, unless you want to do.

I've been testing the PR in various places and it works well.

fballiano: done, and if it's all green I'll merge it right away

fballiano:
the copyright action doesn't work but whatever :-D

PR #87, opened 2025-01-04
Rewrote WYSIWYG without PrototypeJS

justinbeaty:
Cleans up:

* WYSIWYG instances, both on CMS > Pages / Static Blocks, and the popup on Catalog > Manage Products / Categories.
* Insert Variables... popup
* Translate Inline popup


app/code/core/Mage/Core/controllers/AjaxController.php
```diff
@@ -26,14 +27,14 @@ public function translateAction()
...
-        //filtering
         /** @var Mage_Core_Model_Input_Filter_MaliciousCode $filter */
         $filter = Mage::getModel('core/input_filter_maliciousCode');
         foreach ($translation as &$item) {
             $item['custom'] = $filter->filter($item['custom']);
         }
 
         $response = Mage::helper('core/translate')->apply($translation, $area);
+        $this->getResponse()->setHeader('Content-type', 'application/json', true);
         $this->getResponse()->setBody($response);
```
fballiano, line 38: could we use setBodyJson here? that's my motto lately :-D

fballiano:
tested and couldn't find any fault

app/code/core/Mage/Core/controllers/AjaxController.php
fballiano: I see, could the setBodyJson check if it's a string return it as it is, otherwise json_encode it?

justinbeaty: There's one edge case to consider -- it's actually valid to JSON encode a regular string, so `'foo'` is encoded to `'"foo"'`.  If someone passed a string that wasn't already valid JSON, then we would get an error when trying to parse it. For example:

```js
JSON.parse('foo');
// Error: Unexpected token 'o', "foo" is not valid JSON
```

But I like the idea, and I think that one edge case isn't huge enough to outweigh the consistency we'd gain.

fballiano: could we use json_validate to avoid that edge case? which actually would kinda bother me

justinbeaty: Perfect, yeah that will absolutely work. Should we do this in another PR after this is merged? Cause I can convert the instances I had to use setHeader manually in the ExtJS PR too.

fballiano: sure, why not.

also, json_validate is php8.3+ so.. either we introduce a polyfill or we update system requirements... i'm fine with both actually

justinbeaty: I was wondering how I never knew of that function! Looks like 8.2 just ended active support at the start of the year, although it still has security releases for two more years. Would be an easy enough polyfill though.

PR #89, opened 2025-01-04
Updated `$this->getResponse()->setBodyJson()` to accept already encoded string value

justinbeaty:
See: https://github.com/MahoCommerce/maho/pull/87#pullrequestreview-2530773606

Will do some quick testing, but nothing should break.

Btw, the removal of `_prepareDataJSON` in SetController was from the ExtJS PR, I forgot to remove it once we added the `setBodyJson` method.

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Catalog/Category/WidgetController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Catalog/Category/WidgetController.php
@@ -51,8 +51,7 @@ class Mage_Adminhtml_Catalog_Category_WidgetController extends Mage_Adminhtml_Co
             Mage::register('category', $category);
             Mage::register('current_category', $category);
 
-            $this->getResponse()->setHeader('Content-type', 'application/json', true);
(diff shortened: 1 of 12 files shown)


justinbeaty:
So I think PHPStan fails for all versions actually because of `phpVersion: 80200` in the neon file, and that overrides the PHP version. I recall now I had a fix for that in #66, need to finish that...

fballiano:
we can use https://github.com/symfony/polyfill-php83/blob/1.x/Php83.php as json_validate polyfill and it should fix everything ;-)

justinbeaty:
> we can use https://github.com/symfony/polyfill-php83/blob/1.x/Php83.php as json_validate polyfill and it should fix everything ;-)

Yeah, sounds good. Do you want to make a commit bumping our minimum version and include the polyfill? Then we can merge into this branch?

fballiano:
yep, just a sec

fballiano:
let's keep 8.2 at the moment with symfony's polyfill and let's see :-)

Issue #86, opened 2025-01-04
New Maho Category Tree: "+" Sign Displayed for Folders Without Children


empiricompany:
First of all, a huge thank you to the amazing work of @justinbeaty and all the heroic developers of this project!  

I’m testing the new category tree as a replacement for the ExtJS libraries.

I've encountered an issue:
In the category tree, folders without children are still displayed with a "+" sign.

![tree](https://github.com/user-attachments/assets/532b9576-644b-4f4f-acf7-09a9c5c0100e)
![product-tree](https://github.com/user-attachments/assets/9bb0b749-f9e4-4112-9400-aac9449228a0)

justinbeaty:
Hi Tony, there's two reasons why this is the case:

1\. Magento used to load practically the entire tree from the database every time you expanded a folder, but it only returned one level at a time in the AJAX response. While this is super inefficient, it was able to tell if a node had children even though those children weren't included in the response. I changed this so that all nodes that are loaded are returned, but this means we are unable to tell beforehand if the deepest node has any children. This issue could be solved by returning just one level less than is loaded from the database, which would at least still be more efficient than what Magento was doing. 

2\. In the Manage Categories page, an empty node could become non-empty because of drag and drop, or vice-versa. This issue is a bit more complicated since we would have to keep track of some more state, such as when a node is expanded with children, but you drag the only child node elsewhere. I know ExtJS did something like this, but it was buggy. For example in certain cases you'd see the expand button, click it, and then the expand button would disappear because there are no child nodes.

Ultimately I figured for consistency that if a node was a folder type, then it should have the expand button. It depends on how much this bothers you.

empiricompany:
Hi @justinbeaty,
I tried to fix the issue, but I’m not that skilled with JavaScript.
While working on the code, I realized how much work and effort you’ve put into it (kudos to you!), but it’s important for me to maintain consistent behavior. Otherwise, the client might see the "+" sign and click it multiple times, thinking there’s an issue loading the (non-existent) children.

Would it be possible to sacrifice the performance and load the entire tree as before? Would this solve or help resolve the issue?

empiricompany:
Maybe I’ve fixed it! Tested on Firefox and Chrome.

```css
.maho-tree details:not(:has(ul > li)) summary::before {
    background: none !important;
}
```

justinbeaty:
The CSS only solution unfortunately won't work when you have more than 4 levels deep. 

[...]

PR #90, opened 2025-01-05
Hide plus sign if folder is empty

empiricompany:
Attempt to fix issues https://github.com/MahoCommerce/maho/issues/86


fballiano:
this is not working properly, lev3 has children

Screenshot 2025-01-09 alle 12 21 27

fballiano:
Fixed by #93

PR #92, opened 2025-01-10
New feature: added Order Status color

fballiano:
This PR allows for this:

Screenshot 2025-01-10 alle 16 47 26

Screenshot 2025-01-10 alle 16 47 30

Screenshot 2025-01-10 alle 16 47 36

### DIFF
--- a/app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
+++ b/app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
@@ -16,7 +16,7 @@ $installer->startSetup();
 
 $installer->run("
     ALTER TABLE {$this->getTable('sales/order_status')}
-    ADD COLUMN `color` VARCHAR(20) NULL,
+    ADD COLUMN `color` VARCHAR(20) NULL
 ");
 
 $installer->endSetup();
--- a/app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
+++ b/app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
@@ -1,0 +1,22 @@
+<?php
+
+/**
+ * Maho
+ *
+ * @category   Mage
+ * @package    Mage_Core
+ * @copyright  @copyright  Copyright (c) 2025 Maho (https://mahocommerce.com)
+ * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+/** @var Mage_Checkout_Model_Resource_Setup $this */
+$installer = $this;
+
+$installer->startSetup();
+
+$installer->run("
+    ALTER TABLE {$this->getTable('sales/order_status')}
+    ADD COLUMN `color` VARCHAR(20) NULL,
+");
+
+$installer->endSetup();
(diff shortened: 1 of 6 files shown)


app/code/core/Mage/Adminhtml/Block/Sales/Order/Grid.php
```diff
@@ -244,4 +245,29 @@ public function getGridUrl()
...
+            }
+
+            if ($colors) {
+                Mage::register('order_status_colors', $colors);
+            }
+        }
+
+        if ($colors) {
+            $color = $colors[$row['status']] ?? '';
+            return "<span style='display:inline-block;width:15px;height:15px;vertical-align:text-bottom;background:{$color}'></span> {$value}";
```
fballiano, line 268: should we add this to `override.css`?

justinbeaty:
Will review shortly!

app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+/** @var Mage_Checkout_Model_Resource_Setup $this */
+$installer = $this;
+
+$installer->startSetup();
+
+$installer->run("
+    ALTER TABLE {$this->getTable('sales/order_status')}
+    ADD COLUMN `color` VARCHAR(20) NULL,
```
justinbeaty, line 19:
```suggestion
    ADD COLUMN `color` VARCHAR(20) NULL
```

I had a SQL error with the trailing comma.

app/code/core/Mage/Adminhtml/Block/Sales/Order/Grid.php
```diff
@@ -244,4 +245,29 @@ public function getGridUrl()
     {
         return $this->getUrl('*/*/grid', ['_current' => true]);
     }
+
+    public function decorateStatus($value, $row, $column, $isExport): string
+    {
+        $colors = Mage::registry('order_status_colors');
```
justinbeaty, line 251: Could we alternatively use a protected variable `$this->orderStatusColors` since it's only being used in this block?

It's actually really not a big deal either way, just a thought.

justinbeaty:
I like it! Only real change is the install script error, rest is just nit-picks that in no way blocks merging this as is.

app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
fballiano: absolutely, since I was using the same script name for another something I'm working so I manually created the column

fballiano:
Tomorrow I'll fix the remaining things, I thought this PR was a nice quick addition since a customer I was working on had this as a custom module and I've seen it used many times

app/code/core/Mage/Adminhtml/Block/Sales/Order/Grid.php
fballiano: done ;-)

app/code/core/Mage/Adminhtml/Block/Sales/Order/Grid.php
fballiano: no no, I think it should be done, and it's done

justinbeaty:
Retested -- looks good!

PR #93, opened 2025-01-10
MahoTree Improvements

justinbeaty:
Related #86 and #90 

Empty nodes now no longer have the `[+]` expand icon. 

Clicking a folder now expands / collapses it for `mode='radio'` type trees (i.e. "Manage Categories" and "Insert Image"). I don't think it works well with non-radio modes where clicking again also deselects the node. But in radio mode clicking again would otherwise do nothing. @fballiano, you can remove that check in `maho-tree.js` on line 505 to see how it would work on all types and see if you'd prefer it on all trees.

Renamed `hideInputs` -> `showInputs` and fixed usages. I missed this during initial PR [(ref)](https://github.com/MahoCommerce/maho/pull/81#pullrequestreview-2525010369).

Made the CMS tree detect if there were any children, else the `[+]` sign shows up initially, then clicking it makes it disappear. This is how ExtJS worked, but I didn't really like it.


fballiano:
tested it and it is amazing as usual

PR #97, opened 2025-01-15
New feature: passkey authentication for backend

fballiano:
This PR adds basic passkey support to the backend.

Screenshot 2025-01-15 alle 22 35 53

and in the login page:

Screenshot 2025-01-15 alle 22 36 34

Passkeys only work if the website is under SSL, that's also necessary in order to be able to test this PR.


justinbeaty:
Cool! Checking it out now.

app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
```diff
@@ -11,12 +11,12 @@
...
-
 $installer->startSetup();
 
 $installer->run("
-    ALTER TABLE {$this->getTable('sales/order_status')}
-    ADD COLUMN `color` VARCHAR(20) NULL
+    ALTER TABLE {$this->getTable('admin/user')}
+    ADD COLUMN `passkey_credential_id_hash` varchar(255) NULL,
+    ADD COLUMN `passkey_public_key` VARCHAR(255) NULL
 ");
```
justinbeaty, line 20: Was this left over code from testing?

justinbeaty:
Do you mind if I push a few changes? I'd like to remove the inline JS from `Mage_Adminhtml_Block_System_Account_Edit_Form` and put into a separate JS file and a few other small improvements.

app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
```diff
@@ -11,12 +11,12 @@
...
 $installer = $this;
-
 $installer->startSetup();
 
 $installer->run("
-    ALTER TABLE {$this->getTable('sales/order_status')}
-    ADD COLUMN `color` VARCHAR(20) NULL
+    ALTER TABLE {$this->getTable('admin/user')}
+    ADD COLUMN `passkey_credential_id_hash` varchar(255) NULL,
+    ADD COLUMN `passkey_public_key` VARCHAR(255) NULL
```
fballiano, line 19:
```suggestion
    ALTER TABLE {$this->getTable('sales/order_status')}
    ADD COLUMN `color` VARCHAR(20) NULL
```

fballiano:
> Do you mind if I push a few changes? I'd like to remove the inline JS from `Mage_Adminhtml_Block_System_Account_Edit_Form` and put into a separate JS file and a few other small improvements.

@justinbeaty feel free! absolutely :-)

app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
fballiano: It got broken from a `git stash` situation 😅 now it should be ok

public/js/mage/adminhtml/tools.js
```diff
@@ -701,7 +701,21 @@ var Base64 = {
...
+        const bytes = new Uint8Array(binaryString.length);
+        for (let i = 0; i < binaryString.length; i++) {
+            bytes[i] = binaryString.charCodeAt(i);
+        }
+        return bytes.buffer;
+    },
+
+    fromArrayBuffer: function(buffer) {
+        return btoa(String.fromCharCode(...new Uint8Array(buffer)))
+            .replace(/\+/g, '-').replace(/\//g, '_').replace(/=/g, '');
```
fballiano, line 717: do we want this in tools.js since (if i'm not mistaken) it's only used in account.js?

justinbeaty:
Couple of changes, as always all are just suggestions and we can modify / revert any of them.

1\. I moved the base64 methods to the already existing Base64 JS object.
2\. Moved the main code to a new class at `public/js/mage/adminhtml/system/account.js`, it takes a config object with URLs.
3\. Used the `$this->getButtonHtml()` method for generating the buttons.

I'm actually not sure yet how to test this, it seems I'll need a hardware key as I don't have one built into my computer.

public/js/mage/adminhtml/tools.js
justinbeaty: I figured better to keep it with the other Base64 stuff, but certainly can revert that change.

fballiano:
> I'm actually not sure yet how to test this, it seems I'll need a hardware key as I don't have one built into my computer.

I can use it with my password manager without any problem

app/code/core/Mage/Adminhtml/Block/System/Account/Edit/Form.php
```diff
@@ -97,6 +97,38 @@ protected function _prepareForm()
...
+        ]);
+
+        if ($user->getPasskeyCredentialIdHash()) {
+            $fieldset->addField('passkey_status', 'note', [
+                'label' => Mage::helper('adminhtml')->__('Status'),
+                'text'  => '<span class="grid-severity-notice"><span>' .
+                    Mage::helper('adminhtml')->__('Passkey Registered') . '</span></span>',
+            ]);
+            $fieldset->addField('passkey_status', 'note', [
+                'text'  => $this->getButtonHtml($this->__('Remove Passkey'), null, 'delete', 'remove-passkey-btn'),
```
fballiano, line 111: I keep wondering how I don't know these basic stuff 🤣

fballiano:
@justinbeaty what do you think about the questions at the end of the PR description?

app/code/core/Mage/Adminhtml/Block/System/Account/Edit/Form.php
justinbeaty: I knew there was a way, but had to grep to find the exact method.

justinbeaty:
> @justinbeaty what do you think about the questions at the end of the PR description?

Let me look closer, also see if I can figure out how to make it work with bitwarden.

app/code/core/Mage/Adminhtml/controllers/System/AccountController.php
```diff
@@ -25,6 +25,11 @@ class Mage_Adminhtml_System_AccountController extends Mage_Adminhtml_Controller_
      */
     public const ADMIN_RESOURCE = 'system/myaccount';
 
+    protected $_publicActions = [
+        'passkeyregisterstart',
+        'passkeyregistersave',
+    ];
+
```
fballiano, line 32: @justinbeaty do you know something about these?

justinbeaty: Doc comment says "Array of actions which can be processed without secret key validation", so I think it's for any action that you're not logged in for. It checks for that `/key/<random_string>/` part in the URL.

 I think that since you are logged in on this page you don't need it, but maybe instead use `_forcedFormKeyActions`

fballiano: I'll retest everything (making this work was a bit cumbersome) and try to remove those 2 actions from that array

public/js/mage/adminhtml/tools.js
```diff
@@ -701,7 +701,21 @@ var Base64 = {
             }
         }
         return string;
-    }
+    },
+
+    toArrayBuffer: function(base64) {
+        const binaryString = atob(base64);
```
justinbeaty, line 707:
```suggestion
        const binaryString = this.mageDecode(base64);
```

Because the base64 from the controller is in Magento's special format.

Or, I think we could not replace the chars here:
```php
            $challenge = strtr(base64_encode(random_bytes(32)), '+/', '-_');
```

app/code/core/Mage/Adminhtml/controllers/System/AccountController.php
```diff
@@ -102,4 +107,195 @@ public function saveAction()
...
+    public function passkeyregisterstartAction()
+    {
+        try {
+            $user = Mage::getSingleton('admin/session')->getUser();
+            if (!$user) {
+                throw new Exception('Not authenticated');
+            }
+
+            $challenge = strtr(base64_encode(random_bytes(32)), '+/', '-_');
```
justinbeaty, line 119:
```suggestion
            $challenge = base64_encode(random_bytes(32));
```

justinbeaty:
I figured out the problem with bitwarden, I had my local install URL on the ignore list because I was tired of it asking to update the password when testing the 2fa code. Just fixed an error where I had used the same field name in my previous commit.

justinbeaty:
One thing regarding `setBodyJson`:

```php
            $this->getResponse()
                ->setHttpResponseCode(400)
                ->setBodyJson(['error' => $e->getMessage()]);
```

The `setHttpResponseCode(400)` will actually make it so the error message isn't read by `mahoFetch`, because it fails the `if (!response.ok)` check.

But I feel not sending a 200 code is better, so probably better to fix mahoFetch, right?

fballiano:
> But I feel not sending a 200 code is better, so probably better to fix mahoFetch, right?

yep I think mahoFetch should be able to manage a 4xx error

justinbeaty:
> > But I feel not sending a 200 code is better, so probably better to fix mahoFetch, right?
> 
> yep I think mahoFetch should be able to manage a 4xx error

Okay, it was an easy fix. I just checked if the JSON response contains an error key first, and only if it doesn't then will we return the `Server returned status %s` message.

fballiano:
> One other question, are all of the `Invalid request method`, `Not authenticated`, `Missing required fields`, `Invalid challenge`, `Invalid origin`, `Credential already registered to another user`, etc errors meant to be visible to the user? I think it's fine but just double checking.

I would leave them at the moment, since I'm not a security expert I would try to be extra cautious, even with more errors and error messages than necessary (for easier debug just in case?)

justinbeaty:
> I would leave them at the moment, since I'm not a security expert I would try to be extra cautious, even with more errors and error messages than necessary (for easier debug just in case?)

Yeah I think it's fine to leave them visible, but will make a commit to differentiate between `Mage_Core_Exception` and `Exception` using the same pattern used in other controllers to make only the former is visible to the user. That will also make it easy to make certains errors throw a generic message in the future.

app/code/core/Mage/Adminhtml/controllers/System/AccountController.php
```diff
@@ -102,4 +107,195 @@ public function saveAction()
...
+            $user = Mage::getModel('admin/user')->loadByUsername($username);
+
+            if (!$user->getId()) {
+                throw new Exception('User not found');
+            }
+
+            $credentials = $user->getResource()->getReadConnection()->fetchAll(
+                "SELECT credential_id_hash FROM {$user->getResource()->getTable('admin/user_credentials')} WHERE user_id = ?",
+                [$user->getId()],
+            );
```
fballiano, line 241: couldn't we use a fetchRow or a fetchOne here?

justinbeaty: Alternatively is there a model we could use `->load($user->getId())` with?

fballiano: actually the user_credentials doesn't even exist, WTH is going on there...

fballiano: It was a leftover method from my first tests

justinbeaty:
Hm, there's still an error with the base64 encoding not matching PHP. I get an error when there's a `+` sign. Investigating now.

justinbeaty:
One other idea before I investigate the base 64 thing: Would it be worthwhile to make the fieldset similar to the 2fa box?

So we have an "Enable Passkey" dropdown box, and when set to "Yes" we show the button and when registered it saves the result into a hidden input that is processed in the normal save action? That would standardize the user flow of enabling and having to press the "Save account" button to finalize it.

Does that make sense? It would remove the `passkeyregistersaveAction` and `removepasskeyAction` methods and potentially move some logic into a helper class like `Mage_Adminhtml_Helper_TwoFactorAuthentication`.

justinbeaty:
> mmm the base64 thing was happening to me at the beginning, then I solved it adding those "replace" in the javascript functions.

Originally I was getting sporadic `atob` errors when the challenge contained certain characters. Now, I don't get those errors, but get `Invalid challenge` errors.

justinbeaty:
I wonder if it's easier to just forgo base64 and use hex strings... What do you think?

fballiano:
all the code I've found used those base64 stuff so it's probably needed somehow by the passkey protocol?

fballiano:
I get invalid challange now too so something broke today :-\

justinbeaty:
> I get invalid challange now too so something broke today :-\

Try several times, it should work occasionally. But then also try to checkout the earlier commit and you should also get sporadic errors. At least that's what happened for me.

fballiano:
I had to try a million times but, after adding those "replace()" I never got that problem again

justinbeaty:
I did:

```sh
git checkout 230b115d9ad0685993028e817956bc6fdf620d06
#  HEAD is now at 230b115d9a Update app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
```

And I get this error:
```
Failed to register passkey: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.
```

fballiano:
mmmm now my original commit doesn't work to me and the last one neither 😭

fballiano:
@justinbeaty could you give a try to this last commit?

justinbeaty:
Will try shortly. Thanks for removing that Mage::log lines, didn't mean to commit that!

fballiano:
those Mage::log pointed me in some direction :-D

justinbeaty:
@fballiano yes, it seems to fix it! But I still don't understand why it's a problem in the first place... 😕

justinbeaty:
Okay, it's that `navigator.credentials.create` specifically returns the challenges as `base64url` instead of regular base64...:

https://developer.mozilla.org/en-US/docs/Web/API/AuthenticatorResponse/clientDataJSON#challenge

fballiano:
seems to work fine :-) both registration and login

I thought I managed to block password login when a passkey is registered but seems not to work, checking

justinbeaty:
> I thought I managed to block password login when a passkey is registered but seems not to work, checking

Sent DM via Discord

fballiano:
@justinbeaty this one should work now. I migrated all the server-side code to lbuchs/webauthn and also the javascripts part had to be rewritten, but at least now we're sure about the implementation and it should be secure.

Thank you @gilbertococchi for pointing me in the right direction!!

justinbeaty:
Sweet, I’ll take a look tomorrow morning!

justinbeaty:
Pushed a few changes, and have a few more things in progress but too late for today.

1) Improved the login page a bit, mainly just added a lightweight mahoFetch compatible function as to not duplicate stuff with the 2fa code. Also improved some of the error handling.

2) I made sure `passkeyloginstartAction` validates the form key and only reads from POST. It should improve the security.

A question: Should we make it controllable whether or not to disable password logins? Since passkeys can be per-device, the current behavior could lock people out if not on their main machine.

fballiano:
> Pushed a few changes, and have a few more things in progress but too late for today.

niceee thanks, I was a bit "tired" of the topic to polish it up 😅

> A question: Should we make it controllable whether or not to disable password logins? Since passkeys can be per-device, the current behavior could lock people out if not on their main machine.

From what I've read it seems that the consensus is "disable password authentication because otherwise you're still open to low security authentication" and I'm fine with it. I think that, if a users sets up a passkey that's device related, that's their choice maybe because they want it this way, but most password managers carry passkeys though multiple devices so it's no problem.

I don't know, we could have a "enable password authentication" which is disabled by default but mmmm I'm not 100% convinced that's the best choice. I don't think I've seen it in other websites.

app/design/adminhtml/default/default/layout/admin.xml
```diff
@@ -21,6 +21,14 @@
         </reference>
     </adminhtml_input_counter_handle>
 
+    <!-- admin account edit page -->
+    <adminhtml_system_account_index>
+        <reference name="head">
+            <action method="addJs"><name>maho-passkey-tools.js</name></action>
```
fballiano, line 27: @justinbeaty I created this file because those functions are not 100% base64 and they can be used only for passkey related stuff (shared between frontend and backend)

justinbeaty: Got it, will leave it alone. 😄

fballiano: I mean, let's say we add passkey support for frontend one day... I don't know :-D

justinbeaty:
> From what I've read it seems that the consensus is "disable password authentication because otherwise you're still open to low security authentication" and I'm fine with it. I think that, if a users sets up a passkey that's device related, that's their choice maybe because they want it this way, but most password managers carry passkeys though multiple devices so it's no problem.
> 
> I don't know, we could have a "enable password authentication" which is disabled by default but mmmm I'm not 100% convinced that's the best choice. I don't think I've seen it in other websites.

I found there's varying levels of 2fa when it comes to passkeys: https://www.corbado.com/blog/webauthn-user-verification#1-introduction-webauthn-user-presence-and-user-verification

So, I'm thinking that 2fa can exist alongside passkeys if the user verified flag is false. That should be the case for password managers (as opposed to yubikeys or secure enclave.)

I think that makes sense because a passkey stored in a password manager isn't much different than a strong password stored there too. If someone compromises the manager then 2fa can stop that.

Alternatively, if we have the user verified flag then no 2fa prompt is needed because we know they used a PIN, fingerprint, facial recognition, etc.

Regarding enable / disable password auth, I haven't done a lot of research, but it seems most services allow passkeys along side traditional passwords.

app/design/adminhtml/default/default/layout/admin.xml
justinbeaty: Yeah, that would be excellent!

fballiano:
this is what i gathered:

> No, it generally doesn't make sense to add traditional 2FA when using passkeys, as passkeys already provide equal or superior security through their inherent design. Here's why:
> Built-in Two-Factor Security
> Passkeys already incorporate two authentication factors by default:
> Something you have: the device containing the passkey1
> Something you are/know: biometric verification (like FaceID/TouchID) or device PIN1
> Superior Security Features
> Phishing Resistance: Passkeys are inherently phishing-resistant because they use public-key cryptography where the private key never leaves your device23.
> No Password Vulnerabilities: Unlike traditional 2FA which builds upon potentially vulnerable passwords, passkeys completely eliminate password-related risks23.
> Interception Protection: Passkeys cannot be intercepted since they don't require manual input of codes or tokens, unlike traditional 2FA methods which can be vulnerable to interception through SMS or email3.
> User Experience
> Adding traditional 2FA on top of passkeys would create unnecessary friction without meaningful security benefits5. Passkeys already provide strong security while maintaining a smooth, single-step authentication process2.
> Exception Case
> The only time you might encounter 2FA with passkeys is when signing in on a new device for the first time, if the account specifically requires it4. However, this is a policy decision rather than a security necessity.

fballiano:
and since password login is super lower level of security, I'd avoid it. in case needed one can clear the 2 database fields and gain access again (ok that's not really comfortable but...)

justinbeaty:
> and since password login is super lower level of security, I'd avoid it. in case needed one can clear the 2 database fields and gain access again (ok that's not really comfortable but...)

I wouldn't always enable it, but I'm still thinking a disable password login setting might make sense. I could make a test commit so we could see how it works and make a decision.

fballiano:
> I wouldn't blanket allow password logins, but I'm still thinking a disable password login setting might make sense. I could make a test commit so we could see how it works and make a decision.

personally I wouldn't care but until the setting is disabled (meaning disable password login by default) i'm ok with it

> I gather that problem is that password managers don't always have the biometric / PIN step, or they unlock the vault for some arbitrary time period upon master password / PIN entry. Part of the client data is whether or not the user has been "verified", i.e. they used biometric or a PIN at the time of login. I think it depends on the password manager implementation, but if it reports that the client was verified I'd skip 2fa. However if the flag is false we could still prompt 2fa.

i'm not against supporting 2fa + passkey :-)

justinbeaty:
I have some commits I should be pushing tomorrow, just don't want you to think this has gone stale.

fballiano:
no problem, I'm working on the other thing in the meanwhile ;-)

justinbeaty:
Okay, here's what I changed. About to do testing now. You'll have to first run this, or reset the upgrade script state:

```sql
ALTER TABLE `admin_user`
  ADD COLUMN `password_enabled` TINYINT(1) NOT NULL DEFAULT 1,
  ADD COLUMN `passkey_enabled` TINYINT(1) NOT NULL DEFAULT 0;
```
---

1\. Account Edit Form -- I made a separate fieldset for password authentication options, and made validation so that at least one of password / passkey methods must be enabled.

After registering a passkey the user must press the "Save" button in the top right, but this makes it more consistent with enabling 2fa, etc. The old way you could set a passkey on the account without entering the current password, so I think the new way makes sense.

image

2\. I consolidated the 2fa and WebAuthn helper logic into `Mage::helper('admin/auth')`.

3\. I also consolidated some logic into Mage_Admin_Model_User.

4\. I refactored the `authenticate` method. I remember during the 2fa review I stared at that `$sensitive` variable for like 10 minutes trying to understand the logic. I think it's much more clear now.
* There was a bug where passkey authentication returned true before the account is active and role checks. So you were able to login with a passkey on an inactive account.
* 2fa might be required during passkey auth, if the user verified bit is not set. I'm not prompted for 2fa with Bitwarden.
* There's one minor change (breaking?) change: before the `admin_user_authenticate_after` event would fire with `$result = false` if you had a wrong username or wrong password, but not if the account was inactive or wrong role. I'm not really sure if that makes sense at all. Now the event only fires if `$result = true`, but we could easily add the event with `$result = false` in the catch block if we wanted to.

5\. Translations, including some from 2fa that we missed.

---

There were also some small framework changes here: https://github.com/MahoCommerce/maho/pull/97/commits/0644bd9db813969b0c597e29f9742a9d098cb679. Just to allow the `autocomplete` attribute in form fields, and to allow checkboxes in the form dependency block. We could put those in a separate PR, but otherwise it's fine here I think.

Let me know if I missed anything, or if you don't like how I changed something.

justinbeaty:
One bug fix:

Before the `admin_user_authenticate_after` event didn't fire when authenticating with a passkey, but after my changes it does. However, there was an event to upgrade the password hash, and since the passkey data is sent as the password field, this was causing the account password to be changed to the JSON data string when logging in via passkey.

I moved the logic out of the observer and put it directly into the authenticate method. Actually I think that event really shouldn't even be passing the password to observers, and at least in core it's no longer needed.

fballiano:
mmm weirdly I can't login anymore to the backend now, but I'll retest later.

A couple of quick notes

I wouldn't expose which kind of login method a user has, IMHO it's safer to only say "login failed".
Screenshot 2025-01-24 alle 18 33 04

I think I wouldn't like this layout too much, in the sense that now it seems you can disable password auth without having enabled passkey (I can't login to the backend now so maybe the field dependency is making it clean but I can's see it now) and, if we move that combobox (which to me it should be below the passkey options) then the "change password" checkbox doesn't really make sense to be there all alone no?
Screenshot 2025-01-24 alle 18 35 17

justinbeaty:
> mmm weirdly I can't login anymore to the backend now, but I'll retest later.

Were you trying password or passkey login? If passkey, since there is a new DB column `passkey_enabled` it wouldn't work until you go enable it. This of course wouldn't be a problem going forward.

> I think I wouldn't like this layout too much...

Frontend (and backend) validation ensure at least one is enabled. I originally thought to move the "Enable Password Auth" option in the passkey section, but with the change password box it seemed to make more sense this way...?

image


>  also IMHO if there is a passkey then "passkey login" should be enabled, I don't see a case when somebody would want to remove the possibility to login with the most secure method

I thought a bit about this, the reason I came to this form layout was for consistency with other sections. I'm not sure why one would disable the passkey option, but it could be helpful to do so without having to remove and recreate the passkey if you re-enable it later. Which is consistent with the password authentication option.

---

In any case, when you have time see if you can get logged in and test the form, we can make changes, revert, or whatever.

justinbeaty:
> I wouldn't expose which kind of login method a user has, IMHO it's safer to only say "login failed".

It was actually already exposed whether the user had passkey or not enabled if they saw the error message "Internal Error", or by visiting this URL (which is now POST only with form key validation.)
https://store.example.com/index.php/admin/index/passkeyloginstart/username/admin


I don't think it's a security issue though, you'll get "Passkey is not enabled for this account." even if you provide an incorrect username. I don't think there's a way around it, but I figured the error message is more helpful than "Internal Error".

fballiano:
oki I can login now (the 2 missing columns heheh)

when I select "no" to "enable password login" the checkbox to change the password disappears but I think it's not correct, I still may want to change the password for some reason (maybe because I want to disable to passkey or whatever).

IMHO we're making things more difficult that they should be, I would
- disabling passkey -> you remove the passkey and that's it
- disabling password auth doesn't hide the change password field
- I'd restore position of the change password field and move the "enable password auth", adding a comment on why there is such a field

If you want I can do the modifications :-)

justinbeaty:
> the checkbox to change the password disappears but I think it's not correct, I still may want to change the password for some reason (maybe because I want to disable to passkey or whatever).

You would only need your current password to disable passkey, but you make a point that you might want to change password for the other places that require it in the admin screen (changing customer password, api roles, etc). However that second thing should really eventually be replaced by some sort of modal letting you auth with password / passkey / 2fa. 

But yes, please go ahead and do the modifications if you don't mind. The form layout I chose was just a test concept.

fballiano:
> You would only need your current password to disable passkey, but you make a point that you might want to change password for the other places that require it in the admin screen (changing customer password, api roles, etc).

yes yes, I meant I want to set a secure(r) password before turning off passkey

> However that second thing should really eventually be replaced by some sort of modal letting you auth with password / passkey / 2fa.

that would be the best but... not for now :-D

> But yes, please go ahead and do the modifications if you don't mind. The form layout I chose was just a test concept.

max tomorrow I'll work on it :-)))

sorry for wanting to change something, you know I love the way you work.

justinbeaty:
It's absolutely not a problem to want changes, I fully admit I overcomplicate things some times. 😄

fballiano:
mmm some things are not clear to me
- the "remove passkey" button doesn't work (it doesn't do any ajax call nor form submit), but it was working when I left the PR
- when I register a passkey I get this
Screenshot 2025-01-25 alle 13 28 04  
but it doesn't show the "enable password authentication" because I have to refresh

sorry man, I kinda liked it more as it was before, with the complete submit of the form instead of the ajax call.

now I don't know what to do

fballiano:
What I've committed now:
- removed password_enabled field
- different layout of the form
- adding a check on user->save to avoid wrong disabling of password login if there's no passkey (I preferred when it was simpler -> passkey = no password and that's it, this passkey_enabled I think it's a bit convoluted and possible to errors and people unable to login)

but still doesn't work, the register passkey experience is not good enough because it doesn't refresh the page and the remove passkey doesn't work.

justinbeaty:
@fballiano The added or deleted passkeys aren't completed until the save button on the whole form is pressed. Before you could register a passkey without the current password, but with the logic consolidated in the saveAction it seemed cleaner to me. It also keeps it consistent with how 2fa is enabled, password changes, etc. 

I think there's other organizational stuff in the commits I pushed that are worth keeping, but let's get this form layout nailed down. If you push your work so far I can probably fix the few issues.

justinbeaty:
> The added or deleted passkeys aren't completed until the save button on the whole form is pressed.

And I'm not saying it has to be this way, but we should require the current password, which I know you handled on the delete action, but it might be a bit trickier on the create passkey option. Example:

- User leaves current password field empty
- User clicks register passkey and creates it through their password manager
- Our JS can't yet make the POST request, need to prompt for the current password
- How do we allow the user to send the POST request once they enter their password?

fballiano:
Ahhhh then i guess it was only the messages, cause i dee “passkey deleted”
and it’s not deleted but it wasnt clear i had to submit the form.
Then probably simply auto submitting the form (which will trigger field
validation) would solve.
I had to leave home at the moment so i cant rethink it right now

Il giorno sab 25 gen 2025 alle 13:37 Justin Beaty ***@***.***>
ha scritto:

justinbeaty:
I'll try adding the auto submit and other bug fixes.

justinbeaty:
Okay I think it is better now. 😃

app/code/core/Mage/Admin/Model/User.php
```diff
@@ -526,7 +526,7 @@ public function getPasskeyCreateArgs(): stdClass
     public function getPasskeyGetArgs(): stdClass
     {
         if (!$this->getPasskeyCredentialIdHash() || !$this->getPasskeyPublicKey()) {
-            Mage::throwException(Mage::helper('adminhtml')->__('Passkey is not enabled for this account.'));
+            Mage::throwException(Mage::helper('adminhtml')->__('You did not sign in correctly or your account is temporarily disabled.'));
```
fballiano: I also thought this was important. Although this error message is very unclear, I think it's better to be safer

fballiano: I also thought this was important. Although this error message is very unclear, I think it's better to be safer

justinbeaty: :+1: Better on the safe side.

fballiano:
niceeee, I've changed this sentence to make it even clearer

Screenshot 2025-01-25 alle 16 00 39

if you're ok with everything, I'd merge and start spamming 😂

justinbeaty:
Yeah, you're much better wording those messages than I am, haha.

One quick change I'd do is add a getter `Mage_Admin_Model_User::getPasskeyEnabled()` so that we don't have to do this everywhere:
```php
$passkeyEnabled = $this->getPasskeyCredentialIdHash() !== null;
```

fballiano:
will do right away

fballiano:
Added, but I've called it isPasskeyEnabled() since it's not a real "get" of a property/variable

justinbeaty:
> Added, but I've called it isPasskeyEnabled() since it's not a real "get" of a property/variable

Great, I'll test one more time then merge button. 😉

fballiano:
I've broke something cause I can't login anymore 🤦‍♂️😂

fballiano:
mmm seems that when the form is re-saved (let's say changing the username) after the passkey way already enabled -> can't login anymore

fballiano:
nothing, I'm allucinating, cause I've re-did everything and it's working perfectly

justinbeaty:
> nothing, I'm allucinating, cause I've re-did everything and it's working perfectly

I've tested saving, and every combination of passkey +/- password, and 2FA on both types of logins. Checked the DB values at each step and all seemed to work perfect.

If you can't reproduce again I'd say it's ready!

fballiano:
retested again and no problem, let's go!

fballiano:
and Maho is the first magento based platform to have passkey support

PR #98, opened 2025-01-17
FIX Layout shift

Genaker:


fballiano:
unfortunately the fixed min-height creates these kinds of problems 

Screenshot 2025-01-17 alle 08 23 44

:-(((

fballiano:
Thank you for your PR man, it poked me into fixing this, which actually was easier than expected: https://github.com/MahoCommerce/maho/commit/b9f7b64a153b5e61bb39c04bffcc0b00a42d1f48

fballiano:
At least for product page: https://demo.mahocommerce.com/men/shirts/plaid-cotton-shirt-476.html

For home page, since it's all CMS, it's the duty of the CMS people to create the blocks correctly ;-)

Genaker:
Oh. Maybe because it is a flexible size. I will check it out

fballiano:
it's been solved already I think, both in product and category page ;-)

PR #99, opened 2025-01-25
Removed password from `admin_user_authenticate_after` and added `admin_user_authenticate_failed` event

justinbeaty:
Mentioned in https://github.com/MahoCommerce/maho/pull/97#issuecomment-2613082519: 
> [The `admin_user_authenticate_after`] event really shouldn't even be passing the password to observers, and at least in core it's no longer needed.

Since the only other observer that hooks into that event doesn't use the `$password` value, I think we should remove it.

### DIFF
--- a/app/code/core/Mage/Admin/Model/User.php
+++ b/app/code/core/Mage/Admin/Model/User.php
@@ -488,7 +488,6 @@ class Mage_Admin_Model_User extends Mage_Core_Model_Abstract
 
             Mage::dispatchEvent('admin_user_authenticate_after', [
                 'username' => $username,
-                'password' => $password,
                 'user'     => $this,
                 'result'   => true,
             ]);
--- a/app/code/core/Mage/Admin/Model/User.php
+++ b/app/code/core/Mage/Admin/Model/User.php
@@ -495,6 +495,12 @@ class Mage_Admin_Model_User extends Mage_Core_Model_Abstract
             return true;
 
         } catch (Exception $e) {
+            Mage::dispatchEvent('admin_user_authenticate_after', [
+                'username' => $username,
+                'user'     => $this,
+                'result'   => false,
+            ]);
+
             $this->unsetData();
             throw $e;
         }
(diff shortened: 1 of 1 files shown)


fballiano:
agree 100%

justinbeaty:
One other change, I added the `Mage::dispatchEvent('admin_user_authenticate_after')` in the catch block too. Before Mage sometimes called that event with `$result = false` and sometimes just returned early. Might as well add it so modules could log failed attempts, or whatever else.

fballiano:
philosophical question, should observers be able to observe failed authentications? I wonder if they don't check the $result and think it's a successful authentication...

justinbeaty:
> philosophical question, should observers be able to observe failed authentications? I wonder if they don't check the $result and think it's a successful authentication...

Could be something to integrate into a fraud detection library, or industries that need high regulation and logging everything.

fballiano:
could we, just for safety, have admin_user_authenticate_after and admin_user_authenticate_failed_after (this names suck)?

justinbeaty:
> could we, just for safety, have admin_user_authenticate_after and admin_user_authenticate_failed_after (this names suck)?

I think `admin_user_authenticate_failed` is okay, because the "failed" part implies "after".

justinbeaty:
One more question, do any events in Magento pass the error object to the observer? Might be useful to be able to check the error codes we defined:

```php
    /**
     * Authentication error codes
     */
    public const AUTH_ERR_ACCOUNT_INACTIVE = 1;
    public const AUTH_ERR_ACCESS_DENIED = 2;
    public const AUTH_ERR_2FA_INVALID = 3;

```

fballiano:
I guess there was no standard but we can decide, if the error object contains code and message, I think it's totally fine

justinbeaty:
I think this is the best way. Dispatch the failed event only on `Mage_Core_Exception` since the above try block throws those with non-specific messages and the `AUTH_ERR_` constants. If there's a PHP exception something else went wrong during the authentication.

https://github.com/MahoCommerce/maho/blob/db13134daedd7e1f1c53d668caf7a5e7ab56a078/app/code/core/Mage/Admin/Model/User.php#L497-L508

PR #102, opened 2025-01-29
New header layout

fballiano:
We already made RWD a bit better with small updates, but the header was still looking pretty old, this PR should take it to a modern version that's pretty much used by everybody now. For sure the trigger for this redesign was @empiricompany.

This is the new header for desktop
Screenshot 2025-01-29 alle 22 42 45

and for mobile
Screenshot 2025-01-29 alle 22 32 39

Notes:
- gone is the dedicated line for store-switch and currency-switch
- gone is the welcome message (in high traffic websites it can create performance problems since it's a hole punch in FPC)
- gone is the "customer account" submenu, all the elements were redundant and the logout button has been now moved to the "customer navigation sidebar" in the customer area
- I've also done some cleanup to the CSS
- I'd like to do more work on the icons but it would have been too much for this PR and it will come next
- I've also separated che "checkout related" CSS into a separate file, since it was a request made to me on twitter (I did it in this PR because it would create a conflict otherwise). it's not a big deal but it still spares 1300 lines of CSS from the normal website navigation


justinbeaty:
Nice, I'll take a look tomorrow morning!

justinbeaty:
For me the account and wishlist buttons aren't working, is there something I am missing?

fballiano:
fixing them in a bit

fballiano:
should be ok now

justinbeaty:
This looks really good to me. 😎

One idea, to fix the minicart pop-up positioning:

```css
.header-minicart {
    position: relative;
}
```

fballiano:
I didn't notice that cause I always had the developer console open 😂 thanks for noticing! already committed

justinbeaty:
This looks good to me. I saw you came to the same conclusion about using image mask + background color for the icons. I was doing the same for the admin.

justinbeaty (approved):

fballiano:
Thanks @justinbeaty!

about the mask, yeah that's the only way AFAIK but it has a lot of limitations... I'm thinking that there's [a wonderful SVG icon set](https://tabler.io/icons) that I'd like to include somehow, which supports not only color, but also configurable stroke size, but that's only possible if we don't use the SVGs as background...

fballiano:
wow the Mage_Core_Block_Icon it's a freaking good idea!!

I think we can be ok with having them inline.

fballiano:
ps: I've examined a lot of icon library (usually I prefer tailwind's heroicon) but tabler is so much better with the vatly bigger number of icons and the customizability

empiricompany:
I’m reporting these issues on Firefox:
firefox

- logo don't show
- select for switch language have background

chrome seems ok:
chrome

fballiano:
Thanks man, I think I've solved those :-)

PR #104, opened 2025-02-03
Frontend: added `srcset 2x` to all templates

fballiano:
This PR should fix the old problem of blurry images

### DIFF
--- a/app/design/frontend/base/default/template/bundle/catalog/product/list/partof.phtml
+++ b/app/design/frontend/base/default/template/bundle/catalog/product/list/partof.phtml
@@ -21,7 +21,7 @@
         <?php for($_j=0;$_j<$this->getColumnCount();$_j++): ?>
             <?php if($_bundle=$this->getIterableItem()): ?>
             <td>
-                <a href="<?= $_bundle->getProductUrl() ?>" title="<?= $this->escapeHtml($_bundle->getName()) ?>" class="product-image"><img src="<?= $this->helper('catalog/image')->init($_bundle, 'small_image')->resize(125) ?>" width="125" height="125" alt="<?= $this->escapeHtml($_bundle->getName()) ?>" /></a>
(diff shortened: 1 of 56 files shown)


justinbeaty:
Reviewed code change on mobile, but looks like it does exactly what it says.

I had one thought on a block class for rendering catalog images with an optional link wrapper. Just pass it the `$productId`, pixel size, and title, srcset, alt, plus other attributes for both the <a> and  tags. Kind of like that other block idea I already forgot...

But this can absolutely be merged as is if that's too complex. I can test in morning unless you're sure it's ready to merge then go ahead.

fballiano:
it's a very good idea but let's merge this and do that separately, I'd also like to remove RWD altogether and merge everything in base/default, finally ;-)

fballiano:
weirdly enough pagespeed complains that images are not sized properly https://pagespeed.web.dev/analysis/https-demo-mahocommerce-com-home-decor-bed-bath-html/wakr841vk9?form_factor=mobile but... I don't understand...

fballiano:
> weirdly enough pagespeed complains that images are not sized properly https://pagespeed.web.dev/analysis/https-demo-mahocommerce-com-home-decor-bed-bath-html/wakr841vk9?form_factor=mobile but... I don't understand...

@justinbeaty do you know something about this?

justinbeaty:
Because with that screen size the products are 1/2 the screen width, so you can't size it exactly to whatever device Google picks to run the default tests. You can aim for an average image size for certain breakpoints, and with the 2x images they'll still be sharp on high DPI.

justinbeaty:
image

I did math wrong -- first I forgot to actually select `Moto G Power` from the device list so my previous post's screenshot wasn't loading the 2x images from the srcset... 😅 

Updated math is $(178px*1.75)^2/420px^2$ which matches PageSpeed at 45%. I would wait until you choose finalized breakpoints for the theme, then figure the average width images will use in the two column layout, then set base/small/thumb image sizes based on that and you'll do the best for most users.

justinbeaty:
You could actually also target 1.5x, 1.75, and 2x DPI if you want to super optimized images.

Maybe you'd use an image helper method like this:

```php
<?php foreach ($this->getProducts() as $product): ?>
    <?= $this->getCatalogImageHtml($product, [
        'link' => $this->getUrl('...', [ 'product_id' => $product->getId() ]),
        'image_size' => 'sm',
    ]) ?>
<?php endforeach ?>
```

And it could be configured to include different images in the srcset attribute.

fballiano:
mmmmm I wouldn't add to many "breakpoints" because they will make the images cache multiply so...

at the moment I think the best is to leave it as it is, on the demo on my devices the images look so much better now, which is good.

one day we should revise the whole images of the default template, there are too many resolutions used across the templates, that generate too many cache files, multiplied by every store... but today it's not that day :-D

PR #106, opened 2025-02-04
Configurable Product Improvements

justinbeaty:
Configurable Product Improvements
========================

I discovered some bugs while working on the admin theme, and created this PR with multiple areas of improvement around the Product Config classes. Here's an overview of the commits:

## Added symfony/polyfill-php84
If our supported range is 8.2-8.4, we should have polyfills for both 8.3 and 8.4 so we can use the features of our highest version. We had 8.3 already, but I wanted to use PHP 8.4's new `array_any` function. 😃 

## Maho Template fix for PrototypeJS Hash Object
There was something broken from #68 but we didn't notice: the new Template class did not accept a PrototypeJs Hash object for the data object. There were two possible fixes and I chose the more generic one, but I left the prototype specific fix commented out.

To reproduce go to `Manage Products > Configurable Product > Associated Products` and you'll see before there's no form values and with this commit it is fixed.

image

This actually seemed to cause configurable prices to be wiped upon product save. I was a bit confused for a while because for some products additional prices weren't showing on demo.mahocommerce.com nor my local install while in OM they were. I finally figured out this was the cause. I haven't checked if sample data had any corrupted values as not sure how you generate that.

## Maho Dialog Improvements
No PR is complete without a maho dialog improvement... 😅 

1\. Remove button styling, was actually part of the admin theme but I think it's fine here.
2\. Add optional `options.className` setting.
3\. Focus the dialog body with tabindex so the close button isn't focused by default.
4\. Most importantly, allow the `ok` handler to return false to prevent dialog close. This is to prevent dialog close on failed form validation.

## Product Config: Merge JS Classes
First, I had not realize how bad the `varien/product.js` and `varien/configurable.js` files were before you even cleaned them up. I noticed some odd things while reviewing #68, but I didn't realize how much code was duplicated between the various files.

While working on the admin theme, I found a few bugs introduced in #68.  Disable swatches and go to these frontend pages:
* URL Hash params doesn't work: https://store.example.com/plaid-cotton-shirt-404.html#92=17
* URL Query params breaks form: https://store.example.com/plaid-cotton-shirt-404.html?___store=en

The second issue actually affected the admin too, below are screenshots of how the configure form was broken:

image

I was able to fix some of these bugs in `varien/configurable.js`, but you'll notice I renamed the class to `Product.Config2` and the class is not used. I went a step further and consolidated logic into one `Product.Config` class located in `varien/product.js`.

There shouldn't be any breaking changes here at all, however I need to do more testing to make sure nothing broke on the frontend. For now, the swatches portion still extends `Product.Config`, but in the future that could be made to work on both frontend and backend. 

## Product Config: Remove obsolete JS code
This removes the old `varien/configurable.js` file as well as removes the `Product.Super.Configurable` class which seems to be obsolete code. Running `git grep -F Product.Super` returns nothing...

## Product Config: AJAX Result Rendering
The [Manage Categories code from #81](https://github.com/MahoCommerce/maho/pull/81#issue-2761784337:~:text=returned%20even%20more%20JS%20which%20was%20evaluated%20in%20an%20iframe%20to%20call%20functions%20from%20the%20phtml%20files) was the worst code I've seen in Magento, but the product config AJAX code was similarly bad or maybe takes the prize.

Like the manage categories form, the configurable form used a pattern where the results of a form is rendered into an iframe, which contains a single `<script>` tag that sets a global variable with the results of the API call. The JS code then reads the global variable with the API response on the iframe's onload callback. Example:
```html
    <iframe id="product_composite_configure_iframe" src="" onload="productConfigure.onLoadIFrame()"></iframe>
    <form action="" method="post" enctype="multipart/form-data" target="product_composite_configure_iframe">
        <!-- ... -->
    </form>
```

Even worse, Magento used the layout XML system to let blocks render JSON results. It is a million times better to use mahoFetch to POST the data and read the results as normal JSON.

If you checkout this commit, you'll be able to POST data to save configurations, but the UI will not reload because it is still expecting `<script>` tags with results. See next commit for template and JS changes.

## Product Config: Adminhtml Templates + JS Dialog Controller

This commit mostly deals with `mage/adminhtml/product/composite/configure.js` and the adminhtml templates used for configuring products. The JS class has been rewritten with modern JS, works with maho dialog instead of it's own absolutely positioned window element, and many other improvements.

## Product Config: Adminhtml Controllers Improve AJAX Handling
Clean up the rest of the controllers after the "AJAX Result Rendering" commit. This is basically using our beloved `$this->getResponse()->setBodyJson()` method, but there was also a significant security issue here. Before the form_key was serialized into the buyRequest, and if an admin was the last to modify then their form_key could be exposed to the frontend.

I am pretty sure there aren't major breaking change here. With this and the previous two commits, everything should be compatible API wise, just with the AJAX handling greatly improved.

## Mage Wishlist: Model Improvements and Grid Blocks

Holy, there was a lot broken in the wishlist module. These are major bugs present in 2021 branch of OM and probably Magento 1.9.4.x or older. Go to Admin > Manage Customer > Wishlist Tab. If you:

1\. Configure a product, *change* a product attribute, and save then **ALL OTHER ITEMS** except the one you saved are removed from the customer's wishlist.
2\. Configure a product, *do not change* a product attribute, and save then **ALL ITEMS INCLUDING** the one you saved are removed from the customer's wishlist.
3\. Weird stuff with the `store_id` on the `wishlist_item` table, it would reset to admin `store_id=0` when saving, making it impossible for customer to move item to their cart.

I copied enough code from `Mage_Sales_Model_Quote`, `Mage_Sales_Model_Quote_Item`, and other classes to `Mage_Wishlist_Model_Wishlist`, `Mage_Wishlist_Model_Item`, etc to finally make the module work.

> [!IMPORTANT] 
> When adding a new item to a wishlist, you must call the `save()` method. I only found one place where the `save()` method was missing, and it's not code a developer would have referenced when making their own module.  
> ```php
> // Simply adding the item no longer automatically saves
> $wishlist->addNewItem(...);
> 
> // Use:
> $wishlist->addNewItem(...)->save()
> // Or:
> $wishlist->addNewItem(...);
> $wishlist->save()
> ```
>
> There may be other Breaking Changes in the Wishlist module, but (IMO) if it was broken already then it can hardly be a breaking change... 

More improvements: the grid blocks for both manage wishlist and quote items have been improved. They can be made better to make 3rd party multi-list modules just obsolete, but at least for now it doesn't destroy customer's wishlists...

Edit: some formatting


justinbeaty:
P.S. I still need to do final round of testing, but this branch is currently rebased on top of main. If the commit messages are okay, you could merge without squashing or having a merge commit, would keep cleaner history and I tried to organize those commits logically.

app/code/core/Mage/Wishlist/Model/Resource/Item/Collection.php
```diff
@@ -22,6 +23,11 @@
  */
 class Mage_Wishlist_Model_Resource_Item_Collection extends Mage_Core_Model_Resource_Db_Collection_Abstract
 {
+    /**
+     * Collection wishlist instance
+     */
+    protected Mage_Wishlist_Model_Wishlist $_wishlist;
```
fballiano, line 29: the comment says it's a collection but the type is not a collection. also are we supposed to not use $_ anymore?

app/code/core/Mage/Adminhtml/controllers/Customer/Wishlist/Product/Composite/WishlistController.php
```diff
@@ -25,10 +26,17 @@ class Mage_Adminhtml_Customer_Wishlist_Product_Composite_WishlistController exte
...
     /**
-    * Wishlist we're working with
-    *
-    * @var Mage_Wishlist_Model_Wishlist
-    */
+     * Customer we're working with
+     *
+     * @var Mage_Customer_Model_Customer
+     */
+    protected $_customer = null;
```
fballiano, line 33:
```suggestion
    protected Mage_Customer_Model_Customer $customer = null;
```

what about this?

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist.php
```diff
@@ -227,6 +227,37 @@ protected function _setCollectionOrder($column)
...
+    {
+        return Mage::registry('current_customer');
+    }
+
+    /**
+     * Gets customer assigned to this block
+     *
+     * @return Mage_Customer_Model_Customer
+     */
+    public function getCustomer()
```
fballiano, line 245:
```suggestion
    public function getCustomer(): Mage_Customer_Model_Customer
```

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist.php
```diff
@@ -227,6 +227,37 @@ protected function _setCollectionOrder($column)
...
 
+    /**
+     * Retrieve current customer object
+     *
+     * @return Mage_Customer_Model_Customer
+     */
+    protected function _getCustomer()
+    {
+        return Mage::registry('current_customer');
+    }
```
fballiano, line 238:
```suggestion
    protected function _getCustomer(): Mage_Customer_Model_Customer
    {
        return Mage::registry('current_customer');
    }
```

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist.php
```diff
@@ -48,22 +47,18 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_Wishlist extends Mage_Adminhtml_Blo
...
     }
 
-    /**
-     * Retrieve current customer object
-     *
-     * @return Mage_Customer_Model_Customer
-     */
-    protected function _getCustomer()
+    #[\Override]
+    protected function _prepareGrid()
```
fballiano, line 58:
```suggestion
    protected function _prepareGrid(): Mage_Customer_Model_Customer
```

public/js/varien/js.js
```diff
@@ -835,6 +835,10 @@ class Template
      * @returns {string} Interpolated string
      */
     evaluate(data = {}) {
+        // if (typeof Hash !== 'undefined' && data instanceof Hash) {
```
fballiano, line 838:
```suggestion
```

app/code/core/Mage/Wishlist/Model/Wishlist.php
```diff
@@ -74,31 +74,124 @@ protected function _construct()
...
+    {
+        $this->setUpdatedAt(Mage::getSingleton('core/date')->gmtDate());
+        return parent::_beforeSave();
+    }
+
+    /**
+     * Save related items
+     */
+    #[\Override]
+    protected function _afterSave()
```
fballiano, line 91: can we add return type here?

app/code/core/Mage/Wishlist/Model/Wishlist.php
```diff
@@ -74,31 +74,124 @@ protected function _construct()
         $this->_init('wishlist/wishlist');
     }
 
+    /**
+     * Set date of last update for wishlist
+     */
+    #[\Override]
+    protected function _beforeSave()
```
fballiano, line 81: can we add return type here?

public/js/varien/js.js
fballiano: sorry, I read that it's there on purpose

fballiano:
on an empty customer's wishlist, I click search and get an error:

Screenshot 2025-02-04 alle 23 02 10

fballiano:
I had a product with "dates" custom options in my wishlist:

Screenshot 2025-02-04 alle 23 08 16

justinbeaty:
:+1: thanks for finding so many bugs and return type updates so quickly.

Do you want to eventually merge without squashing the commits here? If so I'll amend the fixes onto the appropriate commits for that clean commit history, otherwise if you'd just squash it then doesn't matter and can commit changes via GitHub UI.

fballiano:
> 👍 thanks for finding so many bugs and return type updates so quickly.

man, these are super minor things :-)

> Do you want to eventually merge without squashing the commits here?

I'll do as you wish, I've no problem with not squashing

public/js/varien/js.js
justinbeaty: Oh yeah, I'll totally remove it though. I just wanted to point out it was my initial solution, but I think checking for a `data?.toObject()` method (which prototypejs has), is the better solution right?

app/code/core/Mage/Wishlist/Model/Wishlist.php
justinbeaty: These are just moved methods to more closely match `Mage_Sales_Model_Quote`. 

However for research we have `class Mage_Wishlist_Model_Wishlist extends Mage_Core_Model_Abstract` and:

```php
abstract class Mage_Core_Model_Abstract extends Varien_Object
{
    /**
     * Processing object after save data
     *
     * @return $this
     */
    protected function _afterSave()
    {
        Mage::dispatchEvent('model_save_after', ['object' => $this]);
        Mage::dispatchEvent($this->_eventPrefix . '_save_after', $this->_getEventData());
        return $this;
    }
```

I think we could still technically add the return type, but so much code probably extends `Mage_Wishlist_Model_Wishlist` that I'm not sure how beneficial it is.

justinbeaty:
> on an empty customer's wishlist, I click search (or reset filter) and get an error:

I get that same error using OM main branch on the customer's cart page when pressing reset filter. So apparently I copied a bug too.

> I had a product with "dates" custom options in my wishlist (maybe not related to this PR):

Final TODO list included checking all other input types. It looks like that error probably occurs without this PR, but I'll check it out. I mostly just did style changes and removing useless `id=""` attributes on div tags. 

> man, these are super minor things :-)

Yeah, I had some product configurable stuff I experimented a few weeks back with, but these one-off dialog controllers were being a roadblock to the adminhtml theme. Easier to rewrite it but I had just as much hell of a time as you on these classes, plus the weird form + iframe API calls and it's better to renew.

I'm also excited to have a decent product configure pop-up "controller" class. It's a decent system to use the layout XML for rendering the forms, but using it to render JSON and `<script>` tags was just wild. This was probably original code Varien Inc. wrote for Magento back in 2005... 😅 

> I'll do as you wish, I've no problem with not squashing

Also makes `git bisect` better for changes like this. Some of the more sensitive changes are in smaller commits here so you don't have to look through 1,000 lines of code change to find the obvious problem. Will finalize tomorrow if no other bugs are found.

fballiano:
> Also makes `git bisect` better for changes like this. Some of the more sensitive changes are in smaller commits here so you don't have to look through 1,000 lines of code change to find the obvious problem. Will finalize tomorrow if no other bugs are found.

for sure, in this case that you have these very nicely separated commits it's a no brainer :-)

fballiano:
@justinbeaty shall I fix the conflict? I'd like to merge this since we have quite some stuff that's 90% there but not merged

justinbeaty:
> @justinbeaty shall I fix the conflict? I'd like to merge this since we have quite some stuff that's 90% there but not merged

I am about to rebase and fix this and the admin theme PR, need to test them both together again. I've discovered the custom options tab on the edit product page that can add options to the configure window, which I hadn't tested previously. A lot of this stuff was neglected by Magento devs so I want to make sure it all works finally. :)

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist.php
justinbeaty: This one is not possible as the method existed previously.

justinbeaty:
@fballiano This is ready for review again. Here's a highlight of the changes:

## Flatpickr: enable static option for usage in dialogs
Commit 9b4ced995ef1d0e6c375a81ecd8b5990b7768963 - fixes the bug you reported where the flatpickr doesn't show up in the `<dialog>` element. The `static: true` option could break layout if the parent is a flexbox container for example, but there are ways to work around that and I'm not sure yet if it's a problem anywhere. The PHP code update is just a bonus, I though the error was there at first.

## Maho Validation: Remove Function constructor
Commit ca1a4474b38e8341e9c7afe7d90987d8163ba953 - just gets rid of one more `eval` / `Function` constructor. I don't think it will break anything, but since it's a separate commit will be easy to track down in case it does.

## Maho Dialog Improvements
Commit 55545295e12565f5eb960683911ecbbbc20a3515 - there are some more updates here. Mainly I was getting double fire events (both an `ok` and `cancel`) which caused problems. It's because `closeDialog()` called `dialog.close()`, which called `closeDialog('cancel')`.

## Product Config: Remove obsolete JS code
These two JS classes weren't used anywhere.
https://github.com/MahoCommerce/maho/blob/ccb4887f4b0f8d821406381a4e940884c86721e8/app/design/frontend/base/default/template/catalog/product/view/options.phtml#L17-L106

## Product Config: Frontend Templates
Commit b5965440fe8006f77c82abdd35813e45168e1dfd - these templates were nearly identical to templates in admin, so I updated them here too. There was actually some PrototypeJS [hiding](https://github.com/MahoCommerce/maho/blob/ccb4887f4b0f8d821406381a4e940884c86721e8/app/design/frontend/base/default/template/catalog/product/view/options/type/file.phtml#L71) in one of the frontend templates still, so that's eradicated. :smiley:

justinbeaty:
I haven't gone through final testing yet, so I expect you'll find something. here's my notes 

Types of product configurations:
1\. Different product types (grouped, configurable, downloadable).
2\. Edit Bundle Product > Bundle Items > Input Type (dropdown, multi-select, checkbox, radio).
3\. Custom Options.
4\. Combinations of 1, 2, 3.

For each product configuration, test:
1\. Add products to wishlist, cart, compare on frontend.
2\. Add products to new order on admin.
3\. Configure products in customer cart and wishlist tab in admin.

Other testing:
1\. Admin order sidebar move items
2\. Configure and add to cart from sidebar (Create order Jane Doe English)

fballiano:
huge job @justinbeaty, today I'll be traveling for a few hours and try to test everything while offline!

justinbeaty:
I forgot to mention this one:

## Product Config: Adminhtml Templates + JS Dialog Controller
Commit 1f257dd1724961f3919acf3611a2115acc7cab03 - In `js/mage/adminhtml/product/composite/configure.js` I have a `compatibilityFlags.detachConfigureContainerElement` flag that changes how Magento / Maho keeps track of the configured item state.

Before: there was a hidden `<div>` that contains copies of the product configure form. The JS clones the form into the dialog as needed. This causes at least two separate issues caused by duplicate IDs. (1) `<label>` elements can point to the wrong input since they're duplicated, (2) validation for radio / checkboxes can fail if it's `one-of` type, since the hidden inputs are counted towards that validation rule.

Now: we still keep track of state like in Magento, except with the nicer `FormData` object, but by default we disconnect the node from the DOM so that duplicate IDs are not an issue. This required one change in [sales.js](https://github.com/MahoCommerce/maho/pull/106/commits/1f257dd1724961f3919acf3611a2115acc7cab03#diff-5acc99a91811853972e4209785bc38eaca1253069c434df0c3633c67a6690f82) to check `if (productConfigure.itemConfigured())` instead of checking for the `<div>` element in the DOM.

Originally I had the compatibility flag set to `false` to preserve compatibility, but there are too many bugs caused by it. Some of the bugs only show up in Maho because our dialog element happens to go after the hidden `<div>` so `<label>` elements work fine in OM even with duplicate IDs, but the other bugs are present in OM too. Keeping track of state in DOM is such a terrible idea that it's worth breaking the compatibility here.

fballiano:
Hi @justinbeaty:
- **flatpickr**, on grids (like customer grid), I can't click on the "customer since", it takes 2 clicks for flatpickr to open, the first doesn't do anything (same for orders grid)
- **validation**, for example in the customer registration form, if you try to submit, some fields get validated some others no
Screenshot 2025-02-28 alle 16 12 15

I found out that's not because of this PR but because of the code in this method:
```php
public function getAttributeValidationClass($attributeCode)
    {
        /** @var Mage_Customer_Model_Attribute $attribute */
        $attribute = $this->_attributes[$attributeCode] ?? Mage::getSingleton('eav/config')->getAttribute('customer_address', $attributeCode);
        $class = $attribute ? $attribute->getFrontend()->getClass() : '';

        if (in_array($attributeCode, ['firstname', 'middlename', 'lastname', 'prefix', 'suffix', 'taxvat'])) {
            if ($class && !$attribute->getIsVisible()) {
                $class = ''; // address attribute is not visible thus its validation rules are not applied
            }

            /** @var Mage_Customer_Model_Attribute $customerAttribute */
            $customerAttribute = Mage::getSingleton('eav/config')->getAttribute('customer', $attributeCode);
            $class .= $customerAttribute && $customerAttribute->getIsVisible()
                ? $customerAttribute->getFrontend()->getClass() : '';
            $class = implode(' ', array_unique(array_filter(explode(' ', $class))));
        }

        return $class;
    }
```
the $class becomes "required-entryrequired-entry" with no space in between and it breaks the validation. at the moment I can't work on a fix.

- **dialog** couldn't find problems with it
- **admin wishlist**: seems to me that works great
- **product config on frontend**: I didn't know it was possible to have `#92=25`in the url, but anyway now it works ok
- **product config ajax on backend**: tested order creating with product configuration with all product types and I couldn't find any problem
- about your last comment, I'm not sure how to test

justinbeaty:
@fballiano regarding flatpickr, I updated this commit: ea3640e5fcf0c94d7bb8580891729cd836eaff31

Now, it uses `static: true` only if it's detected in a dialog. I think the double click thing was something to do with the events, as the `removeEventListener` wasn't properly working.

I apologize about the force pushes, hopefully you have a clean way to reset the changes to this branch.

fballiano:
retested flatpickr now works perfectly in both grids and dialogs.

would you want me to test something specific?

justinbeaty:
> would you want me to test something specific?

No, I'm just going to do some final testing myself with my multiple choice test product, haha 😅 

image

> retested flatpickr now works perfectly in both grids and dialogs.

I'll mostly revert that commit and just leave the `static: !!inputEl.closest('dialog')` line for the dialogs, because I was experiementing with using flatpickr for both datetime and time only fields, but that's a larger change so we can leave for now and I'll open a new PR for flatpickr today/tomorrow.

public/js/varien/product_downloadable.js
```diff
@@ -0,0 +1,40 @@
...
+ * @copyright   Copyright (c) 2021-2024 The OpenMage Contributors (https://openmage.org)
+ * @copyright   Copyright (c) 2024-2025 Maho (https://mahocommerce.com)
+ * @license     https://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
+ */
+
+var Product = Product ?? {};
+
+/**
+ * Downloadable Product Controller
+ */
```
fballiano, line 15:
```suggestion
```

public/js/varien/product_bundle.js
```diff
@@ -10,6 +10,9 @@
 
 var Product = Product ?? {};
 
+/**
+ * Bundle Product Controller
+ */
```
fballiano, line 15:
```suggestion
```

public/js/varien/product.js
```diff
@@ -10,7 +10,9 @@
 
 var Product = Product ?? {};
 
-/**************************** CONFIGURABLE PRODUCT **************************/
+/**
+ * Configurable Product Controller
+ */
```
fballiano, line 15:
```suggestion
```

justinbeaty:
Ok, this should be the last round of review. Here's the changes since your last review: [Compare Changes](https://github.com/MahoCommerce/maho/compare/e37d3da30764da84dd57e4bd7b39ba30d382f552..afa5bbbd0b0daf9c3db156e7eea6214447ee7301)

I found another bug in OM regarding wishlists. To reproduce, create a new admin order, add SKU acj003, then move the item between the wishlist and the quote until the item disappears and the wishlist is emptied. I had already fixed that specific bug with this PR, but instead the quantity kept going up in the wishlist since the "Add To Order" checkbox only copies the item from the wishlist. I changed this so that the item is deleted from the wishlist when adding to the order becausue that's consistent with how moving from the shopping cart works, as well as how it works on the frontend.

> [!IMPORTANT] 
> When adding a new item to a wishlist, you must call the `save()` method. I only found one place where the `save()` method was missing, and it's not code a developer would have referenced when making their own module.  
> ```php
> // Simply adding the item no longer automatically saves
> $wishlist->addNewItem(...);
> 
> // Use:
> $wishlist->addNewItem(...)->save()
> // Or:
> $wishlist->addNewItem(...);
> $wishlist->save()
> ```

It is clear to me while working on this that at some point the wishlist and quote item models were similar, but the quote item code has seem improvements while the wishlist module has stagnated. Interestingly I can see some left over support for multiple wishlists. I assume this was an EE feature, but I have a custom multi wishlist module that we could pull from to create that feature here. In any case, I don't think breaking changes is a huge deal for the Wishlist is we're improving it at the same time.

If the changes look good to you, I'll re-org the commits and force push one last time and we can merge. Do all [the commit messages](https://github.com/MahoCommerce/maho/pull/106/commits) up until *Mage Wishlist: Model Improvements and Grid Blocks* look good to you? I believe you will need to enable the "Rebase and Merge" option to not squash the commits, or alternatively it could be done via CLI but probably better with the GitHub UI.

fballiano:
> [1] Bug: All custom options transfer their configured values when moving products between cart, wishlist, re-ordering in admin, etc, EXCEPT the file type. This makes it impossible to "Add to Cart" from the wishlist a product with a required file option.

I'd say it's not a problem, I'll open a new issue about that anyway just to keep track of it

justinbeaty:
> is the change to maho dialog in this last set of modifications intentional? should I retest the dialog related stuff?

Yes, we need to `await options.ok()` because the configure popup will send to the server and possibly get an error message in response. In that case, we want to prevent the dialog from closing to show the error. Here's the full diff this PR will make to that part of the code:
```diff
-        function closeDialog(action) {
+        async function closeDialog(action) {
             if (action === 'ok' && typeof options.ok === 'function') {
-                options.ok(dialog);
+                const result = await options.ok(dialog);
+                if (result === false) {
+                    return;
+                }
```
It should not affect anything anything else that uses maho dialogs. You can also await a non async function with no problem so it doesn't mean `options.ok` must be an async function.

> I see a new save() method in app/code/core/Mage/Wishlist/Model/Wishlist.php which sets _hasDataChanges to true but why is that necessary?

That method already exists in the current codebase: https://github.com/MahoCommerce/maho/blob/e423a3363f6ae3da86d14d0c3f659ade319582aa/app/code/core/Mage/Wishlist/Model/Wishlist.php#L594-L599 

I had removed it because `Mage_Sales_Model_Quote` doesn't have it (only `_beforeSave` and `_afterSave`) and I thought everything was working, but I had broken the edit product configuration from the wishlist on the frontend. Mage_Wishlist is still kind of a mess, but I hadn't set out to improve the wishlist functionality in this PR... It's only that, at a very bare minimum, the admin should be able to configure products in the customer's wishlist without randomly erasing everything.

> I've enabled rebase&merge in the meanwhile
I could edit the commit messages so they reference the PR, i.e. `Product Config: Remove obsolete JS code (PR #106)`. Alternatively, we could also use the regular merge button, so all of the commits and then the one merge commit that *should* reference this PR. Any thoughts?

fballiano:
I'd say that, for this PR, if you force push all the commits as you did before, I'd go for rebase&merge because they're connected but they're also separate issues :-)

justinbeaty:
I had a few final changes, mainly I improved some templates on the frontend and forgot to make the changes in the backend, vice-versa. [Changes](https://github.com/MahoCommerce/maho/compare/afa5bbbd0b0daf9c3db156e7eea6214447ee7301..76c0622eb44446b7e0a7c165c5bcf1616ec17c4b)

I also reduced the Flatpickr commit to the bare-minimum to allow it to work in dialogs. I'll open a new PR as soon as we merge this one to introduce those other improvements, and more. For some reason the double clicking is back for me in dialogs, but not on the grid. I'd also just remove this commit from this PR and we fix it in the new PR.

fballiano:
yeah I'd also remove the flatpickr one if you agree

justinbeaty:
> yeah I'd also remove the flatpickr one if you agree

Done, feel free to rebase and merge when the checks are complete. :smiley:

PR #103, opened 2025-01-30
Added `blog` module with admin interface, frontend, search and APIs integration

fballiano:
This PR introduces a comprehensive blog module to Maho, providing a full-featured blogging platform with modern architecture, admin interface, API support, and seamless frontend integration.

  **Features**

  - Post Creation & Editing: Rich WYSIWYG editor (TipTap) with media support
  - Publishing Control: Scheduled publishing with publish dates
  - SEO Optimization: Full meta tags support (title, keywords, description, robots)
  - Multi-store Support: Store-specific post visibility and management
  - URL Management: SEO-friendly URLs with custom URL keys
  - Image Support: Featured images with proper media handling

**Architecture**

  - Hybrid EAV Design: Static columns for performance + EAV flexibility
  - Static Attributes: title, url_key, is_active, publish_date, content, meta fields
  - EAV Attributes: image, stores for multi-store flexibility
  - Performance Optimized: Single-table queries for common operations
  - Smart Collections: Automatic static vs EAV attribute detection

**Frontend Integration**

  - Smart Navigation: Automatic blog link in main navigation menu
  - Active State Management: Proper highlighting when on blog pages
  - Conditional Display: Link appears only when published posts exist
  - SEO-Friendly URLs: Clean URLs like /my-blog-post-title
  - Responsive Design: Mobile-friendly blog listing and post views

**API Support**

  - SOAP API: Full CRUD operations (blogPostList, blogPostInfo, blogPostCreate, etc.)
  - REST API v2: Modern REST endpoints with proper HTTP methods
  - Schema Definitions: Complete WSDL/WSI schemas for client generation
  - All Attributes Exposed: Meta fields, publishing dates, content, etc.
  - Store Filtering: API respects store-specific visibility

PR #109, opened 2025-02-17
Added self-hosted GDPR compliant captcha module

fballiano:
This PR adds a new Maho_Captcha module, which implements self-hosted GDPR compliant captcha based on https://altcha.org. Research was done but Altcha seems to be the most active other open source PoW based captcha project.

At the moment the implementation is almost the same as my Turnstile module: https://github.com/fballiano/openmage-cloudflare-turnstile with a lot of observers and a "css selectors" settings that (IMHO) allows for maximum flexibility.

I called it Maho_Captcha cause I think Maho should provide a basic captcha module and, since this one doesn't rely on 3rd party services (like cloudflare/recaptcha) it seems the perfect candidate.

Questions:
- I used `maho_captcha.(xml|csv)` naming instead of just `captcha.(xml|csv)` because I didn't want it to collide with the old Mage_Captcha. Is this a good choice?
- I used `maho/captcha/footer.html` as folder structure for templates. I don't like that it differs from maho_captcha but at the same time made more sense to have all modules under the `maho/` folder. Is this a good choice?
- Should we apply this naming (or the final one that we decide) to https://github.com/MahoCommerce/maho/pull/103 too? We should have a standard for every new development.
- ~~Since the module positions the captcha widget "just before the ending" of the `form`, this position may not be perfectly aligned, is this a dealbreaker? ideas on how to make it better?~~ activating the "floating" catpcha if works perfectly


app/code/core/Maho/Captcha/Model/Observer.php
```diff
@@ -0,0 +1,96 @@
...
+
+    public function verifyAdmin(Varien_Event_Observer $observer): void
+    {
+        $helper = Mage::helper('maho_captcha');
+        if (!$helper->isEnabled()) {
+            return;
+        }
+
+        $request = Mage::app()->getRequest();
+        if ($request->getActionName() == 'prelogin' || !$request->isPost()) {
```
fballiano, line 59: I don't like this line...

app/code/core/Maho/Captcha/Helper/Data.php
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Maho
+ *
+ * @package    Maho_Captcha
+ * @copyright  Copyright (c) 2025 Maho (https://mahocommerce.com)
+ * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+class Maho_Captcha_Helper_Data extends Mage_Core_Helper_Abstract
+{
```
justinbeaty, line 12:
```suggestion
class Maho_Captcha_Helper_Data extends Mage_Core_Helper_Abstract
{
    protected $_moduleName = 'Maho_Captcha';
```

fballiano:
Converted to draft because of https://github.com/altcha-org/altcha/issues/92

fballiano:
I found a way to work around the limitation of a single captcha per page so I think this PR is testable

fballiano:
As per https://altcha.org/docs/server-integration/

> Replay attacks
> 
> To prevent the vulnerability of “replay attacks,” where a client resubmits the same solution multiple times, the server should implement measures that invalidate previously solved challenges.
> 
> The server should maintain a registry of solved challenges and reject any submissions that attempt to reuse a challenge that has already been successfully solved.

we have to create a database table to store solved challanges and a cleanup routing after a few days (I'd use 7)

fballiano:
I'd added the challenge-log in the captcha_challenge table, with a cron based cleanup routine

justinbeaty:
> we have to create a database table to store solved challanges and a cleanup routing after a few days (I'd use 7)

One thing I don't get is that the challenge expires in `+30 seconds`, so why do we have to keep track of them beyond that long? Is the documentation old or written before the `expires` field was supported?

fballiano:
The thing that is worrying me is that the expire parameter doesn't seem to be baked into the challenge, i'm not sure about that

justinbeaty:
> The thing that is worrying me is that the expire parameter doesn't seem to be baked into the challenge, i'm not sure about that

It seems like it was, I watched the recaptcha widget re-spin on the frontend after around 30 seconds. It just seems to auto update if it knows it's expired.


Also, this commit seems to break the admin login for me. I haven't investigated why yet.
93b55ffc13 using events to solve the single catcha per page limitation

fballiano:
auch, i'll check out backend ASAP, thanks!

fballiano:
I'd say we could lower the cleaning timeout to 1day, would you be ok with merging it then?

fballiano:
but I'd still like to discuss these

- I used maho_captcha.(xml|csv) naming instead of just captcha.(xml|csv) because I didn't want it to collide with the old Mage_Captcha. Is this a good choice?
- I used maho/captcha/footer.html as folder structure for templates. I don't like that it differs from maho_captcha but at the same time made more sense to have all modules under the maho/ folder. Is this a good choice?
- Should we apply this naming (or the final one that we decide) to New feature: blog #103 too? We should have a standard for every new development.

justinbeaty:
> I'd say we could lower the cleaning timeout to 1day, would you be ok with merging it then?

Yeah, I think that's reasonable.

> I kinda think that everything "core" shouldn't have prefix, like it was in magento...

How about `captcha2.(xml|csv)` and `template/captcha2/footer.phtml`? I guess that would mean the code is at `app/code/core/Maho/Captcha2` which is a bit odd since there was no `maho/captcha` v1.

But, I mentioned on Discord a while back maybe at some point we have a class map compatibility layer, so `Mage_Core` could be renamed to `Maho_Core`, but still support references to the old classnames. Might even be able to switch to PSR-4 at the same time. Anyway I only say that all because it might affect the decision here?

fballiano:
> How about `captcha2.(xml|csv)` and `template/captcha2/footer.phtml`? I guess that would mean the code is at `app/code/core/Maho/Captcha2` which is a bit odd since there was no `maho/captcha` v1.

mmmmmmmmm this one I kinda don't like 

> But, I mentioned on Discord a while back maybe at some point we have a class map compatibility layer, so `Mage_Core` could be renamed to `Maho_Core`, but still support references to the old classnames. Might even be able to switch to PSR-4 at the same time. Anyway I only say that all because it might affect the decision here?

It would be interesting but it's probably too early for that?

fballiano:
I mean, if we use captcha.xml etc (without any prefix) we could add a check to `maho health-check` that checks that there's no Mage_Captcha code?

justinbeaty:
> I mean, if we use captcha.xml etc (without any prefix) we could add a check to `maho health-check` that checks that there's no Mage_Captcha code?

Yeah, that's actually a fine idea. After all, you couldn't install two blog modules, two captcha modules, etc without them conflicting.

> It would be interesting but it's probably too early for that?

Yeah, just putting the idea out there incase the legacy `Mage_Captcha` classes would become `\Maho\Captcha` or something. But actually we already removed it so we wouldn't port that namespace.

Anyway, that's the only reason I said `captcha2`. This seems to be the way Mage devs did it with apit vs api2, not that what Mage devs chose over a decade ago bears any relevance on what's a good idea today.

fballiano:
> > I mean, if we use captcha.xml etc (without any prefix) we could add a check to `maho health-check` that checks that there's no Mage_Captcha code?
> 
> Yeah, that's actually a fine idea. After all, you couldn't install two 3rd party blog modules, two captcha modules, etc without them conflicting.

I've renamed everything already, I'll write the health-check now and retest the module installation

fballiano:
ehheheh thanks! :-D

fballiano:
actually there's a problem with the health-check, we already check that no `app/code/core` folder exist, which also means no Mage_Captcha will be present.

But I can't check for a captcha.xml or something like that because now these are our new files and somebody could rightfully override them...

justinbeaty:
Can you parse the module’s config.xml file and look at the version? If you version your module higher than what’s in Mage…

fballiano:
> Can you parse the module’s config.xml file and look at the version? If you version your module higher than what’s in Mage…

no no I was referring to the layout xml files, because the main module file can't exist since app/code/core can't exist (for the health check)

fballiano:
i've a problem with the setup script now... working on it

justinbeaty:
> no no I was referring to the layout xml files, because the main module file can't exist since app/code/core can't exist (for the health check)

I don't understand, shouldn't health-check make sure DOESN'T have `app/design/frontend/base/default/layout/captcha.xml`? Don't we want to prevent overriding these files? Users should use `local.xml`, or use a module to add `vendor.xml`, etc.

fballiano:
> I don't understand, shouldn't health-check make sure users DON'T have `app/design/frontend/base/default/layout/captcha.xml`? Don't we want to prevent overriding these files? Users should use `local.xml`, or use a module to add `my_vendor.xml`, etc.

mmmmm, thinking

because i'm also overriding a couple of files in demo.mahocommerce.com in order to disable stuff and the autoload system manages that perfectly.

wouldn't be a pity to lose it?

justinbeaty:
> mmmmm, thinking
> 
> because i'm also overriding a couple of files in demo.mahocommerce.com in order to disable stuff and the autoload system manages that perfectly.
> 
> wouldn't be a pity to lose it?

Isn't it bad practice though? Because if you override `catalog.xml` then we change something then everyone has to migrate their files. Encouraging it might lead us to question every change to a layout.xml file...

It's like overriding a core class file, no? The autoloader works great for that, too, but it doesn't pass health-check.

fballiano:
mmmm but then what about phtml files? people should create their own theme and override there right?

anyway don't test the module yet cause this new version doesn't work well, I'll work on it asap :-(((

justinbeaty:
> mmmm but then what about phtml files? people should create their own theme and override there right?
> 
> anyway don't test the module yet cause this new version doesn't work well, I'll work on it asap :-(((

I suppose best practice is a child theme, but can we have a critical vs warning for health check? A phtml file is less severe than a core PHP file for sure, layout xml is maybe somewhere in between?

fballiano:
I've to retest a few things so I'll have to postpone this :-\

justinbeaty:
> hey @justinbeaty, I think the PR is at a point that it finally works 🤞🤞🤞

Sweet -- I'll test now.

app/design/frontend/base/default/layout/captcha.xml
```diff
@@ -0,0 +1,15 @@
...
+            <block type="core/template" template="captcha/footer.phtml"></block>
+        </reference>
+    </default>
+    <checkout_onepage_index>
+        <reference name="checkout.onepage.billing">
+            <reference name="form.additional.info">
+                <block type="core/template" template="captcha/default_checkout.phtml"></block>
+            </reference>
+        </reference>
+    </checkout_onepage_index>
```
justinbeaty, line 14: I don't see a `captcha/default_checkout.phtml` file.

public/js/maho-captcha.js
```diff
@@ -0,0 +1,174 @@
+/**
+ * Maho
+ *
+ * @package     Maho_Captcha
+ * @copyright   Copyright (c) 2025 Maho (https://mahocommerce.com)
+ * @license     https://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
+ */
+
+const MahoRecaptcha = {
```
fballiano, line 9: hiihihh what a set of changes :-)))

the only thing is that I wouldn't use the name recaptcha, it could be protected but also a bit misleading, i'd use MahoCaptcha probably, what do you think?

justinbeaty:
I really like this implementation! However I did find a few issues if you submit the form while the captcha is validating (either the first time, or revalidating). You'd get a JS alert that says "please wait..." but actually you need to close the alert and re-submit the form. If you wait as instructed you'll be waiting forever...

image

After you dismiss the alert, you also get a JS error `An invalid form control with name='' is not focusable.`, which should be fixed upstream (I'll create an issue in their repo later).

In any case, what I tried to do (and think it's working) is to add an event listener for form submission, and if the captcha is still validating, show a "loader" and then once it's validated submit the form programmatically.

In `app/code/core/Maho/Captcha/Helper/Data.php`, you can set the `maxNumber` option to `2500000` to be able to test this more easily.

Also note, I updated the "CSS Selectors" to include `#page-login #loginForm` so that the same implementation is shared across frontend / backend. You'll need to add this in your config. If you're locked out, you can add this in `public/js/maho-captcha.js`:
```js
this.frontendSelectors += ', #page-login #loginForm';
```

I actually have to leave the house very shortly, so normally I'd write more and test more thoroughly before pushing. If there's a problem as always we can revert or modify.

public/js/maho-captcha.js
justinbeaty: Oops, you're right, just habit from using Google. 😄

fballiano: renamed it :-)

public/js/maho-captcha.js
```diff
@@ -0,0 +1,174 @@
...
+                    padding: 8px;
+                    border: none;
+                    border-radius: 6px;
+                    box-shadow: 0 0 10px rgba(0, 0, 0, 0.3);
+                    body:has(&) {
+                        overflow: hidden;
+                    }
+                    &::backdrop {
+                        background: rgba(0, 0, 0, 0.7);
+                    }
```
fballiano, line 74: is this necessary? I didn't really see too much difference and to re-create big js files for new functionality scare me.

public/js/maho-captcha.js
```diff
@@ -0,0 +1,174 @@
...
+                new Promise((resolve, reject) => {
+                    const script = document.createElement('script');
+                    script.src = '/js/altcha.min.js';
+                    script.type = 'module';
+                    script.onload = resolve;
+                    script.onerror = () => reject(`${script.src} Not Found`);
+                    document.head.appendChild(script);
+                }),
+            ]);
+        }
```
fballiano, line 31: this loads the altcha script automatically which is something that we have to avoid. my implementation lazy loaded it only when interacting with the form

app/design/frontend/base/default/layout/captcha.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<layout version="0.1.0">
+    <default>
+        <reference name="head">
+            <action method="addJs"><name>maho-captcha.js</name></action>
```
fballiano, line 5: I understand the rationale but I don't know... it's another http request which then loads the altcha script, IMHO we should have avoided another external js :-\

app/code/core/Maho/Captcha/etc/config.xml
```diff
@@ -0,0 +1,187 @@
...
+                <schedule><cron_expr>0 3 * * *</cron_expr></schedule>
+                <run><model>captcha/observer::cleanup</model></run>
+            </captcha_cleanup>
+        </jobs>
+    </crontab>
+    <default>
+        <admin>
+            <captcha>
+                <enabled>1</enabled>
+                <selectors>#page-login #loginForm
```
fballiano, line 175: I would prefer to have this line not customizable, backend selectors IMHO shouldn't be configurable because it's too easy to mess up, people would usually want to customize frontend stuff because they can change it, but backend to me shouldn't be messed with

public/js/maho-captcha.js
fballiano: ok sorry, this is the new feature I couldn't see cause it's not happening to me

app/design/frontend/base/default/layout/captcha.xml
justinbeaty: Hm, but isn't it better to serve an external JS and cache it instead of putting inline JS in the HTML body? The latter adds bytes to every request instead of just once.

app/code/core/Maho/Captcha/etc/config.xml
justinbeaty: Yes, you're right it would be too easy to mess up. Commit a9d1ee7079f4ca3b6fe15df9a8b63b0c6a671e3e is perfect.

fballiano:
I've changed a few things like auto-verification, lazy-load, backend selector.

I don't love that the js became a bit too complex in my opinion and it's a separate http request but probably it's good cause it can be cached.

I think it should be testable again.

app/design/frontend/base/default/layout/captcha.xml
fballiano: if the JS was short short then to me it's better to have it inline, if become too big (but big js come with another set of problems :-() then external

justinbeaty: > it's another http request which then loads the altcha script

Also, I think that's okay because the altcha script doesn't need to be high priority. It just needs to be loaded by the time the user clicks into a form element.

fballiano: we've no way of setting priority to `<script src="http://maho.test/js/maho-captcha.js"></script>` like defer/async right? cause this specific script would need that

public/js/maho-captcha.js
```diff
@@ -0,0 +1,178 @@
...
+        if (!this.altchaWidget) {
+            if (typeof config.widgetAttributes !== 'object' || config.widgetAttributes === null) {
+                throw new Error('widgetAttributes must be specified');
+            }
+            this.altchaWidget = document.createElement('altcha-widget');
+            for (const [ attr, val ] of Object.entries(config.widgetAttributes)) {
+                this.altchaWidget.setAttribute(attr, val);
+            }
+            document.body.appendChild(this.altchaWidget);
+        }
```
fballiano, line 60: if I'm not missing something, this should not be necessary since the `<altcha-widget` is in the phtml itself

app/design/frontend/base/default/layout/captcha.xml
justinbeaty: I meant more so that if altcha.min.js wasn't loaded via XML, that its loading would be deferred via the JS focusin event. We should have a way to set those script attributes via XML, but we don't currently as far as I know.

But is your idea to output directly to the `<head>` something like this?:
```html
<head>
<script src="http://maho.test/js/maho-captcha.js" defer></script>
<script>
document.addEventListener('DOMContentLoaded', () => {
    MahoCaptcha.setup({ ... });
});
</script>
</head>
```

public/js/maho-captcha.js
justinbeaty: That's correct, it was just to support how you were creating the element before. However if in the other comment thread you were talking about putting the template output in the `<head>` element then we'd want to not output `<altcha-widget>` in the phtml and instead use JS.

app/design/frontend/base/default/layout/captcha.xml
fballiano: I did it in https://github.com/MahoCommerce/maho/pull/109/commits/69754ff7d1fe12f3a1f8df8f6fb1c3539d447982, sorry but I've changed the "if helper->enabled" check again

justinbeaty:
+1 to all the changes. Yes, sorry the JS became more complicated. The form submit issue was just annoying UX wise and the fix was more complicated to keep it inline.

I have a few more things to review if that's okay. One thing I was thinking of, could we replace the mysql table (and thus model, resource model, and install scripts) by instead using Mage's cache? Something like this (not tested).

```php
    protected function checkLoggedChallenge(string $payload): bool
    {
        Mage::app()->getCache()->test($payload) !== false;
    }

    protected function logChallenge(string $payload): void
    {
        Mage::app()->getCache()->save(time(), $payload, ['maho_captcha'], 3600);
    }
```

app/code/core/Maho/Captcha/Block/Captcha.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Maho
+ *
+ * @package    Maho_Captcha
+ * @copyright  Copyright (c) 2025 Maho (https://mahocommerce.com)
+ * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+class Maho_Captcha_Block_Captcha extends Mage_Core_Block_Template
```
justinbeaty, line 11: This block isn't used anymore, is it?

app/code/core/Maho/Captcha/Model/Observer.php
```diff
@@ -0,0 +1,113 @@
...
+
+        $token = $data['maho_captcha'] ?? '';
+        if ($helper->verify((string) $token)) {
+            return;
+        }
+
+        $this->failedVerification($controller);
+    }
+
+    public function verifyAjax(Varien_Event_Observer $observer): void
```
justinbeaty, line 32: We should integrate the changes from #113 right?

fballiano:
> I have a few more things to review if that's okay. 

no hurry

> One thing I was thinking of, could we replace the mysql table (and thus model, resource model, and install scripts) by instead using Mage's cache? Something like this (not tested).

I thought about it but I think somebody could need to monitor the size of that table (to see if there are attacks or something like that) and magento's cache is impossible to monitor or, if it grows too much, could it create problems (for example redis with limite resources or file system cache)?

app/code/core/Maho/Captcha/Block/Captcha.php
fballiano: correct! removed in 432771215d ;-)

app/code/core/Maho/Captcha/Model/Observer.php
fballiano: yep absolutely

fballiano: I'm trying to use the isAjax but it's not working :-\

justinbeaty: I pushed a commit that uses the original `<args><is_ajax>1</is_ajax></args>` idea from #113. I haven't had a chance to test it, but I think it's useful?

fballiano: yes it works but I've a couple of other problems, I'll write another comment in a bit

fballiano:
with the "check for verification" feature, for example the newsletter form gets stuck this way:

Screenshot 2025-03-09 alle 20 10 11

doesn't the feature replace the original submit handler for the form?

Another thing is that it seems not to work in the checkout if I click the "continue" button without first focusing the form, I get an "server returned an invalid json" :-\ I'll try to do more debug asap

justinbeaty:
> with the "check for verification" feature, for example the newsletter form gets stuck this way:

I was unable to replicate this. I can't see the state of the captcha widget in the bottom right; is it still verifying?

> doesn't the feature replace the original submit handler for the form?

I don't think so, no. Here's what it does.

```js
    // This is attached as an submit event handler with `capture: true` so that it 
    // fires before the original event handler. If we are not in a verified state,
    // it will prevent the form submission and remember which form we were 
    // trying to submit via the `onVerifiedCallback` var. When we become verified,
    // we will submit the form again with `requestSubmit`, the below form handler
    // will also fire again, but this time we're verified so it doesn't prevent form submission.
    onFormSubmit(event) {
        const formEl = event.target;
        if (this.altchaState !== 'verified') {
            event.preventDefault();
            event.stopPropagation();

            this.showLoader();
            this.onVerifiedCallback = () => {
                this.hideLoader();
                formEl.requestSubmit(event.submitter)
            }
            this.startVerification();
        }
    },
```

> Another thing is that it seems not to work in the checkout if I click the "continue" button without first focusing the form, I get an "server returned an invalid json" :-\ I'll try to do more debug asap

Hm, I'll check it out as well.

fballiano:
niiiiiiice thanks! I've tweaked a few more things, the only time when it's not working it's this situation:

Screenshot 2025-03-10 alle 17 35 58

loading the page when the form data is already filled and pressing the continue button. in this case the captcha scripts are not loaded and thus you get an "incorrect captcha", which is fine but the use can't understand that they have to click inside one of the inputs.. not a good UX...

we should be able to tap into mahoFetch in some way... mmmmmm

public/js/maho-captcha.js
```diff
@@ -0,0 +1,146 @@
...
+    initForms() {
+        for (const formEl of document.querySelectorAll(this.frontendSelectors)) {
+            formEl.addEventListener('focusin', this.loadAltchaScripts.bind(this), true);
+            formEl.addEventListener('submit', this.onFormSubmit.bind(this), true);
+        }
+    },
+
+    async loadAltchaScripts() {
+        // Load Altcha JS
+        await Promise.all([
```
justinbeaty, line 37: The removed if check here makes it so a new `altcha.min.js` is loaded every time you focus in a form:
```js
if (typeof customElements.get('altcha-widget') !== 'function') {
```

image

public/js/maho-captcha.js
```diff
@@ -0,0 +1,146 @@
...
+                script.src = '/js/altcha.min.js';
+                script.type = 'module';
+                script.onload = resolve;
+                script.onerror = () => reject(`${script.src} Not Found`);
+                document.head.appendChild(script);
+            }),
+        ]);
+
+        // Inject Stylesheet
+        const styleEl = document.createElement('style');
```
justinbeaty, line 49: Same with this check: 
```js
if (!document.querySelector('style[maho-captcha-style]')) {
```

fballiano: what about this last version?

justinbeaty:
I added two commits regarding events:

1) I just move the `this.altchaWidget.addEventListener('load', ...)` line to the `setup()` function. It's actually better to attach that event listener earlier, and it doesn't matter if `altcha.min.js` has loaded at that point or not.

2) I added `once: true` to the form's focus in. Even though you added the check to return early from `loadAltchaScripts()`, no reason to fire the event every focusin.

3) I added an event listener to any click events on submit buttons. This fixes the opcheckout issue, testing elsewhere now.

justinbeaty:
Another commit. Sometimes the loading dialog wouldn't show up which was super strange, but I had wanted to add the timeout anyway just so we don't get quick flickers of the dialog. I think maybe some JS executed in `altcha.min.js` might have been causing the issue, so the 200ms delay seems to fix this.

It seems to be working fine on the newsletter form.

app/code/core/Maho/Captcha/etc/config.xml
```diff
@@ -0,0 +1,182 @@
...
+                <schedule><cron_expr>0 3 * * *</cron_expr></schedule>
+                <run><model>captcha/observer::cleanup</model></run>
+            </captcha_cleanup>
+        </jobs>
+    </crontab>
+    <default>
+        <admin>
+            <captcha>
+                <enabled>1</enabled>
+                <selectors>.checkout-onepage-index #billing-new-address-form
```
justinbeaty, line 171:
```suggestion
                <selectors>.checkout-onepage-index #co-billing-form
```

fballiano:
@justinbeaty all my tests are now green ❤️❤️

I'm wondering, should we keep this feature disabled by default instead of enabled like it is now?

Other than that, should we merge?

fballiano:
I'd say that the original Mage_Captcha was not enabled by default, maybe it's a good choice for this module too.

public/js/maho-captcha.js
fballiano: I'd say that altcha.min.js will always be loaded via loadAltchaScripts() so we should be safe there :-)

fballiano:
> I have a few more things to review if that's okay. One thing I was thinking of, could we replace the mysql table (and thus model, resource model, and install scripts) by instead using Mage's cache? Something like this (not tested).
> 
> ```
>     protected function checkLoggedChallenge(string $payload): bool
>     {
>         Mage::app()->getCache()->test($payload) !== false;
>     }
> 
>     protected function logChallenge(string $payload): void
>     {
>         Mage::app()->getCache()->save(time(), $payload, ['maho_captcha'], 3600);
>     }
> ```

the last thing to discuss would be this one about using cache instead of the database, it would be an easier implementation but it scares me a bit because the number of entries could be very big (depending on the store)

justinbeaty:
> the last thing to discuss would be this one about using cache instead of the database, it would be an easier implementation but it scares me a bit because the number of entries could be very big (depending on the store)

If we did use the cache, I think we could lower the TTL down to as low as 60 seconds. That's because if our challenges are only valid for 60 seconds after created by the Helper class, and they client doesn't solve them immediately. Maybe 120 seconds to be safe, which would cut down a lot of entries compared to 1 hour or 1 day.

fballiano:
mmmmm it makes sense (thinking)

fballiano:
Should we handle the situation where cache is disabled?

Il giorno mar 11 mar 2025 alle 16:50 Justin Beaty ***@***.***>
ha scritto:

justinbeaty:
> Should we handle the situation where cache is disabled?

Hm... I would almost say no as it feels like the responsibility of the store owner to make sure it's enabled. How do you feel about it?

fballiano:
Yeah, worse case you can be susceptible to a replay attack in that 1 minute
window

Il giorno mar 11 mar 2025 alle 17:02 Justin Beaty ***@***.***>
ha scritto:

justinbeaty:
> Yeah, worse case you can be susceptible to a replay attack in that 1 minute

I see at least one other Altcha library that only stores the challenges until the expiration date: https://github.com/Flameborn/verity/blob/36c6bff0e5ec97969f13bc62eda5847a90d5ee85/main.go#L79-L81

Another thought, I'd feel safe hashing `$payload` to make our cache key. It's shorter at only 40 chars (could even truncate to 20) compared to `$payload` which is 255 chars.

```php
    protected function checkLoggedChallenge(string $payload): bool
    {
        $key = sha1($payload);
        Mage::app()->getCache()->test($key) !== false;
    }

    protected function logChallenge(string $payload): void
    {
        $key = sha1($payload);
        Mage::app()->getCache()->save(time(), $key, [self::CACHE_TAG], self::CHALLENGE_EXPIRATION);
    }
```

fballiano:
> I see at least one other Altcha library that only stores the challenges until the expiration date: https://github.com/Flameborn/verity/blob/36c6bff0e5ec97969f13bc62eda5847a90d5ee85/main.go#L79-L81

ok :-)

> Another thought, I'd feel safe hashing `$payload` to make our cache key. It's shorter at only 40 chars (could even truncate to 20) compared to `$payload` which is 255 chars.

why do you think it's better to shrink the key?

justinbeaty:
> why do you think it's better to shrink the key?

Solely as an option for space savings.

fballiano:
@justinbeaty I've just pushed a version that removes every database related thing and switches everything to cache, let me know what you think :-)

fballiano:
@justinbeaty I'd try to merge this one and release 25.3 :-)

justinbeaty (commented):
Tested with both `colinmollenhour/cache-backend-redis` and file backend.

Redis is nice because it automatically expires keys after the TTL is up, although the cache's clean method doesn't cause harm.

The file backend works too, I watched the file be removed after the next captcha attempt.

One thing I noticed is that you actually can't disable the cache for captcha solutions, at least not in System > Cache Management (select all and enable). But I think this is fine.

justinbeaty (approved):

fballiano:
nice :-)

> One thing I noticed is that you actually can't disable the cache for captcha solutions, at least not in System > Cache Management (select all and enable). But I think this is fine.

no exactly, but at the same time we kinda don't want for people to manage it :-)

PR #111, opened 2025-02-19
New feature: added possibility to move products to a different attribute set

fballiano:
This feature was requested multiple time and I thought about giving it a look, the first part (assigning products to a new attribute set) was pretty easy but it lead to some questions:
- what about attributes required in the target set that are not present in the source set?
- what about attributes present in the source set but missing in the target set?

1. we could have built an interface to massively set the missing required attributes (in the target set) but.. it's not easy but it already exists... you can either edit the product and save it and you'll be asked to fill the missing attributes. and if you need to do it massively, you already have the "update attributes" mass action. So I wouldn't do anything regarding this topic.

2. this opens an old topic of cleaning up all the EAV tables from leftover data (eg: imported via SQL queries etc) so I thought, instead of just cleaning the attributes of the products that were moved from an attribute-set to another, I'd rather create a small cleaner method that runs via cron and cleans all entities instead of just the products.

This approach allows for
- when we'll have the full EAV editor, there could be more cases of orphaned data
- I noticed that in sample-data there are cases of orphaned data (who knows...)
- changing attribute sets is faster because we are processing the orphaned records asyncronously


app/code/core/Mage/Eav/etc/config.xml
```diff
@@ -132,6 +132,18 @@
             </modules>
         </translate>
     </frontend>
+    <crontab>
+        <jobs>
+            <clean_orphaned_records>
+                <schedule>
+                    <cron_expr>0 2 * * *</cron_expr>
```
fballiano, line 139: I've used the same schedule as the catalog_product_index_price_reindex_all, is that ok?

justinbeaty:
> This feature was requested multiple time [...] assigning products to a new attribute set

Sweet! I'll respond to the questions first then checkout the PR branch.

> what about attributes required in the target set that are not present in the source set?
Maybe we can have an argument in whatever method that controls if (A) an error is thrown if this is the case, or (B) the entity is disabled until required fields are filled.

I think we might need both scenarios for different pages, so the flag would be a good idea. I think the (B) option is best for the default behavior, because to edit a product and save it you'll need the required attributes.

> what about attributes present in the source set but missing in the target set?
Leave as orphaned for the cron job? 

> we could have built an interface to massively set the missing required attributes (in the target set) but [...] it already exists...
Mass update attributes is totally fine, there could eventually be a modal that asks a few options when there's mismatched attribute sets, such as "Cancel", "Change Set and Delete Old", or some sort of way to merge the attribute sets... Not sure just thinking long term, so I agree no need to do anything now.

> I'd rather create a small cleaner method that runs via cron and cleans all entities instead of just the products.

Yeah, that's great. I'll check out the code now a bit, and try to test a bit.

app/code/core/Mage/Eav/etc/config.xml
justinbeaty: I think it should be `0 0 * * *`, the other two events that happen at midnight are `sales_clean_quotes` and `core_email_queue_clean_up`, so it seems the time to clean things up.

I also think midnight is good because at 01:00 we run `catalogrule_apply_all` and at 02:00 we run `catalog_product_index_price_reindex_all`. Since both of those two events certainly use attributes, best not do the cleaning at the same time.

fballiano:
> > what about attributes required in the target set that are not present in the source set?
> > Maybe we can have an argument in whatever method that controls if (A) an error is thrown if this is the case, or (B) the entity is disabled until required fields are filled.

yes but to me it becomes too complicated and eveybody may be expecting a different behavior 😅
In my opinion, the main use case for this feature is "oh shit I created the product in the wrong attribute set" and in this case we don't need as many safe nets.

Disabling the product is interesting but I kinda think that it would make somebody angry more than having to fix the missing attributes (that is needed anyway).

app/code/core/Mage/Eav/etc/config.xml
fballiano, line 139:
```suggestion
                    <cron_expr>0 0 * * *</cron_expr>
```

app/code/core/Mage/Eav/etc/config.xml
```diff
@@ -132,6 +132,18 @@
             </modules>
         </translate>
     </frontend>
+    <crontab>
+        <jobs>
+            <clean_orphaned_records>
```
justinbeaty, line 137:
```suggestion
            <eav_clean_orphaned_records>
```

All other module's cron events seem to prefix with module name, i.e. `core_clean_cache`, `customer_flowpassword`, `newsletter_send_all`, `sales_clean_quotes`.

justinbeaty:
> Disabling the product is interesting but I kinda think that it would make somebody angry more than having to fix the missing attributes (that is needed anyway).

Were you thinking the user is redirected to the existing mass attribute update form, or just an error message is thrown?

fballiano:
> > Disabling the product is interesting but I kinda think that it would make somebody angry more than having to fix the missing attributes (that is needed anyway).
> 
> Were you thinking the user is redirected to the existing mass attribute update form, or just an error message is thrown?

since I wouldn't even check if data is really missing or not, I'll leave that to the user, there's already a notice before starting

fballiano:
I've committed the suggestions, for some reason github was broken and couldn't apply the suggestion from the website

Screenshot 2025-02-21 alle 23 43 12

justinbeaty:
> since I wouldn't even check if data is really missing or not, I'll leave that to the user, there's already a notice before starting

Hm, but then we need to test `catalogrule_apply_all`, `catalog_product_index_price_reindex_all`, and other things that might happen on products with missing attributes. Even if the intention is only for newly created products, people will certainly use it in all sorts of ways, then forget to update the product, and things will break.

What if when you press submit, we precheck via AJAX to know the number of products that have missing attributes and show a Maho Dialog that says:

> Selected products are missing required attributes from the target set
>  [Cancel] [Continue and disable 42 products] [Continue and mass edit]

None of those options needs to be in this PR, just thinking what's the next move. It would be nice to have some feedback on the number of items affected, though.

fballiano:
> > What if when you press submit, we precheck via AJAX to know the number of products that have missing attributes and show a Maho Dialog that says:
> 
> we could say that in the confirmation notice that says "xxx products updated successfully" as you said and it should be pretty easy too :-)

mmm seems to be not that easy and not that quick to do such a check...

fballiano:
the routine to check for products with missing required attributes is extremely complicated, so far I have:

```php
protected function checkMissingDataFromMandatoryAttributes()
    {
        $coreResource = Mage::getModel('core/resource');
        $readConnection = $coreResource->getConnection('core_read');
        $requiredAttributesBySetAndType = [];
        $requiredAttributes = $readConnection->fetchAll(
            "SELECT eas.attribute_set_id, ea.attribute_id, ea.backend_type
                FROM " . $coreResource->getTableName('eav/attribute') . " ea
                JOIN " . $coreResource->getTableName('catalog/eav_attribute') . " cea ON ea.attribute_id = cea.attribute_id
                JOIN " . $coreResource->getTableName('eav/entity_type') . " eet ON ea.entity_type_id = eet.entity_type_id
                JOIN " . $coreResource->getTableName('eav/attribute_set') . " eas ON eet.entity_type_id = eas.entity_type_id
                WHERE eet.entity_type_code = 'catalog_product'
                AND ea.is_required = 1 AND ea.backend_type <> 'static' AND cea.is_visible = 1
                ORDER BY attribute_set_id, attribute_id, backend_type");
        foreach ($requiredAttributes as $attribute) {
            $requiredAttributesBySetAndType[$attribute['attribute_set_id']][$attribute['backend_type']][] = $attribute['attribute_id'];
        }

        $brokenProducts = [];
        $productTable = $coreResource->getTableName('catalog/product');
        foreach ($requiredAttributesBySetAndType as $attributeSetId => $attributes) {
            foreach ($attributes as $backendType => $attributeIds) {
                $attributeDataTable = $coreResource->getTableName('catalog_product_entity_' . $backendType);
                $attributeIds = implode(',', $attributeIds);
                $results = $readConnection->fetchCol(
                    "SELECT DISTINCT cpe.sku 
                    FROM $productTable cpe
                    LEFT JOIN $attributeDataTable cpev ON (
                    cpe.entity_id = cpev.entity_id 
                    AND cpev.attribute_id IN ($attributeIds)
                    )
                    WHERE cpe.attribute_set_id = $attributeSetId
                    AND cpev.value_id IS NULL");
                $brokenProducts = [...$brokenProducts, ...$results];
            }
        }
        $brokenProducts = array_unique($brokenProducts);
        $brokenProducts = array_values($brokenProducts);
        natcasesort($brokenProducts);

        $numOfBrokenProducts = sizeof($brokenProducts);
        $brokenProducts = array_slice($brokenProducts, 0, 50);
        $brokenProducts = implode(', ', $brokenProducts);

        if ($numOfBrokenProducts) {
            $this->_getSession()->addWarning(
                $this->__('%d product(s) do not have any data for some required attributes: %s.', $numOfBrokenProducts, $brokenProducts),
            );
        }
    }
```

but I just found out that I've to add the check for attributes that are enabled only for some specific product type (eg: weight is required but exists only for simple and bundle).

To me this is too complicated, cumbersome and error prone. I've used almost a day to come to this point but I think it is not worth. The result will never be perfect and it could become slow (depending on the size of the database).

I think this should be scratched and the PR should be merged as it is now.

fballiano:
@justinbeaty is it ok it we merge it the way it is?

fballiano:
Thanks so much for this changes!!

merging it now!

PR #112, opened 2025-02-19
Added SMTP support

fballiano:
While working in this idea I noticed that in app/code/core/Mage/Core/etc/system.xml there were already options for smtp login, very interesting.

Anyway, as a first implementation this could be enough, it's very basic but, at the end of the day, should cover everything.


fballiano:
@justinbeaty idea, what if we migrate the mail component to symfony/mailer? it would give access to a lot of new implementations: https://symfony.com/doc/current/mailer.html

justinbeaty:
> @justinbeaty idea, what if we migrate the mail component to symfony/mailer? it would give access to a lot of new implementations: https://symfony.com/doc/current/mailer.html

That would be nice, since it supports inline CSS it should even be able to replace emogrifier I think. Plus thing slike attachements which currently you have to override models to do.

But, I've tried adding Symphony components before and it didn't seem to work great because of config files needed like ` config/packages/mailer.yaml`, but this is probably only a problem for some packages.

Otherwise I bet there's a few other small dependency libraries for delivering mail.

fballiano:
mainly i'd like to be able to support other providers simply installing symfony/amazon-mailer or more... it would be very nice.

I didn't know about the yaml file, I'll check that

fballiano:
@justinbeaty the CSS inlining is available only as a twig plugin, so it's no fit for us now.

justinbeaty:
> @justinbeaty the CSS inlining is available only as a twig plugin, so it's no fit for us now.

Got it, that's okay but it seems you got it working. :+1:

fballiano:
email_template seems to be working (i've added a Maho CLI command to test it), but I've yet to test the newsletter.
also, i've now added all 3rd party providers but it would be nice to add some info of what package needs to be installed (if it's not)

fballiano:
I've added the check for the installed packages and, to me, it's pretty impressive

Screenshot 2025-02-22 alle 11 13 44

I've also tested the "queue" system and it seems to work fine

fballiano:
If there are no opinions agains this, I'd merge it as it is by tomorrow

justinbeaty:
> I've added the check for the installed packages and, to me, it's pretty impressive

This is super impressive! Checking it out now.

app/code/core/Mage/Newsletter/Model/Resource/Queue/Collection.php
```diff
@@ -43,10 +43,6 @@ protected function _construct()
     }
 
     /**
-     * Joines templates information
-     *
-     * @deprecated since 1.4.0.1
-     *
```
justinbeaty, line 49: How come you just deleted the deprecated message but not the method?

justinbeaty:
Code review wise this looks really clean. I haven't tested it with the various plugins like `Aschroder_SMTPPro` or `Aschroder_Email`, but presumably we can replace these modules that were often required and build the features into core.

app/code/core/Mage/Newsletter/Model/Resource/Queue/Collection.php
fballiano: yep that module is used in a couple of place and I didn't see a replacement

Screenshot 2025-02-25 alle 18 46 43

fballiano:
> Code review wise this looks really clean. I haven't tested it with the various plugins like `Aschroder_SMTPPro` or `Aschroder_Email`, but presumably we can replace these modules that were often required and build the features into core.

I never heard of Aschroder_Email, but the idea of this module is to provide the features of Aschroder_SMTPPro (actually more since symfony/mailer does encryption autodetect and it supports many more providers) out of the box. A modern platform can't not having smtp support

justinbeaty:
> Aschroder_Email

It was a paid Amazon SES module. I can't seem to even find the module's page online, but in any case it would also be replaced by this PR. We can eventually enhance the email log, problem report, etc and make all 3rd party email modules obsolete. 😃

This PR looks good as it is. It's a bit hard for me to test this feature at the moment, but if you think it is good then let's merge and move forward.

fballiano:
The submodules should be compatible with the main interface of symfony/mailer so I think we're quite safe there. I tested a couple of them and it all seems to work, I'll do some more tests on the demo website ;-)

We could add the email log for sure (although there's the core_email table which already is a log, just missing a view on the backend) but I think this feature is important for the future :-)

fballiano:
(but actually all modern smtp solutions have web interfaces with all the info... so there's not a big need for duplicating it in our backend)

justinbeaty:
I don't see an existing `core_email_log` table in our database, unless I'm blind.

How I implemented it was to create an email log table with recipient, subject line, date, etc. But I didn't want to store message data in it, so instead I gzip and store the file in `var/email_log/123.gz`. Then I can just read that gzip file in the Admin > System > Tools > Email Log screen I have.

```php
    // Send email...

    // Write the email to the filesystem compressed
    $output = Mage::getBaseDir('var') . DS . 'email_log' . DS . $log->getId() . '.gz';

    $fp = gzopen($output, 'w9');
    gzwrite($fp, $isHtml ? $email : nl2br($email));
    gzclose($fp);
```

fballiano:
sorry, it's core_email_queue, it's only used when the message queue is enabled, but it should be the default

justinbeaty:
> sorry, it's core_email_queue, it's only used when the message queue is enabled, but it should be the default

Ah, yeah but that is purged after being sent. I am a fan of being able to search the historical Email Log and tell customers exactly what the email contained. I also have a resend button which is nice. I have history going back years and it takes very little space in the DB and also very little disk space due to gzip.

PR #114, opened 2025-02-28
Fix `Mage_Customer_Helper_Address::getAttributeValidationClass()` method

justinbeaty:
Reference https://github.com/MahoCommerce/maho/pull/106#issuecomment-2691133661
>**validation**, for example in the customer registration form, if you try to submit, some fields get validated some others no
> Screenshot 2025-02-28 alle 16 12 15
>
> I found out that's not because of this PR but because of the code in this method:
> ```php
> public function getAttributeValidationClass($attributeCode)
>     {
>         /** @var Mage_Customer_Model_Attribute $attribute */
>         $attribute = $this->_attributes[$attributeCode] ?? Mage::getSingleton('eav/config')->getAttribute('customer_address', $attributeCode);
>         $class = $attribute ? $attribute->getFrontend()->getClass() : '';
> 
>         if (in_array($attributeCode, ['firstname', 'middlename', 'lastname', 'prefix', 'suffix', 'taxvat'])) {
>             if ($class && !$attribute->getIsVisible()) {
>                 $class = ''; // address attribute is not visible thus its validation rules are not applied
>             }
> 
>             /** @var Mage_Customer_Model_Attribute $customerAttribute */
>             $customerAttribute = Mage::getSingleton('eav/config')->getAttribute('customer', $attributeCode);
>             $class .= $customerAttribute && $customerAttribute->getIsVisible()
>                 ? $customerAttribute->getFrontend()->getClass() : '';
>             $class = implode(' ', array_unique(array_filter(explode(' ', $class))));
>         }
> 
>         return $class;
>     }
> ```
> the $class becomes "required-entryrequired-entry" with no space in between and it breaks the validation. at the moment I can't work on a fix.

### DIFF
--- a/app/code/core/Mage/Customer/Helper/Address.php
+++ b/app/code/core/Mage/Customer/Helper/Address.php
@@ -6,6 +6,7 @@
  * @package    Mage_Customer
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright  Copyright (c) 2020-2024 The OpenMage Contributors (https://openmage.org)
+ * @copyright  Copyright (c) 2025 Maho (https://mahocommerce.com)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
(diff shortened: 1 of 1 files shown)


fballiano:
beautiful ❤️

PR #115, opened 2025-03-01
Added a few rector rules

fballiano:
I couldn't add the whole set of UP_TO82 because it was generating some errors, but I think it's nice to start going in that direction maybe step by step.


fballiano:
@justinbeaty would you agree with these changes?

justinbeaty (approved):
I love them all!

fballiano:
and magically https://github.com/OpenMage/magento-lts/pull/4682 appears 😂

PR #119, opened 2025-03-04
Improvements to Flatpickr adapters

justinbeaty:
First commit (e9f2dc0a5eebc59303e2b1fbc6084fd415683a39) is a backup of the commit dropped from #106.

Second commit (db1d96b6698c1b3e946d4e6decfaeb04f0859bce) was me working on improving our date picker:

### Use flatpickr time selector
Product custom options has date, datetime, and time options. Below is how they appear on main:
image

Below is how the latter two now appear with this PR:
image

Luckily, there is no concept of timezones in the Magento product option, it's just a time (i.e. "09:00") that gets converted to 24h format for the DB. But actually I didn't finish the saving time values to the DB, that needs to be done in [this file](app/code/core/Mage/Catalog/Model/Product/Option/Type/Date.php), currently they used some regex but maybe there's a better way to parse it.

### Fix `<depends>` config node

In **System > Config > Catalog > Date & Time Custom Options**, all the config nodes depended on the JS calendar being enabled, but this doesn't make sense as all the options are used when the JS is disabled.

### `Mage_Core_Block_Html_Date` updates

I know you didn't touch this class much, but we will need to break some compatibility here to enable time only inputs. Mage has an `enableTime` option, but it doesn't have a way to enable / disable the date picker separately from the time input.

So far I've added a new `setConfig()` method intended to control flatpickr options not present on the current PHP class. I think we can expand this class to support flatpickr directly instead of hacking around it in JS. For example, can we convert to flatpickr's format in PHP instead of JS? Right now the format is under `config.ifFormat`, so maybe add support for `config.fpFormat`. Maybe detect which one is present in JS act support both?

### JS updates
Flatpickr wasn't working great on mobile for me, and it was caused by the instance not loading until the input has a focus or click event, and which time flatpickr would load, detect mobile, and turn it into a native input. I changed this so that inputs are initialized during the `Calendar.setup()` call.

I also had to fix one character in `strftimeToDateConvertionMap` related to AM / PM. That was `'%p': 'A'` -> `'%p': 'K'`. It seems flatpickr uses it's own slightly custom set of [formatting tokens](https://flatpickr.js.org/formatting/).

Note: there's a temp debug line in the JS to set the format to `m/j/Y h:i K`. But it seems that format is valid for all of date, datetime, and time pickers...


justinbeaty:
@fballiano -- just tagging you again in case you didn't see this.

fballiano:
Sorry, I thought I could finish the captcha more quickly and then answer here but...
I'll still check this out after the captcha cause I want to have it finished sooner than later, then it's probably time for 25.3 release and then I'd work on this one if that's ok with you.
But I'll also have to travel to phpcon netherlands later this month so it's gonna be a busy march 😅

justinbeaty:
No problem, just making sure you saw it. Worst case, we just add the one line to support the dialog usage before release:

```js
const flatpickrOptions = {
    // ...
    static: !!inputEl.closest('dialog'),
}
```

fballiano:
> can we convert to flatpickr's format in PHP instead of JS? Right now the format is under `config.ifFormat`, so maybe add support for `config.fpFormat`. Maybe detect which one is present in JS act support both?

mmm I lean toward thinking that I'd prefer to keep the php as agnostic as possibile, in the sense that these js libraries come and go and the less we're tied to them the more we're free. and to have it both in php and js IMHO could bring more possible bugs if we change one and forget the other.

> Flatpickr wasn't working great on mobile for me, and it was caused by the instance not loading until the input has a focus or click event, and which time flatpickr would load, detect mobile, and turn it into a native input. I changed this so that inputs are initialized during the Calendar.setup() call.

I can't test the backend "for real" cause it seems I can't login via passkey on mobile, very weird.

But It seems to me you already did everything in this PR right?

fballiano:
@justinbeaty should I continue here?

justinbeaty:
> @justinbeaty shall I continue here?

Hey @fballiano, yes I would be more than happy for you to take this one over. I apologize if that wasn't clear in the messages from March 5th.

fballiano:
Hey @justinbeaty, I think I've fixed datetime and time selectors, is that it?

justinbeaty:
> Hey @justinbeaty, I think I've fixed datetime and time selectors, is that it?

I'll check in a couple of hours when I'm back home, but I do recall there was also an issue on mobile browsers. Specifically I was using chrome on iOS, not sure if you've checked that.

fballiano:
will check that asap ;-)

fballiano:
tested it now with on a chrome based browser and it worked fine

justinbeaty:
I am seeing some problems with configurable products. The time field isn't being set, and the date time is displaying wrong. I'll see if I can figure out a solution.

image

image

justinbeaty:
I remember why I had stopped working on this now... 😅 We needed to implement something like the last commit, and I was pretty tired of working on custom product options after #106.

It is almost 100% working. I'll finish and explain everything tomorrow since it's getting late here.

justinbeaty:
And actually I think we might need to remove some more Varien_Date stuff, too.

justinbeaty:
> the only thing is that we have quite some date format conversions, in Varien_Date and now in javascripts too, but I guess we can't do without right?

Yeah, there are way too many different date formats. But ultimately it seems we want to consolidate on ICU formats, right?

> what I don't love is to write too much code custom for flatpickr, since it's semi-abandoned state :-\

I think our only flatpickr specific code is in the `Calendar` JS code, which of course is already where we initialize the flatpickr instances. So that should be okay.

We can migrate to another JS library if needed, it just needs to support two main features:
1. Have both a display format (what's shown to the users) and an input format (what's sent to the server)
2. Support Date, Datetime, and Time inputs.

> btw if you think this is ready for merging, let's go

The only question I had was that since I changed the tokens in `Mage_Core_Model_Local::getTimeFormat()`, did anything break. I originally thought it would require more work, but then I figured out Zend_Date supports those tokens, too. I should have some time to test this tomorrow so we can be sure before we merge.

fballiano:
> Yeah, there are way too many different date formats. But ultimately it seems we want to consolidate on ICU formats, right?

I'd say yes 
 
> We can migrate to another JS library if needed, it just needs to support two main features:

it can wait for now :-)

> The only question I had was that since I changed the tokens in `Mage_Core_Model_Local::getTimeFormat()`, did anything break. I originally thought it would require more work, but then I figured out Zend_Date supports those tokens, too. I should have some time to test this tomorrow so we can be sure before we merge.

oki :-)

justinbeaty:
Okay, so the problem does run a bit deeper. The issue basically stems from the old calendar only supporting one format, which is both displayed to the user, and sent to the server. To handle form submissions with a localized date, there's the following code in the base controller class:

https://github.com/MahoCommerce/maho/blob/635488aab0d1d9ad146c5f8c4661fe2a4a9241c4/app/code/core/Mage/Core/Controller/Varien/Action.php#L970-L991

But since flatpickr (and hopefully other modern date pickers) support a display format and an internal format, this is completely pointless.

What I think we should do is force the internal format to always be `yyyy-MM-dd HH:mm:ss` (or a portion of that for just date or just time inputs). That means what is sent to the server will always be in that format. The only thing the Maho user could control is the display format.

Then we either remove or stub out those functions in the controller. What do you think @fballiano?

justinbeaty:
> I'm wondering, what if somebody writes the date in the field instead of selecting it from the picker?

If it's a flatpickr field, I'm pretty sure it should still send the internal format. But I haven't tested that.

fballiano:
I tried removing the the _filterPostData from Mage_Adminhtml_Cms_PageController and the dates started working very poorly so mmmmmm

fballiano:
@justinbeaty do you think https://github.com/MahoCommerce/maho/pull/119/commits/9325e7c9af6b5115599025f35bd4de1978821c91 makes any sense?

justinbeaty:
> @justinbeaty do you think [9325e7c](https://github.com/MahoCommerce/maho/commit/9325e7c9af6b5115599025f35bd4de1978821c91) makes any sense?

Yes, it does! That's exactly what I was thinking! 

I just read the code though, does everything seem to work? Because I think that was the only blocker to this PR.

fballiano:
@justinbeaty I'm a bit confused on this now because i'm working on a million different things and I don't even know anymore 😅

justinbeaty:
> @justinbeaty I'm a bit confused on this now because i'm working on a million different things and I don't even know anymore 😅

Lol, yeah this PR seemed really simple, but it's sort of a fundamental design of old Magento. But what you did was exactly what I thought the solution should be. Simply show the user one format, but always PSOT the internal one.

PR #120, opened 2025-03-05
Deprecated `Mage_Core_Helper_Data::decorateArray()`

justinbeaty:
I removed most of these in #106, but one remained that didn't have anything to do with configurable products, so it's here in a separate PR.

The `Mage_Core_Helper_Data::decorateArray()` method adds `decorated_is_[first|odd|even|last]` values to an array of varien objects. This was used for adding CSS classes to `<li>`, `<dd>`, and other elements.


fballiano:
absolutely

PR #122, opened 2025-03-05
Removed `Mage_Adminhtml_Block_Notification_Security` Block

justinbeaty:
This block checks to see if `http://example.com/app/etc/local.xml` is accessible from the internet, and displays the message:
*"Your web server is configured incorrectly. As a result, configuration files with sensitive information are accessible from the outside. Please contact your hosting provider."*

Since this is not possible in Maho, let's remove it. 😎


fballiano:
yes!!!

PR #123, opened 2025-03-05
Removed `<em>*</em>` from frontend templates

justinbeaty:
I removed all `<em>*</em>` nodes from `<label class="required">`, since they're hidden by CSS anyway...

Also some JS that liked to create these elements if not exist. f4ed5ad55c9537072f3b91c4056859d7f1ce7a3a

### DIFF
--- a/app/design/frontend/base/default/template/checkout/cart/shipping.phtml
+++ b/app/design/frontend/base/default/template/checkout/cart/shipping.phtml
@@ -5,7 +5,7 @@
  * @package     base_default
  * @copyright   Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright   Copyright (c) 2021-2024 The OpenMage Contributors (https://openmage.org)
- * @copyright   Copyright (c) 2024 Maho (https://mahocommerce.com)
+ * @copyright   Copyright (c) 2024-2025 Maho (https://mahocommerce.com)
  * @license     https://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
  */
 ?>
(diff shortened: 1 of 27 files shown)


fballiano:
could this affect accessibility maybe? cause the * is now visible only via CSS thus I think it's not visible to screen readers (dunno if it should be or I'm just overthinking)

fballiano:
asking perplexity I get this

Screenshot 2025-03-05 alle 17 54 59

I'm not sure if we're really losing some accessibility removing the "em"s here.

But actually I'm reading the the "*" should have aria-hidden so...

justinbeaty:
> could this affect accessibility maybe? cause the * is now visible only via CSS thus I think it's not visible to screen readers (dunno if it should be or I'm just overthinking)

Regarding accessibility, IMO this is an improvement for screen readers, because before they may have announced something like "First name asterisk input field". Maybe screen readers are smart enough to omit that fluff, but the asterisk is styling so I think it is okay in CSS.

What we should do instead is use the `<input>`'s `required` attribute (or `aria-required` which I'm not sure the difference, and that's a rabbit hole.) That would probably coincide with a refactor of the varienForm JS classes. There's been a lot of updates to HTML inputs sincetso long ago...

justinbeaty:
> But actually I'm reading the the "*" should have aria-hidden so...

Yes, that should have been the case if they were displayed. Actually, do screen readers speak hidden elements anyway? All of these removed `<em>`s have `display: hidden` on them anyway...

fballiano:
https://demo.mahocommerce.com updated with latest main

PR #124, opened 2025-03-05
Replaced `adminhtml/system_config_source_yesno` with `boolean`

justinbeaty:
One more thing on my random todo list. 

First in f04e1491248374530c4d47636b3032cce2e3abbf we add a simple Varien form type which we can use in several places:

```php
// Admin form
$fieldset->addField('twofa_enabled', 'boolean', [
    'label' => Mage::helper('adminhtml')->__('Enable 2FA'),
    'name' => 'twofa_enabled',
]);
```

```xml
<!-- system.xml -->
<product_use_categories translate="label">
    <label>Use Categories Path for Product URLs</label>
    <frontend_type>boolean</frontend_type>
    <!-- ... -->
```

Then in the second commit we replace a bunch of usages with the new boolean type.


fballiano:
it seems quite about right :-)

justinbeaty:
I accidentally left `tmp.txt` in the root when I did `git add -A`... Removed in 4069d6a2c5207ba0e63923cdace4d9e57f4799ee.

PR #126, opened 2025-03-11
Removed possibility to disable CSRF in the backend

fballiano:
@justinbeaty I think in 2025 nobody should disable CSRF.

I wanted to also remove the "<default>" configuration value but that could break some stuff (in case you have a configuration saved in core_config_data) in custom code so it's probably better to leave it?

or should we remove the core_config_data in an upgrade script?

### DIFF
--- a/app/code/core/Mage/Core/etc/system.xml
+++ b/app/code/core/Mage/Core/etc/system.xml
@@ -28,23 +28,6 @@
     <sections>
         <system>
             <groups>
-                <csrf translate="label" module="core">
-                    <label>CSRF protection</label>
-                    <sort_order>0</sort_order>
-                    <show_in_default>1</show_in_default>
-                    <show_in_website>1</show_in_website>
-                    <show_in_store>1</show_in_store>
-                    <fields>
-                        <use_form_key translate="label">
-                            <label>Enabled</label>
-                            <frontend_type>boolean</frontend_type>
-                            <sort_order>10</sort_order>
-                            <show_in_default>1</show_in_default>
-                            <show_in_website>1</show_in_website>
-                            <show_in_store>1</show_in_store>
-                        </use_form_key>
-                    </fields>
-                </csrf>
                 <rate_limit translate="label comment" module="core">
                     <label>Rate limit</label>
                     <sort_order>10</sort_order>


justinbeaty:
Totally agree to remove the option to disable.



> I wanted to also remove the "" configuration value but that could break some stuff (in case you have a configuration saved in core_config_data) in custom code so it's probably better to leave it?
> 
> or should we remove the core_config_data in an upgrade script?

I would probably do this because I think that code should be using this helper method instead of reading the config directly.
```diff
# app/code/core/Mage/Core/Helper/Data.php

     public function isFormKeyEnabled(): bool
     {
-        return Mage::getStoreConfigFlag(Mage_Core_Controller_Front_Action::XML_CSRF_USE_FLAG_CONFIG_PATH);
+        return true;
     }
```

But we could also set it to `1` in an install script too. Hmm...

app/design/adminhtml/default/default/layout/main.xml
```diff
@@ -77,7 +77,6 @@ Default layout, loads most of the pages
             <block type="core/text_list" name="notifications" as="notifications">
                 <block type="adminhtml/notification_baseurl" name="notification_baseurl" as="notification_baseurl" template="notification/baseurl.phtml"></block>
                 <block type="adminhtml/cache_notifications" name="cache_notifications" template="system/cache/notifications.phtml"></block>
-                <block type="adminhtml/checkout_formkey" name="checkout_formkey" as="checkout_formkey" template="notification/formkey.phtml"/>
```
fballiano, line 80: I kept it because somebody could still disable it via config, should we remove that?

fballiano:
> ```diff
> # app/code/core/Mage/Core/Helper/Data.php
> 
>      public function isFormKeyEnabled(): bool
>      {
> -        return Mage::getStoreConfigFlag(Mage_Core_Controller_Front_Action::XML_CSRF_USE_FLAG_CONFIG_PATH);
> +        return true;
>      }
> ```

great idea, maybe we should also remove the constant so that people that use it will get an error and recognize they've to fix something?

justinbeaty:
> great idea, maybe we should also remove the constant so that people that use it will get an error and recognize they've to fix something?

Removing the constant sounds good too. Would that resolve https://github.com/MahoCommerce/maho/pull/126#discussion_r1989341599 since it would be impossible to enable it any more?

fballiano:
I've remove quite some more stuff, what do you think?

justinbeaty:
> I've remove quite some more stuff, what do you think?

Reviewing online, it looks perfect!

fballiano:
Maybe we can merge this after 25.3 is released and leave it for the next one?

justinbeaty:
> Maybe we can merge this after 25.3 is released and leave it for the next one?

Yeah, that seems like a good idea.

fballiano:
@justinbeaty I've also done the same thing for the backend

justinbeaty:
> @justinbeaty I've also done the same thing for the backend

The admin form key is actually the secret key in the URL, i.e. the `key` part of `https://example.com/admin/dashboard/index/key/cae7e7e7a8f452b89bf33dfe78ba38f9/`.

I actually do disable this in my store because I have admin emails sent to me with direct links to places in the admin, which doesn't work with the URL secrets. There's no risk in my case because I have the entire admin under basic auth. That doesn't mean I'm against removing it though.

fballiano:
mmmm let's leave it them, let me revert that part of the PR :-)

justinbeaty (approved):
Re-tested and all looks good!

fballiano:
sorry @justinbeaty, I shouldn't have changed the PR last minute, I thought frontend/backend could follow the same rule but the option for the backend makes sense

PR #130, opened 2025-03-15
Added the possibility to disable advanced search

fballiano:
@justinbeaty I've rewritten [this commit](https://github.com/MahoCommerce/maho/commit/770d0eeff9e96b25fa58da8a4dc0d8e18f7f432a) completely and I think it's much simpler and more integrated.

### DIFF
--- a/app/code/core/Maho/Captcha/etc/system.xml
+++ b/app/code/core/Maho/Captcha/etc/system.xml
@@ -12,8 +12,7 @@
                     <fields>
                         <enabled translate="label">
                             <label>Enable Captcha</label>
-                            <frontend_type>select</frontend_type>
-                            <source_model>adminhtml/system_config_source_yesno</source_model>
+                            <frontend_type>boolean</frontend_type>
                             <sort_order>10</sort_order>
                             <show_in_default>1</show_in_default>
                             <show_in_website>1</show_in_website>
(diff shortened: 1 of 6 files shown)


justinbeaty:
Nice! I tested disabling at different store views and it all works perfectly.

I have phpstan running now since the actions are down, maybe you already ran though.

app/code/core/Maho/Captcha/etc/system.xml
```diff
@@ -12,8 +12,7 @@
                     <fields>
                         <enabled translate="label">
                             <label>Enable Captcha</label>
-                            <frontend_type>select</frontend_type>
-                            <source_model>adminhtml/system_config_source_yesno</source_model>
+                            <frontend_type>boolean</frontend_type>
```
fballiano, line 15: i also fixed a couple of system_config_source_yesno that I've seen just by chance

justinbeaty:
> phpstan clean and phpcs too :-)

Yup, but missing copyright, otherwise :+1:

app/code/core/Maho/Captcha/etc/system.xml
justinbeaty: Thank you!

fballiano:
hihihhih right!

justinbeaty (approved):

PR #131, opened 2025-03-16
Added "redirect to product page if search generates one result" feature

fballiano:
this feature was in present in many seo modules out there. [good idea they had to implement in in openmage](https://github.com/OpenMage/magento-lts/pull/4697), although I've rewritten it completely because I think it should have been done in a different way, so here it is.

### DIFF
--- a/app/code/core/Mage/CatalogSearch/controllers/ResultController.php
+++ b/app/code/core/Mage/CatalogSearch/controllers/ResultController.php
@@ -6,6 +6,7 @@
  * @package    Mage_CatalogSearch
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright  Copyright (c) 2020-2024 The OpenMage Contributors (https://openmage.org)
+ * @copyright  Copyright (c) 2025 Maho (https://mahocommerce.com)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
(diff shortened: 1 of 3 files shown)


fballiano:
the approach in the [first version](https://github.com/OpenMage/magento-lts/pull/4697/commits/f7e70cceedf3576ed36ce1c8ffa59109acae84f1) of https://github.com/OpenMage/magento-lts/pull/4697, I didn't even check the small part of the code that's somehow similar because I didn't like the way it was done and I re-thought it completely, I didn't only change the variable names, I rewrote it and that's why some notations are different, I don't check for the resulting url, etc.

there are not a lot of ways of checking if a collection has 1 record, load it and redirect it (which btw was done completely differently in the OM's one, not correctly in my personal opinion).

for the sake of the infinite controversy I've anyway added openmage copyright to this PR, which they don't usually do, although I make sure to port all their "copyright" only related PR.

Because of all these reasons I will merge this PR as it is.

fballiano:
Luckily there's PHPStorm local file versioning, that shows that I wrote this PR character by character, I'm attaching only a few sample screenshots, you can see the timeline on the left and the amount of entries (showing the amount of changes I did).

I had to make a lot or trial&error cause I didn't think I had to use the layout->getBlock, but in other ways I couldn't get the right number of elements (which btw is very weird)

Screenshot 2025-03-20 alle 16 39 01

Screenshot 2025-03-20 alle 16 39 11

Screenshot 2025-03-20 alle 16 39 55

Screenshot 2025-03-20 alle 16 40 06

Screenshot 2025-03-20 alle 16 40 17

Screenshot 2025-03-20 alle 16 40 36

Screenshot 2025-03-20 alle 16 40 43

Screenshot 2025-03-20 alle 16 40 48

Screenshot 2025-03-20 alle 16 40 54

Screenshot 2025-03-20 alle 16 41 00

Screenshot 2025-03-20 alle 16 41 05

Screenshot 2025-03-20 alle 16 41 13

Screenshot 2025-03-20 alle 16 41 17

Screenshot 2025-03-20 alle 16 41 21

Screenshot 2025-03-20 alle 16 41 30

PR #127, opened 2025-03-11
Lossless WebP encoding support

empiricompany:
Set webp quality to 101 for lossless encoding
https://php.watch/versions/8.1/gd-webp-lossless

PR #132, opened 2025-03-16
Unified image processing quality setting

fballiano:
This PR:
- forces the reprocessing of every image uploaded, which is a security feature (which was enabled by default long time ago)
- sets a unified "quality" for every processing of images (thumbnails or uploads-reprocessing)
- closes https://github.com/MahoCommerce/maho/pull/127

@justinbeaty @empiricompany

### DIFF
--- a/lib/Varien/Image/Adapter/Gd2.php
+++ b/lib/Varien/Image/Adapter/Gd2.php
@@ -182,9 +182,8 @@ class Varien_Image_Adapter_Gd2 extends Varien_Image_Adapter_Abstract
         $functionParameters[] = $this->_imageHandler;
         $functionParameters[] = $fileName;
 
-        $quality = $this->quality() ?? Mage::getStoreConfigAsInt('system/media_storage_configuration/image_quality');
-        if (!is_null($this->quality()) && ($this->_fileType == IMAGETYPE_JPEG || $this->_fileType == IMAGETYPE_WEBP)) {
-            $functionParameters[] = $this->quality();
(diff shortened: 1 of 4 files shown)


empiricompany:
ok but in my PR I was investigating quality issues in WebP with png images

fballiano:
@empiricompany I got lost in your comment and I didn't understand what's the bug you're having

empiricompany:
sorry, PR was born to add lossless to WebP to solve a problem but it doesn't actually solve it :) 

I will open a separate case of the quality issue that occurs with PNG images converted to WebP, for which I do not currently have a solution.

fballiano:
ok, I think this PR is useful anyway to have a single image quality settings for all processing

lib/Varien/Image/Adapter/Gd2.php
```diff
@@ -182,7 +182,7 @@ public function save($destination = null, $newName = null)
         $functionParameters[] = $this->_imageHandler;
         $functionParameters[] = $fileName;
 
-        // set quality param for JPG file type
+        $quality = $this->quality() ?? Mage::getStoreConfigAsInt('system/media_storage_configuration/image_quality');
```
justinbeaty, line 185: This var doesn't seem to be used, unless you meant to replace `$this->getQuality()` in the below lines?

fballiano: 😂 you're right

fballiano:
I'll merge this in order to debug https://github.com/MahoCommerce/maho/issues/150

PR #136, opened 2025-03-17
Rewrote mage/adminhtml/sales.js without prototypejs

justinbeaty:
I believe this was the largest JS file still with prototype, and obviously a file that is very crucial to Maho. I've a lot of testing and can't find any more issues.

### Testing

This JS file is only loaded on two pages: https://example.com/admin/sales_order_create/, and https://example.com/admin/sales_order_edit/ (i.e. you press the Edit Order button). So those should be the only pages to test, and the edit page is actually handled by the same controller action, so they should be identical.

To find all places to test, I ran the following grep to find calls to the window's order object, and made sure each re-written method was called.
```sh
git grep -E 'order\.\w+' -- '*.phtml' '*.php' '*.js'
```

Things to test:
- Create admin order for existing customer, new customer, and guest
- Add products from grid
- Change prices and qty for quote items
- Move items between quote, wishlist, cart, etc.
- Apply coupons
- Change addresses, shipping same as billing, etc.
- Validate VAT
- Change payment method
- Change shipping method
- Add gift message
- Add order comment

### Improvements

1\. Change item quantity, don't press Update Items & Qty's, then click Submit Order.
Before you get error message: "You have item changes [OK] [CANCEL]". I had no idea what that meant, but if you press OK then the order is submitted without updating the quantity.

Now, the message is "You have unsaved item changes. Discard and proceed with checkout? [OK] [CANCEL]"

2\. Validate VAT
I never used it, but I still found some bugs around API error handling and customer group switching.

3\. ControlButton JS Class
If you look at the bottom of the old sales.js, you'll see this prototype class that is used in exactly one place in the entire code.

Before, the "Add Products" button disappears when you press it, and showed up again once you're done. Maybe because people were pressing that button instead of the "Add Selected Product(s) to Order" button. Leaving it visible doesn't cause any problems, but if we want to re-hide it we can do with CSS instead of creating, destroying, and recreating the button.

As a side note I also added a "Cancel" button to the add products grid so if you accidentally hit "Add Products", you can cancel without the only option being "Add Selected Product(s)".

4\. Made disabled overlay CSS only
Do you know how if you have a virtual order the shipping address and shipping method boxes are greyed out? Mage devs did that by adding a window resize event observer and absolutely positioning the div on top. I replaced it with a few lines on CSS...

### Misc

Trying to get a rough estimation on how much more prototypejs we have:

```
git grep Class.create -- ':!public/js/prototype' | wc -l
```

| This PR | Adminhtml Theme | OM |
|--------|--------|--------|
| 27 | 21 | 109 |

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Search.php
+++ b/app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Search.php
@@ -6,6 +6,7 @@
  * @package    Mage_Adminhtml
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright  Copyright (c) 2022-2024 The OpenMage Contributors (https://openmage.org)
+ * @copyright  Copyright (c) 2025 Maho (https://mahocommerce.com)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
@@ -16,10 +17,25 @@
  */
 class Mage_Adminhtml_Block_Sales_Order_Create_Search extends Mage_Adminhtml_Block_Sales_Order_Create_Abstract
 {
+    /**
+     * Contains button descriptions to be shown at the top of accordion
+     * @var list<array>
+     */
+    protected $_buttons = [];
+
     public function __construct()
     {
         parent::__construct();
         $this->setId('sales_order_create_search');
+        $this->addButton([
+            'label' => Mage::helper('sales')->__('Add Selected Product(s) to Order'),
+            'onclick' => 'order.productGridAddSelected()',
+            'class' => 'add',
+        ]);
+        $this->addButton([
+            'label' => Mage::helper('sales')->__('Cancel'),
+            'onclick' => 'order.productGridHide()',
+        ]);
     }
 
     /**
@@ -31,16 +47,30 @@ class Mage_Adminhtml_Block_Sales_Order_Create_Search extends Mage_Adminhtml_Bloc
     }
 
     /**
+     * Add button to the items header
+     *
+     * @param array $args
+     */
+    public function addButton($args)
+    {
+        $this->_buttons[] = $args;
+    }
+
+    /**
+     * Render buttons and return HTML code
+     *
      * @return string
      */
     public function getButtonsHtml()
     {
-        $addButtonData = [
-            'label' => Mage::helper('sales')->__('Add Selected Product(s) to Order'),
-            'onclick' => 'order.productGridAddSelected()',
-            'class' => 'add',
-        ];
-        return $this->getLayout()->createBlock('adminhtml/widget_button')->setData($addButtonData)->toHtml();
+        $html = '';
+        // Make buttons to be rendered in opposite order of addition. This makes "Add products" the last one.
+        $this->_buttons = array_reverse($this->_buttons);
+        foreach ($this->_buttons as $buttonData) {
+            $html .= $this->getLayout()->createBlock('adminhtml/widget_button')->setData($buttonData)->toHtml();
+        }
+
+        return $html;
     }
 
     /**
--- a/public/js/mage/adminhtml/tools.js
+++ b/public/js/mage/adminhtml/tools.js
@@ -832,3 +832,15 @@ function setQueryParams(url, params = {}) {
     }
     return url.toString();
 }
+
+/**
+ * Alternative to PrototypeJS's Function.wrap() method
+ */
+function wrapFunction(originalFn, wrapperFn) {
+    if (typeof originalFn !== 'function' || typeof wrapperFn !== 'function') {
+        throw new TypeError('Arguments must be functions');
+    }
+    return function() {
+        return wrapperFn(originalFn, ...arguments);
+    };
+}
--- a/app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Items.php
+++ b/app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Items.php
@@ -6,7 +6,7 @@
  * @package    Mage_Adminhtml
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright  Copyright (c) 2022-2024 The OpenMage Contributors (https://openmage.org)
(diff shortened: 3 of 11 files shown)


app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Search.php
```diff
@@ -16,10 +17,25 @@
  */
 class Mage_Adminhtml_Block_Sales_Order_Create_Search extends Mage_Adminhtml_Block_Sales_Order_Create_Abstract
 {
+    /**
+     * Contains button descriptions to be shown at the top of accordion
+     * @var list<array>
+     */
+    protected $_buttons = [];
```
justinbeaty, line 24: In this class I didn't add native PHP types to methods and properties and I used the underscore.

If we add them, then this block would work differently than the other `Mage_Adminhtml_Block_Sales_Order_Create_*` blocks, and I'm not sure that's great.

Any thoughts?

fballiano: mmm tricky situation but I think coherence makes more sense in this case, so I'd live it as you implement it already. If in the future we mass add types then at least all those classes will still be similar between each other.

justinbeaty: > If in the future we mass add types then at least all those classes will still be similar between each other.

That's what I was thinking too. Maybe someday... :)

fballiano: maybe someday

public/js/mage/adminhtml/tools.js
```diff
@@ -832,3 +832,15 @@ function setQueryParams(url, params = {}) {
...
+ * Alternative to PrototypeJS's Function.wrap() method
+ */
+function wrapFunction(originalFn, wrapperFn) {
+    if (typeof originalFn !== 'function' || typeof wrapperFn !== 'function') {
+        throw new TypeError('Arguments must be functions');
+    }
+    return function() {
+        return wrapperFn(originalFn, ...arguments);
+    };
+}
```
fballiano, line 846: do you think this wrapfunction is absolutely necessary?

my worry is that we replace prototype but add a series of functions (really we didn't do that, but I say that just to explain the thought) that, at the end of the day, do the same thing.

since the wrapFunction is used only 2 times maybe...

fballiano:
I've tested order modification and couldn't find any problem (adding products, modifying, modifying options, configuring configurables, moving to wishlist/cart and back, changing address and syncing it with the billing address, shipping options, gift messages)

public/js/mage/adminhtml/tools.js
justinbeaty: I definitely understand the concern, but I'm a bit worried about breaking admin checkout modules. Any other place in the code and I wouldn't hesitate to remove it.

A module would use it something like this without overriding the parent method:

```js
window.order.dataArea.onLoad = wrapFunction(window.order.dataArea.onLoad, (proceed) => {
    // Some custom logic

    // Call parent method, which calls its parent method, etc
    proceed();

    // More custom logic
});
```

Which seems more elegant than code that does stuff like this:
```js
    // stop Magento's card_validation ajax from running                                                                            
    AdminOrder.prototype._getPaymentDataOriginal = AdminOrder.prototype.getPaymentData;
    AdminOrder.prototype.getPaymentData = function(currentMethod) {
        return (currentMethod == 'gene_braintree_creditcard' || currentMethod == 'gene_braintree_paypal') ? false : this._getPaymentDataOriginal(currentMethod);
    };
```

The above is from the Gene Braintree module, but there was [similar code](https://github.com/MahoCommerce/maho/pull/68/files#diff-8e79ab312dd6720c6ed65cb6ca2fc3de5016d021faf6c304cfb452e959d7ff51) in core, so the pattern is likely common in modules.

fballiano: but they would break anyway the moment we remove prototype if they were using `$something.wrap()` right? maybe i'm missing something

```javascript                                                                   
    AdminOrder.prototype._getPaymentDataOriginal = AdminOrder.prototype.getPaymentData;
    AdminOrder.prototype.getPaymentData = function(currentMethod) {
        return (currentMethod == 'gene_braintree_creditcard' || currentMethod == 'gene_braintree_paypal') ? false : this._getPaymentDataOriginal(currentMethod);
    };
```

I kinda like this one more 😅 just personal preference, it seems easier to read, but as you wish

fballiano:
tested also order creation for the same things, couldn't find any problem

public/js/mage/adminhtml/tools.js
justinbeaty: > but they would break anyway the moment we remove prototype if they were using $something.wrap() right? maybe i'm missing something

It would break, but the fix would be simply changing `myFunc.wrap(myCallable)` with `wrapFunction(myFunc, myCallable)`.

If we are going to break it more than that, then maybe we could create some sort of events that modules could hook into?

fballiano: I see your point, although it would be difficult for people to know that this new function exist, I'm ok with it :-)

fballiano:
@justinbeaty I'm ok merging already, you tell me :-)

public/js/mage/adminhtml/tools.js
justinbeaty: Also, just a thought, but `Function.wrap` could be mapped to `wrapFunction` when we finally tackle #84, which would be nice that nothing would break.

justinbeaty:
> @justinbeaty I'm ok merging already, you tell me :-)

I'm not against improving it in the future, but I think it's ready to merge now if you couldn't find any bugs. :smiley:

public/js/mage/adminhtml/tools.js
justinbeaty: It's still on the back of my mind, I've just been trying to knock out some more of those `Class.create` instances beforehand.

fballiano:
> > @justinbeaty I'm ok merging already, you tell me :-)
> 
> I'm not against improving it in the future, but I think it's ready to merge now if you couldn't find any bugs. 😃

this is already a wonderful effort and we've so many things going on that's totally ok as it is :-) :-)

PR #139, opened 2025-03-19
Removed obsolete and unused Mage_HTTP_* classes

fballiano:
These classes are not used anywhere in the core and in 2025 there are many modern HTTP clients to use instead of the old internal one.

spring cleaning mode 😅


justinbeaty (approved):
Can confirm I find no references to these classes:

```sh
$ git grep --name-only Mage_HTTP
.phpstan.dist.baseline.neon
lib/Mage/HTTP/Client.php
lib/Mage/HTTP/Client/Curl.php
lib/Mage/HTTP/Client/Socket.php
lib/Mage/HTTP/IClient.php
```

PR #140, opened 2025-03-20
[WIP] Migrated to libsodium

fballiano:
@justinbeaty as we were discussing, this first test works although `decrypt()` it's full of `return ''` since it's just a first test.

- should we throw exception?
- should we return '' as now but log the failed decrypt?

Note: phpstan highlights that phpseclib is used in `Varien_Io_Sftp` which only seems to be used in `Mage_Paypal_Model_Report_Settlement::fetchAndSave()` :-\


justinbeaty:
I had no idea it used SFTP to get reports, but I’ve never used those default PayPal modules. Could this be an alternative? https://developer.paypal.com/api/nvp-soap/payflow/reports/use-the-api/

The empty string question, not sure I’ll have to take a look tomorrow when I’m back at my desk.

fballiano:
I'm not super sure because I'm not a paypal expert but reading here and there it seems report-api is not exactly like sftp settlement download :-(

So maybe we could start by just removing mcrypt_compat and leave phpseclib and the sftp thingy as it is now :-(

fballiano:
At the moment my opinion is that we should fail gracefully (`return ''`) and log the failed decryption.

The new compat module (which doesn't exist yet) will take care of the rest.

Still thinking about it

justinbeaty:
I altered the encrypt / decrypt functions a bit:

1\. I used the constant time `sodium_` functions for base64: https://www.php.net/manual/en/function.sodium-bin2base64.php

2\. The key should actually be a binary string, so switched to the non `mb_` variants.

3\. We actually have a problem, our crypt key is only 16 bytes, but we need 32. I temporarily duplicated the `$key` variable to get us there.

4\. Used try catch in decrypt since `sodium_base642bin` can throw`.

5\. Zero out some extra memory in encrypt, not sure if there's a reason not to.

```php
    public function encrypt(#[\SensitiveParameter] string $data): string
    {
        $key = sodium_hex2bin((string) Mage::getConfig()->getNode('global/crypt/key'));
        $nonce = random_bytes(SODIUM_CRYPTO_SECRETBOX_NONCEBYTES);
        $ciphertext = sodium_crypto_secretbox($data, $nonce, $key.$key);
        $encrypted = sodium_bin2base64($nonce . $ciphertext, SODIUM_BASE64_VARIANT_ORIGINAL);

        // Clean sensitive data from memory
        sodium_memzero($data);
        sodium_memzero($nonce);
        sodium_memzero($ciphertext);

        return $encrypted;
    }

    public function decrypt(#[\SensitiveParameter] string $data): string
    {
        try {
            $decoded = sodium_base642bin($data, SODIUM_BASE64_VARIANT_ORIGINAL);

            if (strlen($decoded) < (SODIUM_CRYPTO_SECRETBOX_NONCEBYTES + SODIUM_CRYPTO_SECRETBOX_MACBYTES)) {
                throw new Exception('Data is too short to be valid');
            }

            $key = sodium_hex2bin((string) Mage::getConfig()->getNode('global/crypt/key'));
            $nonce = substr($decoded, 0, SODIUM_CRYPTO_SECRETBOX_NONCEBYTES);
            $ciphertext = substr($decoded, SODIUM_CRYPTO_SECRETBOX_NONCEBYTES);
            $plaintext = sodium_crypto_secretbox_open($ciphertext, $nonce, $key.$key);

            if ($plaintext === false) {
                throw new Exception('Decryption failed: data may be corrupted or tampered with');
            }

            // Clean sensitive data from memory
            sodium_memzero($decoded);
            sodium_memzero($nonce);
            sodium_memzero($ciphertext);

            return $plaintext;

        } catch (Throwable $e) {
            Mage::logException($e);
            return '';
        }
    }
```

fballiano:
my encryption key is 32 char 🧐🧐 how is your only 16?

justinbeaty:
> my encryption key is 32 char 🧐🧐 how is your only 16?

Mine is 32 hex digits, but that only equals 16 bytes. :/

fballiano:
mmmm right, but when I tested (i'm now working on another interesting idea) I didn't get any error about the length 🤔

justinbeaty:
> mmmm right, but when I tested (i'm now working on another interesting idea) I didn't get any error about the length 🤔

It's because any 32 character hex string is also a valid 32 character binary string.

Example:

```php
<?php

$bytes = random_bytes(32);
$hex = sodium_bin2hex($bytes);

var_dump($bytes); // string(32) "/ExVsE`H<pJ%9:zT^X"
var_dump($hex); // string(64) "9b2ffd457856877318ff0245bb6048893c704a2539aff3113a7aaa545eb758cf"
```

justinbeaty:
> should we return ''

How about returning false on error? I say because `''` is a valid string one could encode, and otherwise we don't know if we had an empty string or an error. Unless that messes up a bunch of function signatures...?

fballiano:
maybe we could create a MahoCLI command to generate a new (longer) key?

fballiano:
> > should we return ''
> 
> How about returning false on error? I say because `''` is a valid string one could encrypt, and otherwise we don't know if we had an empty string or an error. Unless that messes up a bunch of function signatures...?

the thing is, if we allow that function to throw an error, if a decryption fails the website breaks (while developing I had errors accessing the configuration page). if we fail silently (returning empty string) the system doesn't know that the decryption failed, but the website continues to work... of course the functionality that needs the decrypted data won't work, but... there will be more error control in the functionality itself for sure.

it's not ideal for sure, but I don't know :-\

fballiano:
@justinbeaty I've done a lot of things today (updated the main description of this PR). question, with the new composer plugin, in https://github.com/MahoCommerce/module-mcrypt-compat do I need to create the modman file?

justinbeaty:
> @justinbeaty I've done a lot of things today (updated the main description of this PR). 

Great, I'll review tomorrow. The key rotation command sounds amazing.

> question, with the new composer plugin, in https://github.com/MahoCommerce/module-mcrypt-compat do I need to create the modman file?

It's not needed. Unlike the OM plugin, we don't need a modman file unless the files are in a non-standard location, i.e. `src/Model/Encryption.php` instead of `app/code/core/Mage/Core/Model/Encryption.php`.

fballiano:
> The key rotation command sounds amazing.

and I just added support for converting blowfish encrypted data too :-) (teste! it works!!!)

> It's not needed. Unlike the OM plugin, we don't need a modman file unless the files are in a non-standard location, i.e. src/Model/Encryption.php instead of app/code/core/Mage/Core/Model/Encryption.php.

great then I think it's all set, although with the key regeneration command and the convertion between mcrypt and sodium crypted data... the compat module could maybe be avoided...

app/code/core/Mage/Core/Model/Encryption.php
```diff
@@ -153,53 +148,52 @@ public function validateHashByVersion(#[\SensitiveParameter] $password, $hash, $
...
-     *
-     * @param string $data
-     * @return string
-     */
-    public function encrypt($data)
+    public function decrypt(#[\SensitiveParameter] string $data): string
     {
-        return base64_encode($this->_getCrypt()->encrypt((string) $data));
-    }
+        $decoded = base64_decode($data);
```
justinbeaty, line 163:
```suggestion
        $decoded = sodium_base642bin($data, SODIUM_BASE64_VARIANT_ORIGINAL);
```

app/code/core/Mage/Core/Model/Encryption.php
```diff
@@ -153,53 +148,52 @@ public function validateHashByVersion(#[\SensitiveParameter] $password, $hash, $
...
-    /**
-     * Decrypt a string
-     *
-     * @param string $data
-     * @return string
-     */
-    public function decrypt($data)
-    {
-        return str_replace("\x0", '', trim($this->_getCrypt()->decrypt(base64_decode((string) $data))));
+        if (mb_strlen($decoded, '8bit') < (SODIUM_CRYPTO_SECRETBOX_NONCEBYTES + SODIUM_CRYPTO_SECRETBOX_MACBYTES)) {
```
justinbeaty, line 170: I don't think we want the `mb_` variants here since we're working with binary strings. Potentially `mb_strlen` could report that the string is less characters than the number of bytes.

app/code/core/Mage/Core/Model/Encryption.php
```diff
@@ -153,53 +148,52 @@ public function validateHashByVersion(#[\SensitiveParameter] $password, $hash, $
...
-            if ($key === null) {
-                $key = (string) Mage::getConfig()->getNode('global/crypt/key');
-            }
-            $this->_crypt = Varien_Crypt::factory()->init($key);
-        }
-        return $this->_crypt;
+        $key = Mage::getEncryptionKeyAsBinary();
+        $nonce = random_bytes(SODIUM_CRYPTO_SECRETBOX_NONCEBYTES);
+        $ciphertext = sodium_crypto_secretbox($data, $nonce, $key);
+        $encrypted = base64_encode($nonce . $ciphertext);
```
justinbeaty, line 156:
```suggestion
        $encrypted = sodium_bin2base64($nonce . $ciphertext, SODIUM_BASE64_VARIANT_ORIGINAL);
```

justinbeaty, line 156:
```suggestion
        $encrypted = sodium_bin2base64($nonce . $ciphertext, SODIUM_BASE64_VARIANT_ORIGINAL);
```

Sorry deleted first suggestion here because not enough coffee yet. 😴 

From the PHP docs:

> Unlike base64_encode(), sodium_bin2base64() is constant-time (a property that is important for any code that touches cryptographic inputs, such as plaintexts or keys)

app/code/core/Mage/Install/Block/End.php
```diff
@@ -26,7 +27,7 @@ public function getEncryptionKey()
     {
```
justinbeaty, line 27: Should we just delete this method? It's usage in `install/end.phtml` is removed, and I can't imagine anyone has custom install themes. :)

app/code/core/Mage/Core/Model/Encryption.php
fballiano: right! the 8bit param should fix that but makes no sense!

app/code/core/Mage/Install/Block/End.php
fballiano: absolutely!

app/code/core/Mage/Core/Model/Encryption.php
justinbeaty: Ah, you're right that it would correctly count bytes. [This comment](https://www.php.net/manual/en/function.mb-strlen.php#77040) gives a reason one would use mb_strlen with 8bit, but it turns out `mbstring.func_overload` is removed in PHP8, and strlen is much faster.

app/code/core/Mage/Core/Model/Encryption.php
fballiano: added try/catch ;-)

app/code/core/Mage/Core/Model/Encryption.php
fballiano: phpstorm says they throw exception but I can't find anything in the doc and i can't make those functions throw exception... I don't think they throw (anymore?)

fballiano: ok for base64decode I managed to throw an exception so that part of code is ok, about the encode I wouldn't know what to do with an exception :-D

justinbeaty: Yeah, I'm not sure how you'd get `sodium_bin2base64` to throw either, and I wouldn't worry about catching it because as you said I'm not sure there's anything to do either.

fballiano: super, I've to check the last point of the checklist (which we could decide to use to alert backend to rotate the encryption key every once in a while?) and it should be done :-D

fballiano:
I'm done with all my code and tests :-)

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
```diff
@@ -0,0 +1,226 @@
...
+        }
+        $output->writeln('<info>Created backup at: ' . $backupPath . '</info>');
+
+        // Read the current configuration file and replace the encryption key
+        $localXmlContent = file_get_contents($localXmlPath);
+        $updatedContent = preg_replace(
+            '/<key(?:\s+date="[^"]*")?><!\[CDATA\[(.*?)\]\]><\/key>/',
+            '<key date="' . $currentDate . '"><![CDATA[' . $newKey . ']]></key>',
+            $localXmlContent,
+        );
```
justinbeaty, line 99:
```suggestion
        $localXmlContent = simplexml_load_file($localXmlPath);
        $localXmlContent->global->crypt->key = $newKey;
        $localXmlContent->global->crypt->key->addAttribute('date', '');
        $localXmlContent->global->crypt->key->attributes()->date = $currentDate;

        $updatedContent = $localXmlContent->asXML();
```

I haven't fully tested this yet, but I'm always reminded of [this stackoverflow answer](https://stackoverflow.com/a/1732454) when parsing XML with regex.

I have a vague recollection that I've removed the `<![CDATA[` tag in my local.xml before because it's actually pointless here. The key will never contain characters that could be mistaken for XML.

fballiano: yep but it's what's already used in the install module for the rest of the things, that's why I did it that way, dunno if I would rewrite all of that in the context of this PR

about the CDATA, what if the password contains a "<" char, which I think it could be pretty normal?

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
```diff
@@ -0,0 +1,226 @@
...
+            $handler = mcrypt_module_open(MCRYPT_BLOWFISH, '', MCRYPT_MODE_ECB, ''); // @phpstan-ignore constant.notFound,constant.notFound
+            $initVector = mcrypt_create_iv(mcrypt_enc_get_iv_size($handler), MCRYPT_RAND); // @phpstan-ignore function.notFound,function.notFound,constant.notFound
+            mcrypt_generic_init($handler, $key, $initVector); // @phpstan-ignore function.notFound
+            $data = mdecrypt_generic($handler, (string) base64_decode((string) $data)); // @phpstan-ignore function.notFound
+            mcrypt_generic_deinit($handler); // @phpstan-ignore function.notFound
+            mcrypt_module_close($handler); // @phpstan-ignore function.notFound
+            return str_replace("\x0", '', trim($data));
+        }
+
+        $decoded = (string) base64_decode($data);
```
justinbeaty, line 206: I think we should use `sodium_base642bin` here too (and inverse in this->encrypt()), but we would need to handle exceptions.

fballiano: right, pushed it now, for the exception I think I still return '', interrupting the flow there will end up in half-converted data and unrecoverable data so... I think it's probably better to continue

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
justinbeaty: The CDATA would need to stay for the database password for sure. I just am not sure why Mage devs felt the need to add it around every other value.

But, If that's how the install module is now, then no reason to change anything in this PR, let's just leave well enough alone. :)

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
justinbeaty: The only other option I could think of is to try to decrypt / encrypt every value before writing anything to the database or updating local.xml. But that would require quite some refactoring.

app/code/core/Mage/Core/Model/Encryption.php
```diff
@@ -153,53 +148,53 @@ public function validateHashByVersion(#[\SensitiveParameter] $password, $hash, $
...
-                $key = (string) Mage::getConfig()->getNode('global/crypt/key');
-            }
-            $this->_crypt = Varien_Crypt::factory()->init($key);
-        }
-        return $this->_crypt;
+        $key = Mage::getEncryptionKeyAsBinary();
+        $nonce = random_bytes(SODIUM_CRYPTO_SECRETBOX_NONCEBYTES);
+        $ciphertext = sodium_crypto_secretbox($data, $nonce, $key);
+        $encrypted = sodium_bin2base64($nonce . $ciphertext, SODIUM_BASE64_VARIANT_ORIGINAL);
+        sodium_memzero($data); // Clean sensitive data from memory
```
justinbeaty, line 157:
```suggestion
        // Clean sensitive data from memory
        sodium_memzero($data);
        sodium_memzero($key);
        sodium_memzero($nonce);
        sodium_memzero($ciphertext);
```

Should we zero out these values too? The example from the link below seems to do so:

https://davegebler.com/post/php/php-encryption-the-right-way-with-libsodium

app/code/core/Mage/Core/Model/Encryption.php
```diff
@@ -153,53 +148,53 @@ public function validateHashByVersion(#[\SensitiveParameter] $password, $hash, $
...
+        if ($plaintext === false) {
+            $exception = new Exception('Decryption failed: data may be corrupted or tampered with');
+            Mage::logException($exception);
+            return '';
+        }
+
+        // Clean sensitive data from memory
+        sodium_memzero($decoded);
+        sodium_memzero($nonce);
+        sodium_memzero($ciphertext);
```
justinbeaty, line 191:
```suggestion
        sodium_memzero($data);
        sodium_memzero($decoded);
        sodium_memzero($key);
        sodium_memzero($nonce);
        sodium_memzero($ciphertext);
```

A few more values here too.

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
```diff
@@ -0,0 +1,233 @@
...
+            $data = mdecrypt_generic($handler, (string) base64_decode((string) $data)); // @phpstan-ignore function.notFound
+            mcrypt_generic_deinit($handler); // @phpstan-ignore function.notFound
+            mcrypt_module_close($handler); // @phpstan-ignore function.notFound
+            return str_replace("\x0", '', trim($data));
+        }
+
+        try {
+            $decoded = sodium_base642bin($data, SODIUM_BASE64_VARIANT_ORIGINAL);
+        } catch (SodiumException $e) {
+            $exception = new Exception('Invalid base64 encoding: ' . $e->getMessage());
```
justinbeaty, line 209:
```suggestion
        } catch (\SodiumException $e) {
            $exception = new \Exception('Invalid base64 encoding: ' . $e->getMessage());
```

This should fix the PHPStan error.

justinbeaty:
For the compat module, two things:

1) The composer.json file has the description as "Image cleaner module for OpenMage (and Magento 1.9)"

2) Do you need to add it to packagist? I had to clone locally to install.

Sorry for so many suggestions -- because it deals with encryption I'm trying to be thorough. :)

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
fballiano: it would be better, we could think about a transaction...

justinbeaty:
> but it works also without a proper "magento module" structure right?

I'm not sure exactly what you're asking here, but the only requirement for composer.json is that type is `(maho|magento)-module`. 

There's also support for [`extra.map`](https://github.com/Cotya/magento-composer-installer/blob/main/doc/Mapping.md), but to be honest I don't think I've ever seen a module use that. But it was only 4 lines of code in the maho composer plugin which is why I added support for it.

In any case, the module installed perfectly via packagist now!

fballiano:
> I'm not sure exactly what you're asking here, but the only requirement for composer.json is that type is `(maho|magento)-module`.

😅 I meant ilke to create a real module with its own config.xml app/etc/modules/xxx.xml entry etc etc

justinbeaty:
> > I'm not sure exactly what you're asking here, but the only requirement for composer.json is that type is `(maho|magento)-module`.
> 
> 😅 I meant ilke to create a real module with its own config.xml app/etc/modules/xxx.xml entry etc etc

Yeah I think it's fine as long as any other changes to `Mage_Core_Model_Encryption` are ported back to that module.  It definitely works because I haven't run the key rotation yet and without the module I get errors.

fballiano:
nice! well that's meant to be a temporary replacement but people should use it only for the minimum amount of time possible :-)

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
justinbeaty: I'm still wondering if returning false might be better than `''` or throwing. Unsure though since that might have rippling changes.

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
```diff
@@ -0,0 +1,244 @@
...
+
+        // Clean sensitive data from memory
+        sodium_memzero($data);
+        sodium_memzero($key);
+        sodium_memzero($nonce);
+        sodium_memzero($ciphertext);
+
+        return $encrypted;
+    }
+}
```
justinbeaty, line 244:
```suggestion
    private function createBackupPath(): string
    {
        $i = 0;
        $currentDate = date('Y-m-d');

        do {
            $backupPath = 'app/etc/local.xml.bak.' . $currentDate;
            if ($i > 0) {
                $backupPath .= "-$i";
            }
            $i++;
        } while (file_exists($backupPath));

        return $backupPath;
    }
}
```

Do you think we should have a function like this? What happened is that I ran `./maho sys:encryptionkey:regenerate` twice (intentionally) but that meant that I lost my original encryption key.

fballiano: mmm instead of that I'll add HHMMSS but I thought it would never happen in production because there's 1 or 2 confirmations asked before actually doing anything

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
fballiano: I don't know, the return '' but logging the exception seems a good compromise to me. if the crypted data is corrupted nothing will work anyway but breaking the flow worries me more

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
fballiano: what do you think about the latest version?

justinbeaty: Yeah, that works too and is much simpler. I think better safe than sorry, because you never know what a user might do.

fballiano: :-) ;-)

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
justinbeaty: Yeah, in the end I think it's fine how it is. Furthermore, I tested with the backend config and it seems you can't encrypt an empty string there. If you leave a field blank it's NULL in core_config_data.

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
```diff
@@ -0,0 +1,244 @@
...
+            'encrypt_callback' => [$this, 'encrypt'],
+            'decrypt_callback' => [$this, 'decrypt'],
+        ]);
+
+        if (\Composer\InstalledVersions::isInstalled('phpseclib/mcrypt_compat')) {
+            $output->writeln('');
+            $output->writeln('<error>Warning: phpseclib/mcrypt_compat is installed. This package can cause encryption issues and should be removed.</error>');
+            $output->writeln('<error>Please remove it using: composer remove phpseclib/mcrypt_compat</error>');
+            $output->writeln('');
+        }
```
justinbeaty, line 188:
```suggestion
        if (\Composer\InstalledVersions::isInstalled('mahocommerce/module-mcrypt-compat')) {
            $output->writeln('');
            $output->writeln('<error>You may now remove the compatibility module using: composer remove mahocommerce/module-mcrypt-compat</error>');
            $output->writeln('');
        } elseif (\Composer\InstalledVersions::isInstalled('phpseclib/mcrypt_compat')) {
            $output->writeln('');
            $output->writeln('<error>Warning: phpseclib/mcrypt_compat is installed. This package can cause encryption issues and should be removed.</error>');
            $output->writeln('<error>Please remove it using: composer remove phpseclib/mcrypt_compat</error>');
            $output->writeln('');
        }
```

I had the compatibility module installed, but the message about removing `phpseclib/mcrypt_compat` could be a bit confusing since I didn't directly install that, and thus `composer remove phpseclib/mcrypt_compat` has no effect.

fballiano: ok i'll write a better message there

fballiano: ahhh sorry I read your message but didn't notice the suggestion, what do you think about https://github.com/MahoCommerce/maho/pull/140/commits/c637eba69cee0df4965eedf1810ea52ab0d57a4f?

justinbeaty:
There are a few other places we might consider migrating data:

1\. Table `admin_user` the `twofa_secret` column. Otherwise I was locked out of the backend and needed to disable 2fa directly in the database.

2\. Table `sales_flat_order_payment` the `cc_number_enc` column.

3\. Table `sales_flat_quote` the `password_hash` column. The checkout flow temporarily stores the encrypted password in the quote table so it can create the customer upon order completion. I'm not sure why it does that instead of storing a password hash.

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
justinbeaty: The updated message is good, but I think we should still check for `mahocommerce/module-mcrypt-compat` because once you've generated a new key, the compat module **must** be removed otherwise you'll get this error:

PHP Fatal error:  Uncaught Varien_Exception: Maximum key size must be smaller 56 in /home/www-data/work/maho/git/lib/Varien/Crypt/Mcrypt.php:72

fballiano: totally right, what about this double message? https://github.com/MahoCommerce/maho/pull/140/commits/8791212ae3127514a1fdd6edcfe1efd0d397a8bd

fballiano:
> 2. Table `sales_flat_order_payment` the `cc_number_enc` column.

this seems to come from the CC payment method which I wanted to remove from OM and that shouldn't be used anymore, maybe we can let this die with it?

I'll work on the other two :-)

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
justinbeaty: Yeah that's perfect. :)

justinbeaty:
> this seems to come from the CC payment method which I wanted to remove from OM and that shouldn't be used anymore, maybe we can let this die with it?

I've never used the native CC method either, but a few thoughts:

1\. I am seeing a few other methods extending it. Would removing it break any of these?

```sh
$ git grep -h 'extends Mage_Payment_Model_Method_Cc'
class Mage_Paygate_Model_Authorizenet extends Mage_Payment_Model_Method_Cc
class Mage_Payment_Model_Method_Ccsave extends Mage_Payment_Model_Method_Cc
class Mage_Paypal_Model_Direct extends Mage_Payment_Model_Method_Cc
class Mage_Paypal_Model_Payflowpro extends Mage_Payment_Model_Method_Cc
```

2\. If we remove the module and prevent new values from being saved, should we still be able to decrypt old values?

https://github.com/MahoCommerce/maho/blob/3c9d2e8e56c5c3af95a5ee6c8c5b9ccbac1b56a7/app/code/core/Mage/Payment/Model/Info.php#L61-L74

3\. Might some other payment modules store info there? If this is the case and we eventually auto-rotate keys, then stored CC info would be wiped every N days.

fballiano:
mmm actually the method I was removing from OM was only ccsave, which I still think should be removed but yeah I guess we can't remove all the others :-\

it seems that encrypting only the credit card number is not a good idea, but also I would never have a CCnumber in my database, encrypted or not... dunno what to do here

easy thing is to re-encrypt those data too but... do we want to have payment modules that store CC data??

justinbeaty:
It looks like the Ccsave method is the only *core* method that stores the cc info: (it also encrypts the 3 digit code).

```
$ git grep _canSaveCc app
app/code/core/Mage/Paygate/Model/Authorizenet.php:    protected $_canSaveCc               = false;
app/code/core/Mage/Payment/Model/Method/Cc.php:    protected $_canSaveCc     = false;
app/code/core/Mage/Payment/Model/Method/Cc.php:        if ($this->_canSaveCc) {
app/code/core/Mage/Payment/Model/Method/Ccsave.php:    protected $_canSaveCc   = true;
app/code/core/Mage/Paypal/Model/Direct.php:    protected $_canSaveCc = false;
app/code/core/Mage/Paypal/Model/Hostedpro.php:    protected $_canSaveCc               = false;
app/code/core/Mage/Paypal/Model/Payflowpro.php:    protected $_canSaveCc = false;
```

I don't store CC info in the DB either, but I don't really know the requirements of every shop. Presumably larger companies store this information, else those levels of PCI compliance wouldn't exist. Alternatively, it's possible some modules only store a masked number in that field, i.e. the first 6 digits (Bank Identification Number) and the last 4.

I think it is better to err on the side of caution, unless we ever run: `ALTER TABLE sales_flat_order_payment DROP COLUMN cc_number_enc`.

I am actually seeing 4 total encrypted columns. There's  `cc_number_enc` and `cc_cid_enc` on both `sales_flat_order_payment` and  `sales_flat_quote_payment`.

fballiano:
@justinbeaty I've creates recryptors for all those data, at the moment let's keep things as they are and we'll think about removal of stuff in another PR :-)

the only thing is that I don't have cc_cid_enc on sales_flat_order_payment and I couldn't find any setup script where it should have been created

justinbeaty:
> @justinbeaty I've creates recryptors for all those data, at the moment let's keep things as they are and we'll think about removal of stuff in another PR :-)

I had one other thought overnight. What about re-encrypting the order and quote tables using an observer in the Mage_Payment module? That would decouple this specific re-encryption logic from maho-cli. Thinking ahead, it would be nice to remove some of these payment methods from core, and it would give an example to module developers on how to use the re-encryption hook.

> the only thing is that I don't have cc_cid_enc on sales_flat_order_payment and I couldn't find any setup script where it should have been created

Yup, you're totally right, it doesn't exist. :)

fballiano:
> I had one other thought overnight. What about re-encrypting the order and quote tables using an observer in the Mage_Payment module? That would decouple this specific re-encryption logic from maho-cli. Thinking ahead, it would be nice to remove some of these payment methods from core, and it would give an example to module developers on how to use the re-encryption hook.

ok I guess it's fine but it should be 1 observer in Mage_Checkout (the password hash) and one in Mage_Payment (for cc_cid_enc and cc_number_enc), ok?

justinbeaty:
> ok I guess it's fine but it should be 1 observer in Mage_Checkout (the password hash) and one in Mage_Payment (for cc_cid_enc and cc_number_enc), ok?

Yeah, that's what I was thinking.

fballiano:
will do it right now :-) then I'd like to merge this one, the image library one, then I've another PR which I'll publish in a couple of days max then I'd like to move to the big one that you already know about but that's still a secret ehehhehe

justinbeaty:
> will do it right now :-) then I'd like to merge this one, the image library one, then I've another PR which I'll publish in a couple of days max then I'd like to move to the big one that you already know about but that's still a secret ehehhehe

Great, I'll go back and review the image library one now.

One last change request on this PR. I think we should change .gitignore to also ignore the bak files since they should never be committed.

```
# Maho stuff
/app/etc/local.xml*
!/app/etc/local.xml.template
```

fballiano:
thanks!! in the meanwhile I think I did everything and the new changes see to work fine :-)

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
```diff
@@ -0,0 +1,279 @@
...
+            $decoded = sodium_base642bin($data, SODIUM_BASE64_VARIANT_ORIGINAL);
+        } catch (\SodiumException $e) {
+            $exception = new \Exception('Invalid base64 encoding: ' . $e->getMessage());
+            Mage::logException($exception);
+            return '';
+        }
+
+        $key = sodium_hex2bin($this->oldEncryptionKey);
+        $nonce = mb_substr($decoded, 0, SODIUM_CRYPTO_SECRETBOX_NONCEBYTES, '8bit');
+        $ciphertext = mb_substr($decoded, SODIUM_CRYPTO_SECRETBOX_NONCEBYTES, null, '8bit');
```
justinbeaty, line 251:
```suggestion
        $nonce = substr($decoded, 0, SODIUM_CRYPTO_SECRETBOX_NONCEBYTES);
        $ciphertext = substr($decoded, SODIUM_CRYPTO_SECRETBOX_NONCEBYTES);
```

justinbeaty: Unless this was intentional to use the mb variants again.

app/code/core/Mage/Core/Model/Encryption.php
```diff
@@ -153,53 +148,61 @@ public function validateHashByVersion(#[\SensitiveParameter] $password, $hash, $
...
 
-    /**
-     * Encrypt a string
-     *
-     * @param string $data
-     * @return string
-     */
-    public function encrypt($data)
+    public function decrypt(#[\SensitiveParameter] string $data): string
     {
```
justinbeaty, line 168:
```suggestion
    public function decrypt(#[\SensitiveParameter] ?string $data): string
    {
        if (empty($data)) {
            // TODO: also add exception
            return '';
        }
```

I restored by database from a backup from before I set up 2fa, then went to the My Account page and got this error:

`Fatal error:  Uncaught TypeError: Mage_Core_Model_Encryption::decrypt(): Argument #1 ($data) must be of type string, null given, called in /home/www-data/work/maho/git/app/code/core/Mage/Admin/Model/User.php on line 979 and defined in /home/www-data/work/maho/git/app/code/core/Mage/Core/Model/Encryption.php:167`

So before we must have allowed to decrypt a null value.

justinbeaty:
Found hopefully one last bug; we need to flush the configuration cache after regenerating the key. 

```sh
$ ./maho sys:encryptionkey:regenerate
WARNING: This command will replace your encryption key in local.xml
All encrypted data will need to be re-encrypted with the new key.
A backup of local.xml will be created, but please ensure you have a full backup of your database before proceeding.
If possible, encrypted configuration values in core_config_data will be automatically re-encrypted with the new key.

Are you sure you want to continue? (y/N) y
Created backup at: app/etc/local.xml.bak.2025-03-31-16-46-53
Encryption key has been successfully updated
New key: 24c4671c637d55fab23d36a29dd748ba6972d09afa00d3496e394c320b54780b

Re-encrypting data on admin_user table... OK
No encrypted configurations to re-encrypt.
Re-encrypting data on sales_flat_quote_payment table... OK
Re-encrypting data on sales_flat_order_payment table... OK
Re-encrypting data on sales_flat_quote table... OK

$ php .idea/test.php
ec7a2ab77f694226036bd25e998b42534a9a948e62c4dbf4d1a3729a65ec05b6

$ ./maho cache:flush
Caches flushed successfully!

$ php .idea/test.php
24c4671c637d55fab23d36a29dd748ba6972d09afa00d3496e394c320b54780b
```

```php 
<?php # .idea/test.php

Mage::init();

echo Mage::getEncryptionKeyAsHex() . "\n";
```

The weird thing is that I was under the impression that `local.xml` is always loaded from the filesystem on each request and never cached. In actuality, it's *both* loaded from the filesystem, then overwritten from the cache. This happens in OM too.

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
```diff
@@ -0,0 +1,279 @@
...
+        $encryptedData = $readConnection->fetchAll($select);
+
+        foreach ($encryptedData as $encryptedDataRow) {
+            $writeConnection->update(
+                $table,
+                ['twofa_secret' => $this->encrypt($this->decrypt($encryptedDataRow['twofa_secret']))],
+                ['user_id = ?' => $encryptedDataRow['user_id']],
+            );
+        }
+
```
justinbeaty, line 162:
```suggestion

        Mage::app()->getCacheInstance()->clean('config');

```

justinbeaty:
Besides the last two suggestions, I've tested this pretty thoroughly and can't find any more bugs. I even ran the following command and regenerated the key 200+ times, and the data has not been corrupted.

```
while true; do yes | ./maho sys:encryptionkey:regenerate; sleep 1; done
```

fballiano:
> The weird thing is that I was under the impression that `local.xml` is always loaded from the filesystem on each request and never cached. In actuality, it's _both_ loaded from the filesystem, then overwritten from the cache. This happens in OM too.

me too!!! I always new local.xml was re-read at every request! maybe something changed in OM sometime in the past, who know

but when I added the cache-cleaning it was exactly because I couldn't make it work..

lib/MahoCLI/Commands/SysEncryptionKeyRegenerate.php
fballiano: I eneded up doing it in a slightly different way

Screenshot 2025-03-31 alle 18 43 51

one before the event (so that custom code reads everything as updated as possible) and then after the custom observers

justinbeaty:
> me too!!! I always new local.xml was re-read at every request! maybe something changed in OM sometime in the past, who know

I actually wonder if local.xml can ever be cached, since we define the cache module to use *in local.xml*... Anyway that is for another time...

app/code/core/Mage/Core/Model/Encryption.php
fballiano: I did it this way if that's ok with you

Screenshot 2025-03-31 alle 18 48 04

cause '0' also matches empty so I think a strict comparison is better in this case.

also, I don't know, I kinda wouldn't log an exception in this case, we can be tolerant and allow people to decrypt empty strings with no errors IMO

fballiano:
if this final version is fine for you, let's meeeeergeeee

app/code/core/Mage/Core/Model/Encryption.php
justinbeaty: Yup, totally good.

justinbeaty (approved):
Let's go!!!!

fballiano:
booooommmm ❤️ thanks man for all the incredible insights

justinbeaty:
Oh, it is because I don't have an smtp user / password: 
```
    public function getMailerDsn(): string
    {
        $coreHelper = Mage::helper('core');
        $emailTransport = Mage::getStoreConfig('system/smtp/enabled');
        $user = $coreHelper->decrypt(Mage::getStoreConfig('system/smtp/username'));
        $pass = $coreHelper->decrypt(Mage::getStoreConfig('system/smtp/password'));
```

I think we missed it because if you've ever saved that page in the config, you'll have an empty string. But I probably never have so I have null.

PR #143, opened 2025-03-22
Added MahoCLI command to generate PHPStorm metadata

fballiano:
with some AI and a couple more ideas (scanning for all PHP files, which should work for child projects too, and extracting the class names using PHP's tokenizer) I think this version should be kinda good and modern too.

I didn't even open n98 source code so i've no idea how they do that, but this version doesn't use any regex (as AI was proposing at the beginning), we load the configuration via Mage::getConfig etc so it should be kinda bulletproof.

with this we should be able to create our workflow to keep this metadata updated at every PR, could be nice


lib/MahoCLI/Commands/PhpstormMetadataGenerate.php
```diff
@@ -0,0 +1,408 @@
...
+        $this->loadModuleConfigurations($output);
+
+        $this->generateModelsMetadata($output, $metaDir);
+        $this->generateHelpersMetadata($output, $metaDir);
+        $this->generateRegistryMetadata($output, $metaDir);
+        $this->generateBlocksMetadata($output, $metaDir);
+        $this->generateResourceModelsMetadata($output, $metaDir);
+
+        $output->writeln('<info>PhpStorm metadata generation complete!</info>');
+        $output->writeln('<comment>Remember to add .phpstorm.meta.php to your .gitignore file</comment>');
```
justinbeaty, line 82: Do we want this line? We don't currently have `.phpstorm.meta.php` in `.gitignore`.

justinbeaty (approved):
This looks perfect. I don't have a license for PHPStorm right now, but I assume it all works.

lib/MahoCLI/Commands/PhpstormMetadataGenerate.php
fballiano: absolutely right, removed

fballiano:
Thanks for checking it out @justinbeaty, I'll try to create a workflow ASAP too

PR #144, opened 2025-03-22
Added MahoCLI command to connect to database

fballiano:
created with AI, then fixing the way it was reading the DB configuration, since it was reading it local.xml with simplexml

### DIFF
--- a/lib/MahoCLI/Commands/DBConnect.php
+++ b/lib/MahoCLI/Commands/DBConnect.php
@@ -33,16 +33,12 @@ class DBConnect extends BaseMahoCommand
         $user = (string) Mage::getConfig()->getNode('global/resources/default_setup/connection/username');
         $password = (string) Mage::getConfig()->getNode('global/resources/default_setup/connection/password');
 
-        // Create a temporary MySQL config file
         $configFile = $this->createTempMySQLConfig($host, $user, $password);
-
-        // Build MySQL command using the config file
         $mysqlCommand = sprintf(
             'mysql --defaults-extra-file=%s %s',
             escapeshellarg($configFile),
             escapeshellarg($dbname),
         );
-
         $descriptorspec = [
             0 => STDIN,
             1 => STDOUT,
@@ -50,7 +46,6 @@ class DBConnect extends BaseMahoCommand
         ];
 
         $process = proc_open($mysqlCommand, $descriptorspec, $pipes);
-
         if (is_resource($process)) {
             // Set error pipe to non-blocking mode
             stream_set_blocking($pipes[2], false);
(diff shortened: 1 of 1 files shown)


justinbeaty:
I like it, but only one sort of problem I see:

If I use `./maho db:connect` then hit ctrl+c, the temp file remains in `/tmp` with the password.

I was trying to mess around with `pcntl_signal` to catch SIGINT and maybe other signals.

fballiano:
I have an idea and a few other things to push 😉

Il giorno sab 22 mar 2025 alle 19:09 Justin Beaty ***@***.***>
ha scritto:

fballiano:
so, the thing is we have to use a connection file because it's not secure to use the password on the command line, i'll see if I can delete the file immediately after reading it ;-)

fballiano:
I've added an autodelete (in background, after one second), it should cover pretty much all the cases

lib/MahoCLI/Commands/DBConnect.php
```diff
@@ -0,0 +1,98 @@
...
+                    break;
+                }
+                usleep(100000); // Sleep for 0.1 seconds
+            }
+
+            fclose($pipes[2]);
+            proc_close($process);
+
+            // Clean up the temporary config file
+            unlink($configFile);
```
justinbeaty, line 66: Do we still need this? Also on line 71.

justinbeaty:
I was trying to send the password directly to mysql via stdin, but it doesn't work. Then I also tried using `mysql_config_editor` which makes an obfuscated (not encrypted) `my.login.cnf` file, but it also wasn't working.

In any case I think the current implementation is great, I just was wondering if AI missed anything.

fballiano:
> I was trying to send the password directly to mysql via stdin, but it doesn't work. Then I also tried using `mysql_config_editor` which makes an obfuscated (not encrypted) `my.login.cnf` file, but it also wasn't working.

tried both thing, can't make it work securely and without user input :-\

> In any case I think the current implementation is great, I just was wondering if AI missed anything.

me too, I always triple check hehehe

lib/MahoCLI/Commands/DBConnect.php
fballiano: noup, true

justinbeaty:
Double checked, and it looks good. I had one thought that the `chmod` happens after `file_put_contents` and that could leave a small window for attackers, but `tempnam` already creates the file with `600`. But I don't see andy harm in leaving the `chmod` there.

fballiano:
true, I've put the chmod before the file_put_contents just to be extra safe

PR #145, opened 2025-03-22
Removed unused configuration_update_time setting

fballiano:
I couldn't find any place where this code was used, it seems that it was related to the get.php script that we don't have anymore.


justinbeaty (approved):
You are right, the only reference to that code is traced back to `get.php`.

PR #146, opened 2025-03-22
Removed `phpseclib/phpseclib` as a dependency

fballiano:
I've created Mage_Adminhtml_Block_System_Config_Form_Field_Heading_Packagecheck which prints out an alert for a certain group of settings (in the backend configuration section) if a package is not installed.

This way we can remove phpseclib/phpseclib as a direct dependency since it's used only for SFTP of paypal settling reports (which the vast majority do not use).

In this specific moment phpseclib/phpseclib will is be actually removed because it's still required by mcrypt_compat, but when we merge https://github.com/MahoCommerce/maho/pull/140 then all will be fixed.


app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field/Heading/Packagecheck.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+class Mage_Adminhtml_Block_System_Config_Form_Field_Heading_Packagecheck extends Mage_Adminhtml_Block_System_Config_Form_Field_Heading implements Varien_Data_Form_Element_Renderer_Interface
+{
+    #[\Override]
+    public function render(Varien_Data_Form_Element_Abstract $element): string
+    {
+        $originalData = $element->getOriginalData();
+        $package = $originalData['mandatory_package'] ?? null;
+        if ($package && \Composer\InstalledVersions::isInstalled($package)) {
```
justinbeaty, line 18:
```suggestion
        if ($package && !\Composer\InstalledVersions::isInstalled($package)) {
```

fballiano: I committed the test 🤦🏻

app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field/Heading/Packagecheck.php
```diff
@@ -0,0 +1,24 @@
...
+            $element->setLabel($element->getLabel() . "<br /> ⚠️ Install $package");
```
justinbeaty, line 19:
```suggestion
            $element->setLabel($element->getLabel() . "<br> ⚠️ Install $package");
```

Since `<br>` is a void element in HTML5 we don't need the `/>` part. I know we have tons of them in our phtml files, but one day we should clean them up.

justinbeaty:
I love this feature! Seems pretty smart to have some way to show required packages in the config section.

fballiano:
> I love this feature! Seems pretty smart to have some way to show required packages in the config section.

ideas on the plane hehehe

This way we can build stuff in the core but not having to have the dependencies installed for everybody :-)

justinbeaty (approved):

PR #151, opened 2025-03-26
Removed "popular search term" functionality from frontend

fballiano:
this is an ancient approach that's doesn't make anymore sense in 2025

Screenshot 2025-03-26 alle 09 59 20


justinbeaty (approved):
I absolutely agree!

Issue #150, opened 2025-03-26
Poor Quality in PNG to WebP Conversion


empiricompany:
I'm noticing a significant quality degradation when converting PNG images to WebP, regardless of whether the PNG has transparency or not. While JPG images convert correctly, PNGs exhibit noticeable artifacts and loss of detail.

I’m attaching my test images and results:

**png transparent**
original
![placeholder-2](https://github.com/user-attachments/assets/c48a764d-d0b1-4ea6-a462-a26064d0f6dd)
webp
![immagine](https://github.com/user-attachments/assets/bfe8db2c-0f1e-41d4-92c9-bffe36377d08)

**png no transparent**
![placeholder-2-bg](https://github.com/user-attachments/assets/0055f9c5-b1ab-4ad1-a947-6856d466e9ca)
webp
![immagine](https://github.com/user-attachments/assets/b533ca29-fff5-472a-a853-0c6bda372980)

**all jpg it's ok**
![placeholder-2-bg-flag-transparent](https://github.com/user-attachments/assets/40930fe7-e58d-4ac6-bce5-26ca1a59860d)
webp
![immagine](https://github.com/user-attachments/assets/937b32ce-a2c9-409b-b9b1-74e74a2a8383)

fballiano:
Hi @empiricompany, so the problem seems to happen only if you start with a png, other formats convert to webp fine, did I understand correctly?

empiricompany:
yes, only from png with or without a transparency

fballiano:
mmmm I'm investigating but I think we should replace the whole image magento's image generation code with a properly maintained library...

fballiano:
yes it seems to happen to me with both webp and avif

PR #152, opened 2025-03-28
Replaced internal image generation with `intervention/image`

fballiano:
This PR removes all internal image generation routines/models and add [the most used PHP image library](https://packagist.org/packages/intervention/image) (with 150M downloads).

Maho::getImageManager() automatically loads libvips driver (if available) or GD as a fallback. Libvips is so much better than GD/imagick, faster, using less memory, supports HEIC.

I've decided to remove all the Varien/Image related files because, to maintain another abstraction layer only ties us into more work without a really big benefit, at least in this case (IMHO).

I hope you'll like this PR, from my tests it seems to work nicely.

This PR fixes https://github.com/MahoCommerce/maho/issues/150


justinbeaty:
Wow, this is great! Libvips is awesome, I use it in nodejs often. Absolutely agree to drop Varien_Image.

Will test it out now!

app/code/core/Mage/Catalog/Model/Product/Image.php
```diff
@@ -433,67 +361,49 @@ public function getNewFile()
...
-    /**
-     * @param int $angle
-     * @return $this
-     */
-    public function rotate($angle)
+    public function rotate(int $angle): self
     {
         $angle = (int) $angle;
-        $this->getImageProcessor()->rotate($angle);
+        $this->getImage()->rotate($angle);
```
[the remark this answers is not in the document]
fballiano: @justinbeaty good catch as usual :-) I've changed the type and also added the background to the rotate ;-)

empiricompany:
I cleared the image cache and ran composer update, but I get this error

My php version is 8.2.2.

I think $_angle must be initialized:
```php
 protected float $_angle = 0;
```

```php
Fatal error:  Uncaught Error: Typed property Mage_Catalog_Model_Product_Image::$_angle must not be accessed before initialization in /shared/httpd/github-maho/app/code/core/Mage/Catalog/Model/Product/Image.php:324
Stack trace:
#0 /shared/httpd/github-maho/app/code/core/Mage/Catalog/Helper/Image.php(160): Mage_Catalog_Model_Product_Image->setBaseFile('/p/l/placeholde...')
#1 /shared/httpd/github-maho/app/design/frontend/base/default/template/catalog/product/list.phtml(132): Mage_Catalog_Helper_Image->init(Object(Mage_Catalog_Model_Product), 'small_image')
#2 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Template.php(274): include('/shared/httpd/g...')
#3 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Template.php(311): Mage_Core_Block_Template->fetchView('/shared/httpd/g...')
#4 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Template.php(325): Mage_Core_Block_Template->renderView()
#5 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Abstract.php(927): Mage_Core_Block_Template->_toHtml()
#6 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Abstract.php(643): Mage_Core_Block_Abstract->toHtml()
#7 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Abstract.php(587): Mage_Core_Block_Abstract->_getChildHtml('product_list', true)
#8 /shared/httpd/github-maho/app/code/core/Mage/Catalog/Block/Category/View.php(98): Mage_Core_Block_Abstract->getChildHtml('product_list')
#9 /shared/httpd/github-maho/app/design/frontend/base/default/template/catalog/category/view.phtml(57): Mage_Catalog_Block_Category_View->getProductListHtml()
#10 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Template.php(274): include('/shared/httpd/g...')
#11 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Template.php(311): Mage_Core_Block_Template->fetchView('/shared/httpd/g...')
#12 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Template.php(325): Mage_Core_Block_Template->renderView()
#13 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Abstract.php(927): Mage_Core_Block_Template->_toHtml()
#14 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Text/List.php(29): Mage_Core_Block_Abstract->toHtml()
#15 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Abstract.php(927): Mage_Core_Block_Text_List->_toHtml()
#16 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Abstract.php(643): Mage_Core_Block_Abstract->toHtml()
#17 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Abstract.php(587): Mage_Core_Block_Abstract->_getChildHtml('content', true)
#18 /shared/httpd/github-maho/app/design/frontend/base/default/template/page/3columns.phtml(35): Mage_Core_Block_Abstract->getChildHtml('content')
#19 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Template.php(274): include('/shared/httpd/g...')
#20 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Template.php(311): Mage_Core_Block_Template->fetchView('/shared/httpd/g...')
#21 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Template.php(325): Mage_Core_Block_Template->renderView()
#22 /shared/httpd/github-maho/app/code/core/Mage/Core/Block/Abstract.php(927): Mage_Core_Block_Template->_toHtml()
#23 /shared/httpd/github-maho/app/code/core/Mage/Core/Model/Layout.php(566): Mage_Core_Block_Abstract->toHtml()
#24 /shared/httpd/github-maho/app/code/core/Mage/Core/Controller/Varien/Action.php(382): Mage_Core_Model_Layout->getOutput()
#25 /shared/httpd/github-maho/app/code/core/Mage/Catalog/controllers/CategoryController.php(162): Mage_Core_Controller_Varien_Action->renderLayout()
#26 /shared/httpd/github-maho/app/code/core/Mage/Core/Controller/Varien/Action.php(413): Mage_Catalog_CategoryController->viewAction()
#27 /shared/httpd/github-maho/app/code/core/Mage/Core/Controller/Varien/Router/Standard.php(251): Mage_Core_Controller_Varien_Action->dispatch('view')
#28 /shared/httpd/github-maho/app/code/core/Mage/Core/Controller/Varien/Front.php(176): Mage_Core_Controller_Varien_Router_Standard->match(Object(Mage_Core_Controller_Request_Http))
#29 /shared/httpd/github-maho/app/code/core/Mage/Core/Model/App.php(353): Mage_Core_Controller_Varien_Front->dispatch()
#30 /shared/httpd/github-maho/app/Mage.php(662): Mage_Core_Model_App->run(Array)
#31 /shared/httpd/github-maho/public/index.php(55): Mage::run('', 'store')
#32 {main}
  thrown in /shared/httpd/github-maho/app/code/core/Mage/Catalog/Model/Product/Image.php on line 324
```

justinbeaty:
> I think $_angle must be initialized:
> 
> ```
>  protected float $_angle = 0;
> ```

This should also work if null is wanted, not sure what one is preferable.
```php
protected ?float $_angle;
```

app/code/core/Mage/Catalog/Model/Product/Image.php
```diff
@@ -73,16 +72,14 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
...
-    protected $_processor;
+    /** @var \Intervention\Image\Interfaces\ImageInterface */
+    protected $image;
 
     /**
      * @var string e.g. "small_image"
      */
     protected $_destinationSubdir;
-    protected $_angle;
+    protected float $_angle;
```
empiricompany, line 82:
```suggestion
    protected float $_angle = 0;
```

Fatal error:  Uncaught Error: Typed property Mage_Catalog_Model_Product_Image::$_angle must not be accessed before initialization in /shared/httpd/github-maho/app/code/core/Mage/Catalog/Model/Product/Image.php:324

empiricompany (changes requested):

empiricompany:
> This should also work if null is wanted, not sure what one is preferable.

I've tested with `?float`, but the error occurs because all typed variables must be initialized

empiricompany:
```php
public function setAngle(float $angle): self
{
    $this->_angle = $angle;
    return $this;
}
```

IMHO since setAngle strictly requires a float as a parameter, _angle will always be assigned a numeric value. Allowing it to be null (?float) would be unnecessary and could introduce additional checks in the code.

By initializing _angle to 0, we ensure it always has a valid numeric value, preventing potential errors when using it in calculations or other operations.

empiricompany (approved):

justinbeaty:
I will have changes requested, but just taking lunch break now. 😁

lib/Maho.php
```diff
@@ -180,4 +180,25 @@ public static function errorReport(array $reportData = [], int $httpResponseCode
...
+        $defaultOptions = [
+            'autoOrientation' => false,
+            'strip' => true,
+        ];
+        $options = [
+            ...$defaultOptions,
+            ...$customOptions,
+        ];
+
+        $driverClass = extension_loaded('vips') && \Composer\InstalledVersions::isInstalled('intervention/image-driver-vips')
```
justinbeaty, line 195:
```suggestion
        $driverClass = extension_loaded('ffi') === true && ini_get('ffi.enable') === '1' && ini_get('zend.max_allowed_stack_size') === '-1' && \Composer\InstalledVersions::isInstalled('intervention/image-driver-vips')
```

I am needing all of these conditions for vips to be used. But also encountering another error...

justinbeaty: I am Ubuntu 22.04. This github answer helped me: https://github.com/libvips/php-vips/issues/251#issuecomment-2490828273

Without this change, if you put this debug line in lib/Maho.php, are you getting `Intervention\Image\Drivers\Vips\Driver`?

```php
        Mage::log("using $driverClass");
```

fballiano: I've to say I didn't test the vips variant.

but, could we use a function_exists like function_exists('vips_version') or something like that?

justinbeaty: I am pretty sure each of these conditions are needed to be true:

1. `extension_loaded('ffi')` -- should be built-in for most PHP distributions, but probably safer to check
2. `ini_get('ffi.enable')` -- default is `auto`, which will not work, must be `true`.
3. `ini_get('zend.max_allowed_stack_size')` -- must be -1 for PHP 8.3 or above

References: 
https://github.com/libvips/php-vips?tab=readme-ov-file#install

https://github.com/libvips/php-vips/issues/251#issuecomment-2494124994

fballiano: on my macos+brew it doesn't want to work :-(

justinbeaty: Actually, the `ffi.enable = preload` (the default) seems to be working, but the readme is specific that it's not yet supported, so I'm not sure if that's old.

Without `zend.max_allowed_stack_size` we do get the error `PHP Fatal error:  Uncaught Jcupitt\Vips\Exception: zend.max_allowed_stack_size not set to '-1'`, so potentially we don't need to check for that, it just depends if we want an error thrown or GD2 fallback.

justinbeaty: > on my macos+brew it doesn't want to work :-(

I added in my php.ini:

```
[PHP]
zend.max_allowed_stack_size = -1
ffi.enable = true
```

justinbeaty: So I have vips working, but my other problem is that after a few requests my server no longer responds until I `systemctl restart php8.3-fpm`. I had a similar problem before with my nginx + php-fpm stack, but I had resolved that so I'm not sure what's going on here. If we could make it work on your machine then I'd just chalk it up to something broken on my machine...

Is one of the conditions specifically failing for you? Or all these all `1`?

```php
Mage::log([
    extension_loaded('ffi') === true,
    ini_get('ffi.enable') === '1',
    ini_get('zend.max_allowed_stack_size') === '-1',
    \Composer\InstalledVersions::isInstalled('intervention/image-driver-vips'),
]);
```

fballiano: my fpm breaks completely on start but I can't see any proper log..

tomorrow I'll try to setup a containerized environment and see if I can make it work there :-)

justinbeaty: I am able to "fix" the problem on my end by setting `pm.max_requests = 1` in `/etc/php/8.3/fpm/pool.d/www.conf`, but this is a terrible idea for production.

Still trying to figure it out... I'd really like vips to work :)

fballiano: https://github.com/libvips/php-vips/issues/119 seemed to be a progress but it's not working

lib/Maho.php
```diff
@@ -180,4 +180,25 @@ public static function errorReport(array $reportData = [], int $httpResponseCode
...
+        ];
+
+        $driverClass = extension_loaded('ffi') === true && ini_get('ffi.enable') === '1' && ini_get('zend.max_allowed_stack_size') === '-1' && \Composer\InstalledVersions::isInstalled('intervention/image-driver-vips')
+            ? \Intervention\Image\Drivers\Vips\Driver::class // @phpstan-ignore class.notFound
+            : \Intervention\Image\Drivers\Gd\Driver::class;
+
+        return new \Intervention\Image\ImageManager(
+            $driverClass,
+            ...$options,
+        );
```
justinbeaty, line 202:
```suggestion
        $driverClasses = [
            \Intervention\Image\Drivers\Gd\Driver::class,
        ];

        if (\Composer\InstalledVersions::isInstalled('intervention/image-driver-vips')) {
            // @phpstan-ignore class.notFound
            array_unshift($driverClasses, \Intervention\Image\Drivers\Vips\Driver::class);
        }

        foreach ($driverClasses as $driverClass) {
            try {
                return \Intervention\Image\ImageManager::withDriver($driverClass, ...$options);
            } catch (Intervention\Image\Exceptions\DriverException) {
            }
        }

        Mage::throwException('No image driver found');
```

This is much simpler :)

justinbeaty:
I did some performance profiling, and [this template](https://github.com/MahoCommerce/maho/blob/main/app/design/frontend/base/default/template/catalog/product/view/media.phtml#L27) was causing a lot of slowdown because `getOriginal(Width|Height)` caused Intervention to be loaded on every product page just to get the image size.

I timed how long that particular template file took to render, here's the results:

- 15ms GD driver 
- 7.5ms Vips driver
- 0.5ms With below patch

The patch is the same as how the old `Varien_Image_Adapter_Abstract` class calculated the dimensions.

```diff
diff --git a/app/code/core/Mage/Catalog/Helper/Image.php b/app/code/core/Mage/Catalog/Helper/Image.php
index 2670cc54f8..a39336f8ee 100644
--- a/app/code/core/Mage/Catalog/Helper/Image.php
+++ b/app/code/core/Mage/Catalog/Helper/Image.php
@@ -596,7 +596,7 @@ class Mage_Catalog_Helper_Image extends Mage_Core_Helper_Abstract
      */
     public function getOriginalWidth()
     {
-        return $this->_getModel()->getImage()->width();
+        return $this->_getModel()->getOriginalWidth();
     }
 
     /**
@@ -606,7 +606,7 @@ class Mage_Catalog_Helper_Image extends Mage_Core_Helper_Abstract
      */
     public function getOriginalHeight()
     {
-        return $this->_getModel()->getImage()->height();
+        return $this->_getModel()->getOriginalHeight();
     }
 
     /**
diff --git a/app/code/core/Mage/Catalog/Model/Product/Image.php b/app/code/core/Mage/Catalog/Model/Product/Image.php
index 01d512d5d7..f9b0cad415 100644
--- a/app/code/core/Mage/Catalog/Model/Product/Image.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Image.php
@@ -75,6 +75,8 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
     /** @var \Intervention\Image\Interfaces\ImageInterface */
     protected $image;
 
+    protected ?array $imageInfo = null;
+
     /**
      * @var string e.g. "small_image"
      */
@@ -128,6 +130,27 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
         return $this->_height;
     }
 
+    public function getImageInfo(): array
+    {
+        if ($this->imageInfo === null) {
+            $this->imageInfo = @getimagesize($this->_baseFile);
+            if ($this->imageInfo === false) {
+                throw new RuntimeException('Failed to read image at ' . $this->_baseFile);
+            }
+        }
+        return $this->imageInfo;
+    }
+
+    public function getOriginalWidth(): int
+    {
+        return $this->getImageInfo()[0];
+    }
+
+    public function getOriginalHeight(): int
+    {
+        return $this->getImageInfo()[1];
+    }
+
     /**
      * Set image quality, values in percentage from 0 to 100
      *
```

fballiano:
I've made vips work on https://demo.mahocommerce.com:
- updated https://github.com/MahoCommerce/demo.mahocommerce.com (both php.ini and php.Dockerfile)
- you can check FFI is enabled https://demo.mahocommerce.com/i.php
- composer.json require is this now:
Screenshot 2025-04-01 alle 09 57 25
- I've modified a bit of the code (only locally) to log the drivername that it was using and it was really vips
- delete catalog/product/cache

I've to say that, from super empirical tests libvips doesn't really seem faster... but at least it gives us the possibility to have more features down the line

so anyway on demo.mahocommerce.com everything seems stable with libvips too so I'd move forward :-)

fballiano:
I've also committed your last patch to the PR ;-)

justinbeaty:
> I've to say that, from super empirical tests libvips doesn't really seem faster... but at least it gives us the possibility to have more features down the line

I also just saw about the same performance for GD vs Vips for resizing the images, but presumably vips should at least use less memory. I'm sure there's some efficiency to be gained maybe by some sort of parallel processing.

I just pushed a commit (since I couldn't make the suggestion via Github) to reset the imageInfo array whenever the source image changed.

I'll do some final testing then approve. :)

app/code/core/Mage/Catalog/Model/Product/Image.php
```diff
@@ -475,8 +475,29 @@ public function setWatermark($file, $position = null, $size = null, $width = nul
...
+            } elseif ($this->getWatermarkPosition() === 'tile') {
+                $tile = Maho::getImageManager()
+                    ->read($filePath);
+                $element = Maho::getImageManager()
+                    ->create($this->getOriginalWidth(), $this->getOriginalHeight());
+                for ($x = 0; $x < ceil($element->width() / $tile->width()); $x++) {
+                    for ($y = 0; $y < ceil($element->height() / $tile->height()); $y++) {
+                        $element->place($tile, 'top-left', $x * $tile->width(), $y * $tile->height());
+                    }
+                }
```
fballiano, line 491: wow, super interesting

app/code/core/Mage/Catalog/etc/system.xml
```diff
@@ -462,14 +462,6 @@
                     <show_in_website>1</show_in_website>
                     <show_in_store>1</show_in_store>
                     <fields>
-                        <size translate="label">
-                            <label>Watermark Default Size</label>
-                            <sort_order>100</sort_order>
-                            <show_in_default>1</show_in_default>
-                            <show_in_website>1</show_in_website>
-                            <show_in_store>1</show_in_store>
-                            <comment>Example format: 200x300.</comment>
```
fballiano, line 479: what I would find more useful (but I wouldn't care if it's missing now, it a percentage of the size of the image that the watermark is applied to) because this way it auto adapts to various source image sizes

justinbeaty:
Found (and fixed) a few bugs:

03504504dd Fix system_config_backend_file upload path

System Config file uploads were being placed in `/media` instead of `/public/media`. This bug wasn't introduced in this PR, but I fixed it so I could test watermarks.

82f0d30642 Fix watermark

We needed to handle stretch and tile sizing options manually. Also the watermark dimensions option is now pointless, so I removed it but we still have a lot of `->setWatermarkSize()` calls which I haven't yet cleaned up.

1d7fb185c3 Fix configurable swatches

The old varien image must have ensured a directory exists before saving it, but intervention doesn't. This bug happens when you clear the swatch image cache.

9f1b8eb3bf Fix wysiwyg thumbnailAction

I needed to add back this action for uploading images for example at CMS > Static Blocks > Insert Image...

justinbeaty:
I tested various watermarks, but it only seems to be working with GD. With vips I get the error: `Jcupitt\Vips\Exception: libvips error: pngload: no property named 'n'`.

image

I also tested rotation and background color. 😄 

image

app/code/core/Mage/Catalog/etc/system.xml
justinbeaty: I was also thinking we could have options for cover and contain, same as CSS [background-size](https://developer.mozilla.org/en-US/docs/Web/CSS/background-size).

Actually, there should be at least two fields: position (top, left, etc) and size (cover, contain, stretch, tile, percentage). Right now they're combined  into one dropdown which doesn't give as much flexibility, so that's definitely a larger effort.

justinbeaty:
A few last notes:

1\. We have the Mage_Media module, but I don't know where it is used (if at all). It still uses built-in functions for creating and resizing images. I don't think it needs to be taken care of in this PR, but something to keep in mind.

2\. AVIF isn't working for me with either GD or libvips, but I think this might just be that I have older libraries installed.

3\. Once you update demo.mahocommerce.com, you could try both AVIF and watermarks with vips. I'm hopeful it works there without issue.

Otherwise, I'll approve if you can't find any other issues. :smiley:

justinbeaty (approved):

fballiano:
> 1. We have the Mage_Media module, but I don't know where it is used (if at all). It still uses built-in functions for creating and resizing images. I don't think it needs to be taken care of in this PR, but something to keep in mind.

what the heck is mage_media, I swear I didn't even knew it existed, that's the next thing we'll nuke :-D

> 2. AVIF isn't working for me with either GD or libvips, but I think this might just be that I have older libraries installed.

I can confirm avif is working for me on my computer with GD

> 3. Once you update demo.mahocommerce.com, you could try both AVIF and watermarks with vips. I'm hopeful it works there without issue.

I'll test it out right now

fballiano:
@justinbeaty avif working on with vips on demo https://demo.mahocommerce.com/lafayette-convertible-dress.html but it seems to be extremely slow

but when I try to upload the watermark picture I get
Screenshot 2025-04-01 alle 18 35 34

fballiano:
when I add a PNG watermark with AVIF/WEBP target generation (with libvips) I get

Screenshot 2025-04-01 alle 18 39 25

justinbeaty:
> EDIT: SORRY my bad! opcache reset was needed

Haha, how could you make such a mistake! I've *certainly* never done that!! 😅 

> what the heck is mage_media, I swear I didn't even knew it existed, that's the next thing we'll nuke :-D

I have no idea either! I checked `git log` on magento-mirror and it was introduced in Magento 1.1.1, and if I `git diff` between that and the latest OM, it's 99% formatting changes, and a few lines adding webp support. I'm guessing it is no longer used and is just legacy cruft.

> when I add a PNG watermark with AVIF/WEBP target generation (with libvips) I get

Same on my end. If you check exception log I think you'll see the same thing about `pngload: no property named 'n'`. I don't think I'd let that hold us up on this PR, but I'm sure it's fixable somehow.

fballiano:
the demo is almost crashing now... I'll remove the libvips thing and hope it goes back to a stable situation 😅 so I'd say libvips not that stable anyway... I don't know if I'd use that...

fballiano:
demo working again after removing the vips driver

Screenshot 2025-04-01 alle 18 50 53

justinbeaty:
The homepage blocks don't seem to be working on demo, but they do on my local machine:

image

Do we have any extension that generates thumbnails via CLI or admin action?

fballiano:
they are now
Screenshot 2025-04-01 alle 18 57 59

fballiano:
ok I've reverted everything to the default demo configuration, I'll merge and update

❤️❤️❤️

PR #154, opened 2025-03-31
Frontend: an autocomplete that actually makes sense

fballiano:
Autocomplete has always been awful and this is an area we have to improve sooner than later, with this PR I think we're making a huge jump ahead.

we could implement a million more features, but as an MVP I think it's a good result

Screenshot 2025-03-31 alle 22 18 26


justinbeaty (approved):
Looks good, and working as expected.

fballiano:
🥹 I wasn't expecting to get it this easy ehehheh

PR #155, opened 2025-04-01
Removed Mage_Media module

fballiano:
the only "real" usage of this module was `class Mage_Catalog_Model_Product_Media_Config implements Mage_Media_Model_Image_Config_Interface`


justinbeaty (approved):
EZ approve.

I wonder why nothing in Mage_Image even implemented that interface... Seems like long dead code.

PR #158, opened 2025-04-10
Removed HTTP(s) schema from URLs in web installation process

fballiano:
I'm testing a cloud infrastructure based on docker container (for which I'm creating https://hub.docker.com/r/mahocommerce/maho). This infrastructure has its own balancer/SSL terminator in front of the containers, so that it always connect to the PHP containers via HTTP (not HTTPS).

Problem is that the website runs in HTTPS, but when the web installer starts, Maho thinks it's running in HTTP because of what we said in the previous sentence. This causes the skin/js URL generation to fail, they will be generated with HTTP prefix but the inclusion will fail, breaking the web installation process.

This PR makes the web installer have `//` schema instead of the complete one, but that would break the "base url" generation for the "config" step of the installer, which I also fixed in app/code/core/Mage/Install/Model/Installer/Config.php.

In the long run I think we should remove the unsecure URLs completely, nowadays everybody has a single configuration, which can be the secure one (usually), but there's really no need for the double configuration anymore, but that would be for another PR.


justinbeaty:
I definitely agree in the long run there should only be one base url, and users can make that either http or https depending on their needs.

In the meantime, I’ll review this PR first thing tomorrow!

fballiano:
I'm thinking about the next step in using modern dockerized infrastructures and we have a problem with local.xml
- it has to be backed in the docker image, bad idea
- app/etc should be mounted as a volume (most providers support mounting "drives" in some directory but if we mount `app/etc` then we're losing all files that are in the container image in that directory), this is also not ideal

we could use environment variables override to re-configure what's in local.xml, but if the local.xml file doesn't exist (let's say re-deploying the container image) then the installation wizard will be re-triggered.

i also have no idea if the local.xml configurations can really be overridden via environment variables...

not all infrastructure providers have the possibility to write a configuration file on deploy so we're kinda stuck

we could have something in the startup script of the container that reads some specific environment variables and writes them to the local.xml... that's kinda the only way that I can think about, without rewriting the whole local.xml management... which in 2025 is really outdated

justinbeaty:
Another possible solution is either an env variable that points to an alternative path where Maho can find local.xml, or even just a built-in standard location we check, like app/docker/local.xml

fballiano:
nice ideas, I think I'd do both things:
- have an environment variable to set the path of the local.xml
- have an environment variable with the whole local.xml content (for infrastructures that would make you have a "disk" for every mounted folder, which is costly but also can't be usually used with multiple frontends, since "disks" can usually be attached to a single machine)

justinbeaty:
Question, is the `X-Forwarded-Proto: https` header set in the docker config? In nginx I use `proxy_set_header X-Forwarded-Proto $scheme;` which correctly sets the scheme, and I'm not sure I've ever had this problem.

fballiano:
mmm I wouldn't patch frankenphp configuration there, but also because I wouldn't know how to :-\

justinbeaty:
> mmm I wouldn't patch frankenphp configuration there, but also because I wouldn't know how to :-\

It looks like you would use the [trusted_proxies](https://caddyserver.com/docs/caddyfile/options#trusted-proxies) directive, which would make FrankenPHP set the `X-Forwarded-Proto: https` to Maho. Something like this, but it depends on what IP address the SSL terminator is at:

```
{
	servers {
		trusted_proxies static private_ranges
	}
}
```

To be clear, I'm not against the PR as it is, but I think you might run into other problems without the trusted proxies setup, such as all order placed from IPs being reported as 127.0.0.1.

justinbeaty:
@fballiano Not sure if you saw my last comment here from Friday. As mentioned, I have no problem approving, but I still think the trusted_proxies config could solve it. Especially if any end users will use this sort of configuration, they'll want to have that option configured.

fballiano:
yes yes, I'm wondering if that's necessary becuse the list of trusted IPs can't be baked in the docker image so yes I could configure it for myself (although I really would like to avoid changing the caddyfile just for that) but the end users? :-\

fballiano:
right now I'm waiting to see if docker allows us into the open-source sponsorship for their pro plain, it would allow to automatic builds which would be so much better, but it will take up to a couple of weeks to get an answer :-(

justinbeaty (approved):

justinbeaty:
> yes yes, I'm wondering if that's necessary becuse the list of trusted IPs can't be baked in the docker image so yes I could configure it for myself (although I really would like to avoid changing the caddyfile just for that) but the end users? :-\

Yeah, makes sense that it's not an easy problem to solve. And at the very least, I'd rather Maho be able to be installed but maybe some IPs are just reported wrong. Approved. :)

fballiano:
web installer working on https://sevalla.com via latest docker image 

Screenshot 2025-04-14 alle 22 35 48


:-)))))

PR #159, opened 2025-04-14
Added possibility to set `local.xml` content via environment variable `MAHO_LOCAL_XML`

fballiano:
This PR is needed for scalable infrastructures where local.xml is not in the container at time of deployment, we need a way to set the contents of local.xml via ENV variable (which is usually managed by dockerized infrastructure providers).

Since this PR adds a file system hit with the file_exists, I've removed the is_readable later, which IMHO is not necessary, so the hits on the file system stay the same.

**Note:**
I chose not to have the possibility to set the `path`of the file, because I didn't want to generate an inclusion of unwanted files like /etc/passwd or something like that

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -754,7 +754,7 @@ final class Mage
                 }
             }
 
-            if ($localConfig = simplexml_load_file($localConfigFile)) {
+            if ($localConfig = @simplexml_load_file($localConfigFile)) {
                 date_default_timezone_set('UTC');
                 if (($date = $localConfig->global->install->date) && strtotime((string) $date)) {
                     self::$_isInstalled = true;
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -746,8 +746,15 @@ final class Mage
 
             self::$_isInstalled = false;
 
-            if (is_readable($localConfigFile)) {
-                $localConfig = simplexml_load_file($localConfigFile);
+            $localXmlContent = $_ENV['MAHO_LOCAL_XML'] ?? $_SERVER['MAHO_LOCAL_XML'] ?? null;
+            if (!empty($localXmlContent) && !file_exists($localConfigFile)) {
+                $result = file_put_contents($localConfigFile, $localXmlContent, LOCK_EX);
+                if ($result === false) {
+                    throw new Exception("Failed to write $localConfigFile.");
+                }
+            }
+
+            if ($localConfig = simplexml_load_file($localConfigFile)) {
                 date_default_timezone_set('UTC');
                 if (($date = $localConfig->global->install->date) && strtotime((string) $date)) {
                     self::$_isInstalled = true;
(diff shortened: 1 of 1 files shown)


fballiano:
If nobody is agains I'd merge this and https://github.com/MahoCommerce/maho/pull/158 in a few days in order to test the official docker images on a docker platform provider ;-)

justinbeaty:
Pulled the PR, but I think we actually still need the `is_readable` check because otherwise I can't install with developer mode on:

```
Warning: simplexml_load_file(): I/O warning : failed to load external entity "/home/www/maho/app/etc/local.xml"  in /home/www/maho/app/Mage.php on line 757

#0 [internal function]: mageCoreErrorHandler()
#1 /home/www/maho/app/Mage.php(757): simplexml_load_file()
#2 /home/www/maho/app/code/core/Mage/Core/Model/Config.php(467): Mage::isInstalled()
#3 /home/www/maho/app/code/core/Mage/Core/Model/Config.php(368): Mage_Core_Model_Config->_canUseCacheForInit()
#4 /home/www/maho/app/code/core/Mage/Core/Model/App.php(428): Mage_Core_Model_Config->loadModulesCache()
#5 /home/www/maho/app/code/core/Mage/Core/Model/App.php(341): Mage_Core_Model_App->_initModules()
#6 /home/www/maho/app/Mage.php(662): Mage_Core_Model_App->run()
#7 /home/www/maho/public/index.php(51): Mage::run()
#8 {main}
```

I don't think the `is_readable` check should add any extra file system hits, except for the very first time you load the page right? And that's only if the ENV variable is set.

fballiano:
I didn't notice that simplexml_load_file was throwing a warning, but we have no need for that warning because the function already return false in case there's any error and the is_readable doesn't really prevent the error in case the xml file in readable but invalid

justinbeaty (approved):
Looks good now.

fballiano:
@justinbeaty I made a mistake, `app/etc`may not exist on a child project generated this way: https://github.com/mahoCommerce/docker-images

:-(((

should I just add a `@mkdir($etcDir, 0640, true);` before the `file_put_contents`?

justinbeaty:
> @justinbeaty I made a mistake, `app/etc`may not exist on a child project generated this way: https://github.com/mahoCommerce/docker-images
> 
> :-(((
> 
> should I just add a `@mkdir($etcDir, 0640, true);` before the `file_put_contents`?

D’oh! I didn’t think of that. Yeah I think that’s a good solution.

fballiano:
and I think permissions should be 750 cause it's a folder :-\

justinbeaty:
> and I think permissions should be 750 cause it's a folder :-\

Yes, sorry I missed that.

fballiano:
it works! https://maho-41tee.kinsta.app/

(temporary URL, will probably be deleted sometimes soon)

fballiano:
This previous URL is already a redeployed docker image, checking that local.xml wasn't existing at time of deploy, and got created via ENV variable etc etc, niceeeeee

justinbeaty:
> This previous URL is already a redeployed docker image, checking that local.xml wasn't existing at time of deploy, and got created via ENV variable etc etc, niceeeeee

Awesome! This feature should definitely come in handy. :)

PR #161, opened 2025-04-26
New default values for search configurations

fballiano:
- Search is now in "AND" instead of "OR"
- the min query lenght is 3 chars and the max is 50
- redirect in case of a single result is enabled by default


justinbeaty (approved):
Looks very straightforward. I see no problems with the new defaults.

PR #162, opened 2025-04-27
Added default `main_table.` prefix to filter_index in grids that use the additional `_grid` table.

fballiano:
This PR should avoid "ambiguous column name" error when adding the main table (without the _grid suffix) as a join.


justinbeaty:
Going to review this next. At first glance, I see no problems though.

I do recall there is some way to add custom columns to the `_grid` tables, but I'm certainly guilty of using joins or additional queries to add stuff to the grids. I don't recall getting any errors, but I'll look at some of my code to check.

fballiano:
yep we could add more columns to the _grid (although I don't know why 😅) but probably  it shouldn't be mandatory

PR #163, opened 2025-04-28
Removed `Varien_Db_Adapter_Pdo_Mysql::_removeDuplicateEntry()` when creating an index

fballiano:
I think deleting records from the DB it's too dangerous, exception should bubble up and that's it.


justinbeaty (approved):
I agree with the change. I created an empty database and installed with no problem, just in case one of the install scripts tried to add a duplicate index. Of course it's possible this could break a 3rd party extension, but I think throwing an exception is better.

PR #164, opened 2025-04-29
`Mage::getVersion()` now returns Maho version

fballiano:
This PR also removes Magento's edition constants from `app/Mage.php`


fballiano:
super! thanks :-)

PR #165, opened 2025-04-29
Rewrote the whole `cache` system based on `symfony/cache`, removing the old `Zend_Cache`

fballiano:
This PR rewrites the whole caching system that's now based on symfony/cache, much more modern and maintained.

The new backends are:
- file
- redis (now supported by the core without any additional modules)

The full configuration possibilities are, in `local.xml`:
```xml
<global>
    <cache>
        <lifetime>86400</lifetime> <!-- optional, default is 7200 -->
        <backend>file|redis</backend> <!-- optional, default is file -->
        <backend_options> <!-- optional for "file", required for "redis" -->
            <dsn>redis://localhost?dbindex=0</dsn>
        </backend_options>
    </cache>
</global>
```

It also removes the support for 2levels cache, which is now obsolete in 2025.

**Notes:**
- at the moment caching for Zend_Locale (app/code/core/Mage/Core/Model/Locale.php) is disabled, this is based on a custom patch to Zend_Locale itself ("OM-918 - Add runtime cache to Zend_Locale_Data": "https://raw.githubusercontent.com/MahoCommerce/maho-composer-patches/e75f155864138ef4ee2b18a82797d6f40d71ad51/OM-918.patch"). we can port that patch to symfony/cache or completely remove it (we will have to migrate away from Zend_Locale anyway sometimes in the future)


lib/Varien/Db/Adapter/Pdo/Mysql.php
```diff
@@ -131,7 +131,7 @@ class Varien_Db_Adapter_Pdo_Mysql extends Zend_Db_Adapter_Pdo_Mysql implements V
     /**
      * Cache frontend adapter instance
      *
-     * @var Zend_Cache_Core
+     * @var Mage_Core_Model_Cache
```
fballiano, line 134: in theory any Mage_ shouldn't be here but... I don't know, I think we can forget about this

justinbeaty: Yeah, I wouldn't worry about this as I'm pretty sure there's already Mage specific stuff there. Plus the while Varien lib is really only used for Mage.

app/code/core/Mage/Admin/sql/maho_setup/maho-24.12.1.php
```diff
@@ -4,7 +4,7 @@
  * Maho
  *
  * @package    Mage_Core
- * @copyright  @copyright  Copyright (c) 2025 Maho (https://mahocommerce.com)
+ * @copyright  Copyright (c) 2025 Maho (https://mahocommerce.com)
```
fballiano, line 7: these were just typos I found while working on this PR

app/code/core/Mage/Core/Model/Resource/Store/Collection.php
```diff
@@ -193,14 +189,6 @@ public function initConfigCache()
...
         $needsRefresh = false;
 
-        if ($globalConfigCache !== false) {
-            try {
-                $data = unserialize($globalConfigCache);
-            } catch (Exception $exception) {
-                Mage::logException($exception);
-            }
-        }
-
```
fballiano, line 203: the old version was storing only strings in the cache, this new version stores everything

app/code/core/Mage/Core/Model/Cache.php
```diff
@@ -336,65 +190,55 @@ protected function _tags($tags = [])
...
      */
-    public function load($id)
+    public function load(string $id): mixed
     {
-        return $this->getFrontend()->load($this->_id($id));
+        $item = $this->getFrontend()->getItem($this->_id($id)); // @phpstan-ignore method.internalClass
+        if ($item->isHit()) {
+            return $item->get();
+        }
+        return false;
```
fballiano, line 208: this (and the return false) is done to maintain the previous behavior, I've tested not checking for isHit() but all of the framework broke so...

justinbeaty:
There are still some references to `Zend_Cache_Core` in `lib/Varien/Db/Adapter/Pdo/Mysql.php`, for example here:

https://github.com/MahoCommerce/maho/blob/2d7bfe6e33dd60dc534761450455df0ae7c025da/lib/Varien/Db/Adapter/Pdo/Mysql.php#L1593-L1601

Should these be replaced?

fballiano:
ah, nice catch, they wouldn't work anymore so yep I'll remove them immediately :-)

lib/Varien/Db/Adapter/Pdo/Mysql.php
```diff
@@ -1590,14 +1590,9 @@ public function loadDdlCache($tableCacheKey, $ddlType)
             return $this->_ddlCache[$ddlType][$tableCacheKey];
         }
 
-        if ($this->_cacheAdapter instanceof Zend_Cache_Core) {
+        if ($this->_cacheAdapter) {
```
fballiano, line 1593: @justinbeaty actually I've just removed the check, cause I think the rest of the code should be executed, right?

justinbeaty: Yeah, it should because if not `null` then `$this->_cacheAdapter` should be an instance of `Mage_Core_Model_Cache`

fballiano: do you think we should have a check? it seems a bit redundant to me

justinbeaty: I don't think it's necessary since we have the PHP type checking.

fballiano:
mmm yeah, today's changes broke something :-\ probably the ones in Pdo/Mysql.php :-\

fballiano:
yep exactly, if I revert Pdo/Mysql.php everything works fine :-\

justinbeaty:
Yeah, I think it's due to a missing unserialize somewhere, because for example we're trying to iterate over this string which should be an array:

```
string(6078) "a:17:{s:12:"attribute_id";a:14:{s:11:"SCHEMA_NAME";N;s:10:"TABLE_NAME";s:13:"eav_attribute";s:11:"COLUMN_NAME";s:12:"attribute_id";s:15:"COLUMN_POSITION";i:1;s:9:"DATA_TYPE";s:17:"smallint
```

lib/Varien/Db/Adapter/Pdo/Mysql.php
```diff
@@ -1590,14 +1590,9 @@ public function loadDdlCache($tableCacheKey, $ddlType)
...
             $cacheId = $this->_getCacheId($tableCacheKey, $ddlType);
-            $data = $this->_cacheAdapter->load($cacheId);
-            if ($data !== false) {
-                $data = unserialize($data, ['allowed_classes' => false]);
-                $this->_ddlCache[$ddlType][$tableCacheKey] = $data;
-            }
```
justinbeaty, line 1599: Probably these lines I'd guess

fballiano:
@justinbeaty actually I've removed the serialize instead, it seems to work

lib/Varien/Db/Adapter/Pdo/Mysql.php
justinbeaty: Nevermind, I missed your previous comment about it supporting non-strings.

fballiano: ;-) I missed the serialize hehehe

app/code/core/Mage/Core/Model/App.php
```diff
@@ -1196,13 +1179,11 @@ public function saveCache($data, $id, $tags = [], $lifeTime = false)
...
-     *
-     * @param   string $id
-     * @return  false|int
      */
-    public function testCache($id)
+    public function testCache(string $id): bool
     {
-        return $this->_cache->test($id);
+        // @phpstan-ignore method.internalClass
+        return $this->_cache->getFrontend()->getItem($id)->isHit();
```
justinbeaty, line 1186: Should we leave this as:
```php
        return $this->_cache->test($id);
```
?

app/code/core/Mage/Checkout/Block/Cart/Sidebar.php
```diff
@@ -273,12 +268,9 @@ public function getCacheTags()
 
     /**
      * Processing block html after rendering
-     *
-     * @param   string $html
-     * @return  string
      */
     #[\Override]
-    protected function _afterToHtml($html)
+    protected function _afterToHtml(string $html): string
```
justinbeaty, line 273: Two questions:

1) Should this be in this PR?

2) Are we deciding to start adding PHP types on methods? It does break any block that overrides this method.

app/code/core/Mage/Api/Model/Wsdl/Config.php
```diff
@@ -53,7 +53,7 @@ public static function getNamespacesPrefix()
     }
 
     /**
-     * @return Varien_Simplexml_Config_Cache_Abstract|Zend_Cache_Core
+     * @return Mage_Core_Model_Cache|Varien_Simplexml_Config_Cache_Abstract|null
```
justinbeaty, line 56: I think this should just be `Varien_Simplexml_Config` now that we also changed `Varien_Simplexml_Config`, right?

app/code/core/Mage/Core/Model/Cache.php
justinbeaty: It doesn't look like `isHit` has very much overhead ([ref](https://github.com/symfony/symfony/blob/11495f445f0ffb8f55981d2a176b5f0525fcdf53/src/Symfony/Component/Cache/CacheItem.php#L48-L51)), but could this work too?

```php
return $item->get() ?? false;
```

fballiano: @justinbeaty would it make sense to store `null` in cache? cause if the cache is `hit` but null, I'd still return null 🤔

app/code/core/Mage/Core/Model/App.php
fballiano: super right, changing now

app/code/core/Mage/Core/Model/Observer.php
```diff
@@ -82,7 +77,7 @@ public function addSynchronizeNotification(Varien_Event_Observer $observer)
      */
     public function cleanCache(Mage_Cron_Model_Schedule $schedule)
     {
-        Mage::app()->getCache()->clean(Zend_Cache::CLEANING_MODE_OLD);
+        Mage::app()->getCache()->clean();
```
justinbeaty, line 80: This method is called by cron every hour, and it should clean outdated cache entries, but this is flushing the entire cache, right?

I think we need the following in `Mage_Core_Model_Cache`:
```php
    public function prune(): bool
    {
        if ($this->_frontend implements \Symfony\Component\Cache\PruneableInterface) {
            return $this->_frontend->prune();
        }
        return true;
    }
```

And call `Mage::app()->getCache()->prune();` instead, but I haven't tested this yet.

app/code/core/Mage/Checkout/Block/Cart/Sidebar.php
fballiano: you're right, I think I ended up on that code for some reason related to this PR, but then probably something got solved some other way and I left this part thinking it was harmless but... better to revert all of them

app/code/core/Mage/Core/Model/Cache.php
justinbeaty: Yes, you're correct and I hadn't thought of that when I first wrote my comment, so no change needed.

app/code/core/Maho/Captcha/Helper/Data.php
```diff
@@ -100,7 +100,7 @@ public function verify(string $payload): bool
         try {
             $altcha = new Altcha\Altcha($this->getHmacKey());
             $isValid = $altcha->verifySolution($payload, true);
-            Mage::app()->getCache()->clean(Zend_Cache::CLEANING_MODE_OLD, [self::CACHE_TAG]);
+            Mage::app()->getCache()->clean([self::CACHE_TAG]);
```
justinbeaty, line 103: I think we want to actually remove this line, because otherwise it's clearing all captcha cache entries on each altcha attempt. The `self::CHALLENGE_EXPIRATION` on the next line will invalidate entries anyway.

app/code/core/Mage/Core/Model/Observer.php
fballiano: absolutely, done in https://github.com/MahoCommerce/maho/pull/165/commits/e13d094d4371bafc9fdcf7d2e3860d0cf91a6c7a

app/code/core/Maho/Captcha/Helper/Data.php
fballiano: true, done it in https://github.com/MahoCommerce/maho/pull/165/commits/55ff79027c73aca809c907e30e766a6669917521, but I'm wondering how it was working before, only because we tested our single use and not concurrent use?

justinbeaty: I think it worked before because it used `CLEANING_MODE_OLD`, which would only prune expired entries. But I'm not sure that was ever necessary in the first place, maybe it was legacy code from before entries auto-expired?

app/code/core/Mage/Api/Model/Wsdl/Config.php
fballiano: mmm are you sure? the getCache method strictly returns Mage_Core_Model_Cache,  c623fac0b1

am I missing something?

justinbeaty: Sorry I had made a typo and updated the comment. I also meant that we can only return `Mage_Core_Model_Cache`.

justinbeaty: Also here:
app/code/core/Mage/Api/Model/Config.php
app/code/core/Mage/Core/Model/Config.php

fballiano: done it in f8de2a9a6e :-)

app/code/core/Maho/Captcha/Helper/Data.php
fballiano: now I'm thinking all the CLEANING_MODE_OLD I've changed in this PR is actually wrong, I've to check them all

fballiano: actually there aren't any other CLEANING_MODE_OLD that were removed, better 😅

fballiano:
@justinbeaty I think all your comments should be fixed now 🙏🎉

retesting it now, seems to work fine

app/code/core/Mage/Api/Model/Resource/Roles.php
```diff
@@ -77,7 +72,7 @@ protected function _beforeSave(Mage_Core_Model_Abstract $role)
     protected function _afterSave(Mage_Core_Model_Abstract $role)
     {
         $this->_updateRoleUsersAcl($role);
-        Mage::app()->getCache()->clean(Zend_Cache::CLEANING_MODE_MATCHING_TAG);
+        Mage::app()->getCache()->clean();
```
justinbeaty, line 75: I sort of wonder why this is here, but since saving roles doesn't happen often I don't think it's an issue.

fballiano: you mean the original line?

justinbeaty: Yeah, the original line, because it flushes the entire cache, but then why didn't they write `Mage::app()->getCache()->flush()` ?

fballiano: yeah, it's a bit weird, should we change it to flush()?

justinbeaty:
> @justinbeaty I think all your comments should be fixed now 🙏🎉

I will have some more tomorrow, but must stop for tonight. Overall I think this is a very good addition!

app/code/core/Mage/Core/Model/Cache.php
```diff
@@ -405,43 +249,33 @@ public function remove($id)
...
-    /**
-     * Get adapter for database cache backend model
-     *
-     * @return Zend_Db_Adapter_Abstract
-     */
-    public function getDbAdapter()
+    public function flush(): bool
     {
-        return Mage::getSingleton('core/resource')->getConnection($this->_dbConnection);
+        return $this->_frontend->clear($this->_idPrefix); // @phpstan-ignore method.internalClass
```
justinbeaty, line 272: Here we are clearing all entries starting with our prefix, but we never used the prefix when setting keys. Thus, nothing is flushed from the cache.

Example:

```php
<?php

Mage::init();

$key = 'foo';
$val = 'bar';

Mage::app()->getCache()->save($val, $key, [], 5);
Mage::app()->getCache()->flush();

$data = Mage::app()->getCache()->load($key);
var_dump($data); // Prints 'bar'
```

---

If not provided, the prefix will be something like `4c8_`. I would imagine prefixes are useful in cases where you might have multiple instances using the same Redis cache, and probably some other edge-cases, so I don't think we should simply drop the functionality.

Instead, here's two possible solutions:

1\. Append the prefix to the namespace, so it's something like `maho_4c8`.

2\. This patch, which is [similar to how Magento 1.4.0.0-beta-1 did it](https://github.com/OpenMage/magento-lts/blame/788d81fd55312f0d7d462ef10f0317f9a08fab71/app/code/core/Mage/Core/Model/Cache.php#L268): 

```diff
diff --git a/app/code/core/Mage/Core/Model/Cache.php b/app/code/core/Mage/Core/Model/Cache.php
index eead1afd9e..1a8aa00225 100644
--- a/app/code/core/Mage/Core/Model/Cache.php
+++ b/app/code/core/Mage/Core/Model/Cache.php
@@ -171,7 +171,7 @@ class Mage_Core_Model_Cache
     protected function _id($id)
     {
         if ($id) {
-            $id = strtoupper($id);
+            $id = $this->_idPrefix . strtoupper($id);
         }
         return $id;
     }
```

--- 

I've only tested the second option, but I think the first option would actually be more performant when flushing the cache since it doesn't need to iterate and check str_starts_with. Possibly also more performant with read/write depending on the adapter too.

app/code/core/Mage/Core/Model/Cache.php
```diff
@@ -458,12 +292,12 @@ protected function _initOptions()
...
             if (is_array($options)) {
                 $this->_allowedCacheOptions = $options;
-                $this->save(serialize($this->_allowedCacheOptions), self::OPTIONS_CACHE_ID);
+                $this->save($this->_allowedCacheOptions, self::OPTIONS_CACHE_ID);
             } else {
                 $this->_allowedCacheOptions = [];
             }
         } else {
-            $this->_allowedCacheOptions = unserialize($options, ['allowed_classes' => false]);
+            $this->_allowedCacheOptions = $options;
```
justinbeaty, line 300: I am concerned about removing `['allowed_classes' => false]` here. The [default "marshaller"](https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/Cache/Marshaller/DefaultMarshaller.php) also uses `unserialize()`, but has no way of controlling the options. Also it would seem [igbinary](https://github.com/igbinary/igbinary/issues/123) doesn't support `allowed_classes` at all, although that extension needs to be explicitely enabled.

So I think that we should revert the changes to this method. We already have [other places](https://github.com/MahoCommerce/maho/blob/f8de2a9a6e9d6d80ce419f0dfb389705a43b927d/app/code/core/Mage/Core/Model/Store.php#L368-L371
) that we manually serialize / unserialize, so it might just be something we can't escape.


The one downside is that these values end up double-serialized, which isn't great. We could potentially write our own marshaller that supports `allowed_classes`, or raise the issue on the Symfony repo. But I am not sure it's possible to conditionally set that option when storing an item in the cache, and I am also not sure if there are cases in Maho where we *do* want to unserialize classes.

lib/Varien/Db/Adapter/Pdo/Mysql.php
justinbeaty: Revisiting this, same issue as https://github.com/MahoCommerce/maho/pull/165#pullrequestreview-2825997206

I think the issue is that if an attacker finds a way to write a file into `var/cache`, then it could turn into an arbitrary code execution.

fballiano:
Hey @justinbeaty I've created a super basic marshaller, without igbinary support, which we never tested anyway (do you think it's ok this way?), what do you think about it?

justinbeaty:
Regarding `Mage:app()->getCacheInstance()`  vs `Mage:app()->getCache()`, the two classes had nearly identical interfaces, except for one method:

```php
// Old code, getCache()->clean() required two parameters
Mage::app()->getCache()->clean(
    Zend_Cache::CLEANING_MODE_MATCHING_TAG,
    [Mage_Adminhtml_Block_Page_Menu::CACHE_TAGS],
);

// While getCacheInstance()->clean() required only one parameter (string or array)
Mage::app()->getCacheInstance()->clean([Mage_Adminhtml_Block_Page_Menu::CACHE_TAGS]);
```

With this PR, we would break all 3rd party code that called `Mage::app()->getCache()->clean()`. Instead, we could check to see if two parameters were passed and handle it appropriately: 

```php
/**
 * Clean cached data by specific tag
 *
 * @param   array|string $tags
 * @return  bool
 */
public function clean($tags = [])
{
    $args = func_get_args();
    if (count($args) > 1 && is_array($args[1])) {
        $tags = $args[1];
    }

    if (!empty($tags) && !is_array($tags)) {
        $tags = [$tags];
    }

    if (is_array($tags) && count($tags) > 0) {
        return $this->_frontend->invalidateTags($this->_tags($tags)); // @phpstan-ignore method.internalClass
    }

    return $this->flush();
}
```

I tested and it all seems to work. What do you think?

fballiano:
> Regarding `Mage:app()->getCacheInstance()` vs `Mage:app()->getCache()`, the two classes had nearly identical interfaces, except for one method:
> 
> ```
> // Old code, getCache()->clean() required two parameters
> Mage::app()->getCache()->clean(
>     Zend_Cache::CLEANING_MODE_MATCHING_TAG,
>     [Mage_Adminhtml_Block_Page_Menu::CACHE_TAGS],
> );
> 
> // While getCacheInstance()->clean() required only one parameter (string or array)
> Mage::app()->getCacheInstance()->clean([Mage_Adminhtml_Block_Page_Menu::CACHE_TAGS]);
> ```
> 
> With this PR, we would break all 3rd party code that called `Mage::app()->getCache()->clean()`. Instead, we could check to see if two parameters were passed and handle it appropriately:
> 
> ```
> /**
>  * Clean cached data by specific tag
>  *
>  * @param   array|string $tags
>  * @return  bool
>  */
> public function clean($tags = [])
> {
>     $args = func_get_args();
>     if (count($args) > 1 && is_array($args[1])) {
>         $tags = $args[1];
>     }
> 
>     if (!empty($tags) && !is_array($tags)) {
>         $tags = [$tags];
>     }
> 
>     if (is_array($tags) && count($tags) > 0) {
>         return $this->_frontend->invalidateTags($this->_tags($tags)); // @phpstan-ignore method.internalClass
>     }
> 
>     return $this->flush();
> }
> ```
> 
> I tested and it all seems to work. What do you think?

I'm ok with it, I've just pushed your fix :-)

app/code/core/Mage/Core/Model/Cache.php
fballiano: I don't know :-( but the previous implementation had the allowed_classes to false so it's probably better to leave it that way, no?

app/code/core/Mage/Core/Model/Cache.php
fballiano: > If not provided, the prefix will be something like `4c8_`. I would imagine prefixes are useful in cases where you might have multiple instances using the same Redis cache

I'd think that, if multiple instances use the same redis, at least they would use different databases in the 


> 1. Append the prefix to the namespace, so it's something like `maho_4c8`.

I like this idea a bit more than the second one and if it's slightly more performant it's even better, I've committed it here: https://github.com/MahoCommerce/maho/pull/165/commits/65057d1bf4d30bd381865fc6830b54cccc70888e what do you think?

app/code/core/Mage/Core/Model/Cache.php
justinbeaty: Here's what I found:

Previously, all cache values were required to be strings (which we knew), UNLESS you had this in your local.xml:

```xml
<cache>
    <frontend_options>
        <automatic_serialization>1</automatic_serialization>
    </frontend_options>
</cache>
```

With the above, you could serialize an instance of a class, and it would automatically be restored when reading the cache. This is what we want to avoid as it's a security issue.

Anytime the Mage devs wanted to cache an object, they manually called serialize and unserialize since only strings were allowed, and `<automatic_serialization>` was not the default behavior. For example [here](https://github.com/MahoCommerce/maho/blob/f8de2a9a6e9d6d80ce419f0dfb389705a43b927d/app/code/core/Mage/Core/Model/Store.php#L368-L371), but there are more if you `git grep allowed_classes`

So the bottom line, I believe we are safe to use your new marshaller, and remove all cases of serialize / unserialize before caching *as long as* the unserialize call has `[allowed_classes' => false]`.

If some code did want to restore instances of a class, they can manually call both serialize and unserialize (without `[allowed_classes' => false]`). The data will be double serialized, but not a huge issue.

I hope that makes sense... I can help identify any places we need to change, just after a little coffee break.

app/code/core/Mage/Core/Model/Cache.php
justinbeaty: Yeah, I think the first option is better, and that’s how I would have implemented it too. You just need to change the line this suggestion is on to:

    return $this->_frontend->clear();

then I assume it should all work, but I’ll also do some testing.

app/code/core/Mage/Core/Model/Cache.php
fballiano: > If some code did want to restore instances of a class, they can manually call both serialize and unserialize (without `[allowed_classes' => false]`). The data will be double serialized, but not a huge issue.

absolutely :-)

> I hope that makes sense... I can help identify any places we need to change, just after a little coffee break.

do you think there are places we still need to change?

justinbeaty: > do you think there are places we still need to change?

Any place that calls serialize before caching, and deserialize (with allowed classes false) when reading cache should have the (de)serialize calls removed. Else we are double encoding, which doesn’t cause problems, but is a lot less efficient.

There’s at least one place this is true, it was the link I posted in my previous comment. Others could be found by grepping, we just need to make sure it’s not serializing for any other reason like storing in the session or DB.

app/code/core/Mage/Core/Model/Cache.php
```diff
@@ -116,94 +83,34 @@ public function __construct(array $options = [])
...
-
-        if (isset($options['disallow_save'])) {
-            $this->_disallowSave = (bool) $options['disallow_save'];
-        }
+        $this->_frontend = match ($backend['type']) {
+            'redis' => new \Symfony\Component\Cache\Adapter\RedisTagAwareAdapter(
+                \Symfony\Component\Cache\Adapter\RedisTagAwareAdapter::createConnection($backend['options']['dsn']),
+                "maho-{$this->_idPrefix}",
+                $frontend['lifetime'],
+                new \Maho\CacheMarshaller(),
```
justinbeaty, line 91: What do you think about putting our marshaller class at `app/code/core/Mage/Core/Model/Cache/Marshaller.php` or similar, and replacing the above line with `Mage::getModel('core/cache_marshaller')` ?

I think there could be some situations where a module wants to overwrite this, and it would be less great to require the developer to override `Mage_Core_Model_Cache`'s constructor.

fballiano: I don't super too much love it but I think it's worth doing it :-)))) just pushed it

justinbeaty:
@fballiano here's a patch to fix the places that previously called `serialize` manually before storing in the cache; which was necessary before as we could only store strings.

But since our marshaller does this for us, these places were effectively doing `serialize(serialize($data))` and `unserialize(unserialize($data))`.

As mentioned in the other comment, it wouldn't cause problems, but it is less efficient both storage and performance wise. Smaller data had the largest speed increase, with the double serialization being about 50% slower. That didn't translate to a huge increase in terms of overall milliseconds per page load, but I think it's still better to fix.

I'm attaching here so you can review, hopefully it makes more sense now in case my comments weren't quite clear. I haven't fully tested it yet, but so far the parts I have tested seem to work fine.

Download: [serialize.patch](https://github.com/user-attachments/files/20112713/serialize.patch)

justinbeaty:
I have only a few more small suggestions... 😅 But it will have to wait until tomorrow, then some final testing and we should be ready to merge!

fballiano:
> I have only a few more small suggestions... 😅 But it will have to wait until tomorrow, then some final testing and we should be ready to merge!

thanks man, it's always inspiring to work together with you

fballiano:
> @fballiano here's a patch to fix the places that previously called `serialize` manually before storing in the cache; which was necessary before as we could only store strings.
> I'm attaching here so you can review, hopefully it makes more sense now in case my comments weren't quite clear. I haven't fully tested it yet, but so far the parts I have tested seem to work fine.

wow thanks, and btw your comments make total sense, it's me to get overwhelmed sometimes also because it's 00.40 AM now so my brain is not that awake 😅😅😅

justinbeaty:
Some comments regarding `Mage_Core_Model_Cache`:

1\. I think this method is now obsolete:
```php
    /**
     * Get options for database backend type
     *
     * @return array
     */
    protected function getDbAdapterOptions(array $options = [])
    {
        if (isset($options['connection'])) {
            $this->_dbConnection = $options['connection'];
        }

        $options['adapter_callback'] = [$this, 'getDbAdapter'];
        $options['data_table'] = Mage::getSingleton('core/resource')->getTableName('core/cache');
        $options['tags_table'] = Mage::getSingleton('core/resource')->getTableName('core/cache_tag');
        return $options;
    }
```

---

2\. I think a few of the lines in the below are now obsolete:

```php
    /**
     * Get options of cache frontend (options of symfony/cache)
     *
     * @return  array
     */
    protected function _getFrontendOptions(array $cacheOptions)
    {
        $options = $cacheOptions['frontend_options'] ?? [];
        if (!array_key_exists('caching', $options)) {
            $options['caching'] = true;
        }
        if (!array_key_exists('lifetime', $options)) {
            $options['lifetime'] = $cacheOptions['lifetime'] ?? self::DEFAULT_LIFETIME;
        }
        // These lines aren't needed, right?
        // if (!array_key_exists('automatic_cleaning_factor', $options)) {
        //     $options['automatic_cleaning_factor'] = 0;
        // }
        // $options['cache_id_prefix'] = $this->_idPrefix;
        return $options;
    }
```

---

3\. Also in `_getFrontendOptions()`, we have the `caching` option that could be disabled via `local.xml`. That was implemented in `vendor/shardj/zf1-future/library/Zend/Cache/Core.php` in the save, load, test, clean, etc methods like this:

```php
    public function load($id, $doNotTestCacheValidity = false, $doNotUnserialize = false)
    {
        if (!$this->_options['caching']) {
            return false;
        }
        // ...
    }

    public function save($data, $id = null, $tags = [], $specificLifetime = false, $priority = 8)
    {
        if (!$this->_options['caching']) {
            return true;
        }
        // ...
    }
```

Should we move these checks to those same methods in `Mage_Core_Model_Cache` ? If not then we can also remove those lines from `_getFrontendOptions()`

---

4\. How do you feel about renaming `$this->_frontend` to `$this->_cacheAdapter` (with or without underscore)? That's how it is named in `lib/Varien/Db/Adapter/Pdo/Mysql.php` and I think it makes more sense. I don't think this can really be considered an additional breaking change because Symfony cache is already 100% different API than Zend was. So if someone did something like extend `Mage_Core_Model_Cache`, any references to `$this->_frontend` are already broken.

fballiano:
> 1. I think this method is now obsolete:

removed!

> 2. I think a few of the lines in the below are now obsolete:

removed!

> 3. Also in `_getFrontendOptions()`, we have the `caching` option that could be disabled via `local.xml`. That was implemented in `vendor/shardj/zf1-future/library/Zend/Cache/Core.php` in the save, load, test, clean, etc methods like this:

It seems to me that disabling a cache is something due to the many problems some old systems had (especially with block_html cache), I mean, in production could there be specific reasons to disable a cache? I'd remove 'caching' from _getFrontendOptions but at that point _getFrontendOptions only manages the lifetime?

> 4. How do you feel about renaming `$this->_frontend` to `$this->_cacheAdapter` (with or without underscore)? That's how it is named in `lib/Varien/Db/Adapter/Pdo/Mysql.php` and I think it makes more sense. I don't think this can really be considered an additional breaking change because Symfony cache is already 100% different API than Zend was. So if someone did something like extend `Mage_Core_Model_Cache`, any references to `$this->_frontend` are already broken.

Renamed! PHPStorm also renamed the getFrontend() method to getCacheAdapter(), should we keep the old method name?
Should we also rename frontend_options to cacheAdapterOptions?

justinbeaty:
> It seems to me that disabling a cache is something due to the many problems some old systems had (especially with block_html cache), I mean, in production could there be specific reasons to disable a cache?

I guess the local.xml option would disable all caches, as opposed to only disabling specific caches like block_html. This probably existed as a way to disable caching of things outside of the categories listed on the admin cache management page. However I agree that I don't actually see a need for this option to exist any more. Certainly you wouldn't do it in production.

---

> I'd remove 'caching' from _getFrontendOptions but at that point _getFrontendOptions only manages the lifetime?

That was going to be my next thought. I guess we had both `_getFrontendOptions()` and `_getBackendOptions()` because of the structure of local.xml:

```xml
<cache>
    <backend>redis</backend>
    <backend_options>
        <dsn>...</dsn>
    </backend_options>
    <frontend_options>
        <lifetime>7200</lifetime>
    </frontend_options>
</cache>
```

But I feel like the frontend vs backend distinction makes no sense now with Symfony cache. Should we just completely change the structure? Maybe something like:

```xml
<cache>
    <adapter>redis</adapter>
    <adapter_options>
        <dsn>...</dsn>
        <lifetime>7200</lifetime>
    </adapter_options>
</cache>
```

Looking at [colinmollenhour/Cm_Cache_Backend_Redis](https://github.com/colinmollenhour/Cm_Cache_Backend_Redis), we had `<backend>Cm_Cache_Backend_Redis</backend>` which is the name of a class and not a keyword like `redis`, and the `<backend_options>` is completely different. So it already wouldn't be a drop-in replacement.

> Should we also rename frontend_options to cacheAdapterOptions?

So yeah, I think one combined `getCacheAdapterOptions()` or similar is the way to go.

---

> PHPStorm also renamed the getFrontend() method to getCacheAdapter(), should we keep the old method name?

I think it's not necessary for the same reason that `Zend_Cache_Core` is completely different than `AbstractTagAwareAdapter`, so it's not like keeping the old method name would preserve backwards compatibility.

fballiano:
dangerous all this reorganization, I'll try to finish it later cause now I've to leave for a bit

fballiano:
@justinbeaty I've refactored a few things and tests are ok :-)

I'm wondering, does `$_disallowSave` still make sense or should we remove it? cause there's no way of setting it to `true`, if I'm not mistaken

justinbeaty:
> @justinbeaty I've refactored a few things and tests are ok :-)

Great, I'll do final testing now.

> I'm wondering, does `$_disallowSave` still make sense or should we remove it? cause there's no way of setting it to `true`, if I'm not mistaken


I don't see any way to set it to true either, short of overwriting the class. It also sort of overlaps the `<caching>` option that we removed, so I'd say we should remove `$_disallowSave` too. If we ever want to re-introduce these sort of options, we could do it in a way that has more fine-grained controls.

fballiano:
> If we ever want to re-introduce these sort of options, we could do it in a way that has more fine-grained controls.

that would make much more sense :-)

removed the disallowSave already ;-)

lib/Varien/Db/Adapter/Pdo/Mysql.php
```diff
@@ -1643,9 +1637,7 @@ public function resetDdlCache($tableName = null, $schemaName = null)
         }
         if ($tableName === null) {
             $this->_ddlCache = [];
-            if ($this->_cacheAdapter instanceof Zend_Cache_Core) {
-                $this->_cacheAdapter->clean(Zend_Cache::CLEANING_MODE_MATCHING_TAG, [self::DDL_CACHE_TAG]);
-            }
+            $this->_cacheAdapter->clean([self::DDL_CACHE_TAG]);
```
justinbeaty, line 1640:
```suggestion
            if ($this->_cacheAdapter) {
                $this->_cacheAdapter->clean([self::DDL_CACHE_TAG]);
            }
```

I don't know under which circumstance we wouldn't have `$_cacheAdapter`, but we left the checks everywhere else  so I think we should add here.

app/code/core/Mage/Core/Model/Resource/Store/Collection.php
justinbeaty: But since we don't need to `unserialize`, it would just be:

```php
        if ($globalConfigCache !== false) {
            $data = $globalConfigCache;
        }
```

And then also removing the `serialize` call at the end of this method.

justinbeaty: @fballiano also tagging in case you won't see this otherwise

fballiano: @justinbeaty you got a big bug with that code removal, which prevented the needsRefresh from working correctly!

I've committed https://github.com/MahoCommerce/maho/pull/165/commits/d3a661eb6592b4fbd348e683f0f07b839eb7d5ac which also removes `initConfigCache()` from `app/code/core/Mage/Core/Model/Store.php` since it seems not to be used anywhere and phpstan is not complaining nor can I see any error.

🙏

justinbeaty: I had written in my notes to double check this, but almost forgot before finalizing my review. Glad we got it fixed. :)

I agree with the removal of that old function. I think it was just legacy stuff since it saved a separate cache entry for each store config.

justinbeaty (approved):
No more changes from me. I've tested both file and redis backends. Everything seems to be working great. 😃

fballiano:
this is great! let's go!

PR #166, opened 2025-05-05
Added Meta Pixel (Facebook Pixel) support to the GoogleAnalytics module

fballiano:
Nowadays GA4/GTM and enhanced ecommerce data are a must for any ecommerce platform.

Meta pixel is also extremely important, so I thought about expanding the googleanalytics module (which should be called Web Analytics, instead of Google API, at least in the backend) adding this functionality in the easiest way possible.

I think it's a google and important addition to our platform.


fballiano:
I've retested and it seems ok, the code is the same as the original GA4 implementation which was deeply tested, the variable naming should conform the facebook standards, so I'll merge this for 25.5

PR #169, opened 2025-05-08
Solved CodeQL issue 68

fballiano:
CodeQL started complaining about this:
https://github.com/MahoCommerce/maho/security/code-scanning/68

There's one usage that we should fix:
Screenshot 2025-05-08 alle 22 22 01

Those are coming from this:
Screenshot 2025-05-08 alle 22 25 52

I'll leave this PR in draft since at the moment I can't finish it

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Catalog/CategoryController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Catalog/CategoryController.php
@@ -144,11 +144,13 @@ class Mage_Adminhtml_Catalog_CategoryController extends Mage_Adminhtml_Controlle
         if ($this->getRequest()->isAjax()) {
             $this->_renderTitles();
 
+            $messagesCollection = $this->getLayout()->getMessagesBlock()->getMessageCollection();
             $eventResponse = new Varien_Object([
                 'title' => implode(' / ', array_reverse($this->_titles)),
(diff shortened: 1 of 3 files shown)


justinbeaty:
I'll take a look soon, but I think there's an easier way to fix this.

justinbeaty:
> There's one usage that we should fix: 
> [Screenshot of public/js/mage/adminhtml/catalog/category.js]

There are actually more cases where AJAX calls return the result of `getGroupedHtml()` and we use innerHTML. For example in the admin varienForm JS class:

https://github.com/MahoCommerce/maho/blob/61b1d35bbdbbf6b4179b50cb65f907fdb66b39ab/public/js/mage/adminhtml/form.js#L69-L82

It's just that since the above class hadn't been re-written to vanilla JS, I hadn't updated it to use the `setMessagesDivHtml()` function.

So we would have to modify all admin form controllers that use the validate action. There's also places where the messages HTML is passed to observers which would need to be changed too.

So before we go down that rabbit hole, I pushed a potential alternative solution to see if it passes CodeQL. If not, then we'll revert and continue down the original path.

fballiano:
oh wow, this is so much better and it seems to have solved it! thank you!

fballiano:
@justinbeaty https://github.com/MahoCommerce/maho/security/code-scanning/68 is fixed! ❤️

PR #171, opened 2025-05-14
Removed deprecated methods

fballiano:


fballiano:
let's leave it like this, not to make it too big

justinbeaty:
I’ll review tomorrow. 😁

justinbeaty:
Got caught up with other stuff today, but will review first thing tomorrow.

justinbeaty:
This all looks good, just one comment:

We only just marked the `_getHelper()`, `_getModel()`, etc method as deprecated two months ago [in this commit](https://github.com/MahoCommerce/maho/pull/66/commits/523c1fed883265f6d3e346e9e75c9e7cfb58ebdc#diff-3b1b6c330ad5e64651d667f88f52f68d345c669518aa055cf93871cb4adb7b7e). I originally left those so that PHPStan could inform users how to update their code. However in this case it's extremely obvious to just change it to `Mage::getModel()`, etc.

So I'm not against the removal of these methods now, but should we remove all the occurrences of these methods linked in the above commit?

fballiano:
I've reverted those changes, better do that another time :-)

PR #179, opened 2025-05-29
Replaced TinyMCE with TipTap 2.25.0

fballiano:
### Why remove TinyMCE?
- The GPL-licensed version of TinyMCE is incompatible with Maho's OSL-3.0 license
- The MIT-licensed version (6.8) is unmaintained and has unpatched security vulnerabilities and mark a couple of our repositories as "problematic", this is a no go

### Notes
- new SVG placeholders (at the moment there's only one for all widgets but the old gifs where really ugly in 2025)
- new attribute parsing routine based on DOM instead of regex


fballiano:
@justinbeaty I've also seen https://playground.lexical.dev (made by Meta) which seems interesting and has some functionalities similar to a page builder...............

fballiano:
on both PRs we now have double-click to edit a widget, which wasn't a thing before ;-)

fballiano:
I tried to do have a proof of code for Lexical too, but no way to make it work, so let's choose between quill and tiptap :-)

justinbeaty:
I'll be reviewing this and some other PRs today. :)

justinbeaty:
> it seems to me tiptap is maybe a better choice because it seems to be under more active development, it's a bit weird that the quilljs repository is very quite...

Starting to review this one now, and so far I think it probably is the better choice.

app/code/core/Mage/Cms/etc/jstranslator.xml
```diff
@@ -9,7 +9,7 @@
  */
 -->
 <jstranslator>
-    <script path="mage/adminhtml/wysiwyg/tinymce/setup.js" module="cms">
+    <script path="mage/adminhtml/wysiwyg/quill/setup.js" module="cms">
```
justinbeaty, line 12:
```suggestion
    <script path="mage/adminhtml/wysiwyg/tiptap/setup.js" module="cms">
```

app/design/adminhtml/default/default/template/newsletter/template/edit.phtml
```diff
@@ -47,130 +47,188 @@
...
-            $('save_as_flag').value = '1';
+        document.getElementById('convert_button').style.display = 'none';
+        document.getElementById('convert_button_back').style.display = '';
 
-            if(this.isEditor()) {
-                tinymce.triggerSave();
+        // Strip HTML tags and scripts
+        const tempDiv = document.createElement('div');
+        tempDiv.innerHTML = textElement.value;
+        textElement.value = tempDiv.textContent || tempDiv.innerText || '';
```
justinbeaty, line 92: Since we have a helper function in `varien/js.js`:

```suggestion
        // Strip HTML tags and scripts
        textElement.value = stripTags(textElement.value);
```

app/design/adminhtml/default/default/template/system/email/template/edit.phtml
```diff
@@ -103,163 +103,236 @@
...
+        document.getElementById('convert_button').style.display = 'none';
+
+        // Strip scripts, styles and HTML tags
+        let cleanedValue = templateText.value
+            .replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi, '')
+            .replace(/<style[^>]*>[\s\S]*?<\/style>/gi, '');
+
+        const tempDiv = document.createElement('div');
+        tempDiv.innerHTML = cleanedValue;
+        templateText.value = (tempDiv.textContent || tempDiv.innerText || '').trim();
```
justinbeaty, line 141: Related to previous comment. I think that one should strip `<script>` and `<style>` tags too, but I'm not a huge fan of the regex.

What about updating our `stripTags()` function to look like the below? It's similar to the `xssFilter()` helper function. 

```js
function stripTags(str, removeScriptAndStyleContent=false) {
    const doc = new DOMParser().parseFromString(str, 'text/html');
    if (removeScriptAndStyleContent) {
        doc.querySelectorAll('script').forEach(script => script.remove());      
        doc.querySelectorAll('style').forEach(style => style.remove());      
    }
    return doc.body.textContent;
}
```

Although tbh, `removeScriptAndStyleContent` could be default and probably not even an option as I don't really see why you'd ever want the text contents of scripts or styles. It's just that PrototypeJS's `stripTags()` doesn't do this and they have a different `stripScripts()` function.

fballiano: mmm at that point it would be better to directly use `xssFilter` no? but I'm not sure for email templates maybe styles should be kept, I think

justinbeaty: I was going to say we could also use `xssFilter()`, but right now it only strips `<script>` tags. We could also make it strip `<style>` tags, but for sure would need to make it optional (and off by default). Between modifying these two, I think `stripTags()` is probably better.

As far as removing `<style>` tags for email, it should be fine because this is for the "Convert to Plain Text", correct?

justinbeaty:
Here is a patch that changes the mahoFetch then/catch chains to async which IMO is much more readable. Technically we don't even need the try/catch, but I used it to show an alert dialog instead of it just failing silently. I also used `setBodyJson` so that we can skip the manual JSON parsing.

[maho-fetch-async.patch](https://github.com/user-attachments/files/20558577/maho-fetch-async.patch)

public/js/mage/adminhtml/browser.js
```diff
@@ -223,6 +223,8 @@ class Mediabrowser {
     getTargetElement() {
         if (typeof tinyMCE !== 'undefined' && tinyMCE.get(this.targetElementId)) {
```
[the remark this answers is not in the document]
fballiano: I thought I removed them all 😅

done!

justinbeaty:
A couple notes on inserting widgets / images:

1\. When I insert a widget, the `` tag initially has some broken HTML, however if I save and continue it seems okay.

image
image

2\. If I disable the WYSIWYG, I cannot add a widget or image.

justinbeaty:
Overall, this is really nice and is much more future-proof. I think it's probably the correct direction over quilljs, since I think tiptap (and ProseMirror) will live on for a long time. :)

fballiano:
mmmmm weird (the widget thing) cause I've tested it 1000 times, but I'll re-check tomorrow :-)

fballiano:
> Overall, this is really nice and is much more future-proof. I think it's probably the correct direction over quilljs, since I think tiptap (and ProseMirror) will live on for a long time. :)

agree, and compared to quilljs, the generated HTML is cleaner. the whole thing seems more robust.

fballiano:
> 1. When I insert a widget, the `` tag initially has some broken HTML, however if I save and continue it seems okay.

I can't reproduce this, could you re-try?

> 2. If I disable the WYSIWYG, I cannot add a widget or image.

those two should be fixed now!

justinbeaty:
Hmm.. I was thinking we'd always need an update to `Mage_Page_Block_Html_Head` to support adding scripts as modules, but seems we can already do this.

The commit I pushed seems to work fine, and we don't have to use `waitForTiptapModules()` anymore.

fballiano:
tested, it was working with my latest commit sorryyyy

justinbeaty:
> it probably broke in some previous changes but now it doesn't activate on "product -> edit product -> WYSIWYG Editor -> show" :-(

> tested, it was working with my latest commit sorryyyy

It was just because I used `DOMContentLoaded`, but that has already fired by the time the edit product WYSIWYG modal pops up.

Should we put that `onReady` function somewhere? It might be useful in other places, potentially for an idea I have for #193.

justinbeaty:
Hmmm... weird bug caused by PrototypeJS.

image

image

Seems to be caused by Prototype redefining `Array.prototype.find`

Can't wait to get rid of Prototype. I've had similar error with code like this:
```js
const x = [1,2,3]
x.entries();
// Uncaught TypeError: x.entries is not a function
```

justinbeaty:
> I think the widget title is solved now.

Yup!

> I woulnd't care about the prototype bug at the moment, hoping to get rid of it sometimes in the future :-)

Yeah, I was getting pretty close in the admin redesign. I'm feeling really motivated to finish that up now. Just been busy reviewing all your PRs!!!!! 😂😂😂

fballiano:
> Yeah, I was getting pretty close in the admin redesign. I'm feeling really motivated to finish that up now. Just been busy reviewing all your PRs!!!!! 😂😂😂

😂😂😂 sorryyyy

fballiano:
no hurry :-)

justinbeaty:
A couple of small bugs so far:

```html
// Hide WYSIWYG and enter this:
<p><a href="mailto:{{config path="trans_email/ident_sales/email"}}">email me</a></p>

// Switch back to WYSIWYG:
<p><a target="_blank" rel="noopener noreferrer nofollow" href="mailto:{{config path=">email me</a></p>
```

---

Add a widget in WYSIWYG mode. Then either save or toggle editor off/on. Then you can resize the widget the tooltip is different.

Not sure how hard this one is to fix, and it's not a huge deal.

image

justinbeaty:
Still working on this... I've improved the directive handling a bunch and am currently working on getting TipTap to work with `System > Transactional Emails`:

image

fballiano:
niceeeee :-)

justinbeaty:
Also found a few bugs in TipTap, made a PR: https://github.com/ueberdosis/tiptap/pull/6448 😁

fballiano:
> Also found a few bugs in TipTap, made a PR: [ueberdosis/tiptap#6448](https://github.com/ueberdosis/tiptap/pull/6448) 😁

😱😱 you're a beast

justinbeaty:
> @justinbeaty if you've something pushable I'd like to try updating to tiptap v3 beta, it's almost out and filament4 is already using tiptap3 😍

I can push today. Need to just reorganize my local commits so they're easier to review.

justinbeaty:
> hey @justinbeaty, could you just push the code as it is, no need to reorganize it :-)

Yes, sorry I will. A huge deadline just came up at work and have been 24/7 on it.

justinbeaty:
Pushed, will explain the changes tomorrow. I sincerely apologize about the delay again. Only by chance did I check my Github notifications yesterday right after you wrote, otherwise I've been non-stop working on this deadline. Luckily I finished it since it was due Friday.

justinbeaty:
> i'll check it out ASAP (I'm in the middle of another idea).
> 
> question, since there's a conflict with uploader.phtml I checked it and it seems almost empty, can't it just be removed at that point?

I'll take a look and solve conflicts. It might have something to do with the CMS media gallery page.

justinbeaty:
Here's an overview on what I changed:

## TipTap Changes

### Directive Handling

The biggest change is here, and the directive stuff should be much more maintainable and extensible. We handle directives in just two parts:

1. When syncing between the editor and textarea using the `convertFromPlain()` function. This is the necessary first step because the textarea can contain invalid HTML, and then TipTap can't parse it. It also converts non-attribute directives into `<span data-type="maho-widget">` so that TipTap can recognize it as a node instead of just text. The `convertToPlain()` function does this in reverse. 

2. A custom Node View extension that recognizes that span element as a widget. Here is where we parse the actual directive value.

This is almost identical to what you had already, but a few main improvements:

* Uses a `<span>` instead of an `` tag allowing TipTap to not confuse widgets as images, avoiding all the special widget handling code in the `CustomImage` extension.
* Allows us to define many different widget types, which is how I added support for `{{var}}`, `{{customvar}}`, `{{config}}`, etc.
* The `insertMahoWidget()` and `insertMahoVariable()` logic is contained within the widget node view extension.
* We don't need to base64 encode all attribute directives into image preview URLs (see other stuff section for more info).

### Generic Toolbar Creation

I liked what you had for creating the table bubble menu, so I extended it to define the main toolbar as well. Most commands can be done with just `command` and `args` properties, but you can define custom functions too.  I think this was like -200 lines of code which makes me happy. :) The `updateToolbarState()` function is also simpler, and I was able to add support for the text align buttons showing their current state.

### Remove all Hardcoded TipTap References

I think we will be using TipTap for the long haul, but I think it's better for the codebase to be wysiwyg agnostic. I removed  explicit "tiptap" references in various files (catalog.js, widget.js, newsletter phtml, etc) and replaced it with a generic interface. Old Magento had hardcoded references to TinyMCE, so I know you basically had just replaced those.

### Other stuff

**Split setup.js and extensions.js:** Mainly because `setup.js` was really long, but also because there will be more custom extensions in the future. :)

**Beautify HTML:** Tiptap creates minified HTML, which sucks if you simply toggle on/off the editor. Or if you ever need to manually edit HTML, etc. I found the lightest weight formatter that is still maintained and works in the browser. It adds 30kB, but we're loading like 130kB for Tiptap so I don't think it's an issue.

**Debounced wysiwyg -> plain conversion:** if the editor has lots of content, we avoid syncing on every keystroke. It now only syncs if not typing for 200ms. In case someone somehow submits a form in that 200ms, we also sync on various form submit events. I added a `formValidate` event to the adminhtml `varienForm`, not sure why we didn't even have that before.

**Don't destroy / setup when toggling the editor:** I know tinymce did this, but I don't see a reason for it. I think it's better to just leave it initialized, and sync the content when toggling.

**Image alt text editor:** I added an alt text field in the media browser. Now double-click allows you to change alt or change image. This is good for replacing images in the editor without having to delete, place new, resize again, etc. Somehow this was more of a pain that it should have been.

**Image directive preview URLs:** This is the part that turns `` to ``. This turns the directive to Base64, and an admin controller turns that into a preview image. But we were also encoding tags like `<a href="mailto:{{var ...}>` this way, which of course would turn into a broken image.

Because of the broken image, I found out that the placeholder image was broken itself. We were trying to use an SVG as the placeholder, but the default Intervention GD handler doesn't support SVGs. Instead of returning some placeholder, I just made it return a `500` status, and we deal with showing a placeholder in the wysiwyg code.

I *also* realized that we were using `@getimagesize($path)` to get the width / height of the preview image, but we already have that info from Intervention so no need to process the image twice. 

**Catalog wysiwyg popup:** Small bug, I just noticed that if the text area contained any `{{widget}}`, when you opened the wysiwyg they weren't parsed. I think it was caused by the `initialContent` only being processed in `setup()`, and we were setting the content of the wysiwyg after `setup()` when the dialog is ready. It will be easier to avoid this sort of error with the dedicated `convertFromPlain()` / `convertToPlain()` functions.


## Other Improvements

### Maho Dialog

A couple of things here
1. I added support for the HTMLDialog's `returnValue` property. We hadn't needed this anywhere yet, but it let's us get some value back from the `Dialog.info()` call.
2. I separated the `ok` (bool) and `onOk` (function) options for more flexibility. Same with `cancel` / `onCancel`. Then finally added an `onOpen` function so we don't have to do any `setTimeout(..., 20)` tricks.

### Variables.js / Widget.js / Browser.js

I did some of this in #87, but since we are now totally ditching TinyMCE, I cleaned this up. I made each of these files more consistent in order to facilitate not having any hardcoded TipTap references in them. Instead they all use the new `returnValue` feature. This is as opposed to each of those files needing to find the correct TipTap instance and call `insertContent` directly. The variables popup UI is also nicer. 


### Newsletter / Transactional Email Forms

I realized we only resolved https://github.com/MahoCommerce/maho/pull/179#pullrequestreview-2889615659 in the newsletter form, and not on the transactional email form. 90% of the code was duplicated between these three phtml files, and those long script blocks in template files drives me nuts for various reasons. I refactored this into an `email-template.js` file. It's not strictly necessary in this PR, but it's relevant enough. I also found a switch from html / plain bug that's also present in OM.

justinbeaty:
I think there are a few things worth doing before merge:

1\. Editing the home page is completely broken. The main reason for this is TipTap by default treats images as "block" nodes. That means they can't belong in any inline nodes, and if they do, tiptap will force them out. This breaks the slideshow and promos block. If it were any other page that was broken I'd let it go, but the homepage being editable seems like a good requirement for merging. 😅 

The slideshow part is a pretty easy fix. We can make a custom node view extension that recognizes `<div class="slideshow">`. This doesn't need to be a slideshow editor, just to prevent tiptap from breaking the HTML. An editor would be nice, but 100% not needed today. I think there's a lot we can add in the future anyway.

2\. I disabled the wysiwyg on System > Transactional Emails because I hadn't finished adding support for it. This page never had a wysiwyg editor before, but it would be really nice for it to happen. The main thing I needed to add there was special handling of `{{depend some_var}} {{/depend}}`, since that's the only directive that comes in two parts. However I realized the Newsletter Template also has these, so it would be good to fix.

Also, if there were any bugs I introduced, I'll fix them.

fballiano:
wow, jawdrop as usual! I can't test it in this very moment but will do as soon as I finish this idea I'm working on ;-)


> 1. Editing the home page is completely broken. The main reason for this is TipTap by default treats images as "block" nodes. That means they can't belong in any inline nodes, and if they do, tiptap will force them out. This breaks the slideshow and promos block. If it were any other page that was broken I'd let it go, but the homepage being editable seems like a good requirement for merging. 😅

question, does that mean that tiptap doesn't allow images inside any `div` or `p`? that seems a bit of a problem

> 2. I disabled the wysiwyg on System > Transactional Emails because I hadn't finished adding support for it. This page never had a wysiwyg editor before, but it would be really nice for it to happen. The main thing I needed to add there was special handling of `{{depend some_var}} {{/depend}}`, since that's the only directive that comes in two parts. However I realized the Newsletter Template also has these, so it would be good to fix.

it would be an awesome addition, not blocking to merge this IMHO but really nice to have 😍

fballiano:
@justinbeaty the new media library opens this way:

Screenshot 2025-06-28 alle 00 23 22

but it should be like this:

Screenshot 2025-06-28 alle 00 24 29

justinbeaty:
> question, does that mean that tiptap doesn't allow images inside any `div` or `p`? that seems a bit of a problem

By default, it only allows images inside a `<div>`, `<p>`, `<td>` and other block elements, but not `<a>`, `<span>`, `<li>`, etc. However, I just realized that the Image extension has an option for this, not sure how I missed it. I pushed that so actually the homepage breaks a little less.

But, it actually also strips `<div>` elements too. I'm sure there's a good solution for it though. I guess one of the points of tiptap is that content should follow a schema, and if something isn't allowed it is removed. This should be a good thing for content writers since they can't make things look awful, but maybe it could cause problems.

It looks like quill does the same thing though, so I think we can make it work with tiptap. One solution that I just checked is setting:

```js
this.editor = new TiptapModules.Editor({
    enableContentCheck: true,
    // ...
```

Then if the HTML is not valid in the eyes of TipTap, it won't activate. If you set the above option and set wysiwyg as "Disabled by Default" you'll see that you can't enable the editor. It also has an option to strip out the invalid stuff, but emit a warning. So we could do one of two things:

1. Disable the wysiwyg if there's invalid HTML, and making a confirm box when enabling like: "Enabling the wysiwyg may remove custom HTML".
2. Keep it enabled by default (if that's what you set), but show a warning box either above the editor, or when you press save or something similar.

So basically only the first time you'd have to "convert" your custom HTML to what's allowed by tiptap. But after that you'll always have nice and clean HTML. I'm sure I'm not the only person who has had people mess up the HTML by going in there and editing things, so ultimately it would probably be for the best.

Plus, tiptap makes it easy to define custom node types. So we can do stuff like allow `<div class="slideshow">`, and a bunch of other stuff. I think ultimately it's a path to some sort of simplified page builder.

justinbeaty:
> @justinbeaty the new media library opens this way:

Oh sorry, I think that's just cause it looks for the last opened folder from your session. I actually don't like that feature for the dedicated CMS page. The Manage Categories page used to work like that too. Will fix it.

> edit: the idea of adding the media library is actually because I'd like to add an image-editor functionality but i'm not sure if we want to have that when we're selecting an image to enter in a textarea (merely because of the denseness of the UI in that specific use case). does this implementation allow for that?

Yeah, absolutely it will, and that's an excellent idea.

I'd just write the functionality into `browser.js` and then we have a [`setupObject`](https://github.com/MahoCommerce/maho/blob/main/app/code/core/Mage/Adminhtml/Block/Cms/Wysiwyg/Images/Content.php) that you do `->canEditImage(true)`. That's how I handled allowing inserting images for the dialog:

https://github.com/MahoCommerce/maho/blob/ce521f0414c2f4761191e7874c558d9c230ce6c8/app/code/core/Mage/Adminhtml/controllers/Cms/Wysiwyg/ImagesController.php#L46-L59

I think it's just cleaner than `browser.js` having one thing, and then having to override that behavior in another template file. Cause  if that's the case, it seems the Media library should be the default behavior and you override the modal.

justinbeaty:
> I'd just write the functionality into browser.js 

Btw, what I mean by this is just the stuff that would handle stuff like showing the "edit" button when you click an image, opening the dialog (if that's how you'll do it), etc. I don't mean to couple the edit feature solely to the media browser, since it may be used for other stuff too.

justinbeaty:
The other option I thought about was an image specific bubble menu. That's part of the reason I had abstracted the toolbar code.

But it would only have two options, `Edit Image` and `Edit Alt Text`. Then:
* Do we still make double click open the media browser? 
* Do widgets have a bubble menu?
* If so, they'd only have one option, so it's pointless.
* If not, we have inconsistent UI.

I thought putting the alt text box in the modal was just easier.

fballiano:
> By default, it only allows images inside a `<div>`, `<p>`, `<td>` and other block elements, but not `<a>`, `<span>`, `<li>`, etc. However, I just realized that the Image extension has an option for this, not sure how I missed it. I pushed that so actually the homepage breaks a little less.
> 
> But, it actually also strips `<div>` elements too. I'm sure there's a good solution for it though. I guess one of the points of tiptap is that content should follow a schema, and if something isn't allowed it is removed. This should be a good thing for content writers since they can't make things look awful, but maybe it could cause problems.

mmmmm it seems a bit XHTML-ish, I don't know... I'll think about it :-)

app/design/adminhtml/default/default/template/cms/browser/tree.phtml
```diff
@@ -60,6 +60,8 @@
     });
 
     tree.expandPath('<?= $this->getTreeCurrentPath() ?>').then((node) => {
-        MediabrowserInstance.selectFolder(node);
+        MediabrowserInstance.selectFolder(node, true).then(() => {
+            MediabrowserInstance.selectFileById('<?= $this->getTreeCurrentFile() ?>');
+        });
```
justinbeaty, line 65: @fballiano this is the code that should select the correct folder and file in case you want to debug it. Could it just have been a browser cache issue?

fballiano: mmm no, also testing after a hard reload I can't make it work :-\

justinbeaty: Hm, maybe something when I did the path in session stuff. I'll definitely make it work again though.

fballiano:
this is beyond amazing 🤩🤩🤩

Let's see what to do with the div-slideshow situation and we should be good to go 🥳

justinbeaty:
> mmmmm it seems a bit XHTML-ish, I don't know... I'll think about it :-)

To be fair, block elements inside inline elements aren't valid in HTML5 either, but browsers aren't strict about it. I think there's solutions, maybe something like this which can be done with a node view extension:
```html
<div class="custom-content">
  <!-- any HTML you like can go here -->
</div>
```

Tiptap is pretty damn extensible so I do like it, but it is very different from how TinyMCE worked. But eventually with the addition of more custom widgets / blocks we get closer to a page-builder type interface.



---

In any case, I'll research a bit and see if I can find any wysiwyg that allow arbitrary HTML, but so far it seems that's just TinyMCE and CKEditor.

We could potentially support multiple wysiwygs. I don't know about adding back TinyMCE support, but if we did then it would be a much cleaner integration with all the work we've done here. I'd never install it by default or necessarily recommend it out of principle, but it could be an option.

fballiano:
> In any case, I'll research a bit and see if I can find any wysiwyg that allow arbitrary HTML, but so far it seems that's just TinyMCE and CKEditor.

those 2 are more or less the same now, owned by the same company, same licensing strategy etc...

> We could potentially support multiple wysiwygs. I don't know about adding back TinyMCE support, but if we did then it would be a much cleaner integration with all the work we've done here. I'd never install it by default or necessarily recommend it out of principle, but it could be an option.

since the GPL migration tinymce/ckeditor should be forbidden (I do love the *GPL licenses but unfortunately we can't change Maho's and thus we have to educate users that they have not to use GPL software mixed with OSL one so...)

we could pontentially support more editors but, other then quill (and we tested it already), at the moment I don't think there's anything else than TipTap ehhehe

justinbeaty:
> those 2 are more or less the same now, owned by the same company, same licensing strategy etc...

That's why I'd never recommend it out of principle. Luckily we can still criticize TinyMCE in this repo. :)

> since the GPL migration tinymce/ckeditor should be forbidden (I do love the *GPL licenses but unfortunately we can't change Maho's and thus we have to educate users that they have not to use GPL software mixed with OSL one so...)

I guess I was thinking for users who obtain a license, but yeah users would just install the GPL version anyway, so let's not even go down that path.

> we could pontentially support more editors but, other then quill (and we tested it already), at the moment I don't think there's anything else than TipTap ehhehe

Plus from what I can tell, Quill also fillers out any HTML it doesn't recognize, so same issue as tiptap but not as popular. So tiptap it is.

---

I do think we should implement the warning if the content changes due to enabling tiptap. I think the best way is probably to check anytime you switch from plain to tiptap. That handles both the initial load and if you switch to plain, add some html, and switch back to tiptap. I think it should say something like "enabling the wysiwyg will change your content [Ok] [Cancel]", or something worded better. It should probably only be like 20 lines of code, what do you think?

Actually I'd even go farther and show a diff using that new code in the admin activity module, but maybe that can wait, lol...

fballiano:
> > those 2 are more or less the same now, owned by the same company, same licensing strategy etc...
> 
> That's why I'd never recommend it out of principle. Luckily we can still criticize TinyMCE in this repo. :)

🤣
 
> Plus from what I can tell, Quill also fillers out any HTML it doesn't recognize, so same issue as tiptap but not as popular. So tiptap it is.

the more I'm reading about tiptap the more everybody loves it so I think we did the right choice here

> I do think we should implement the warning if the content changes due to enabling tiptap. I think the best way is probably to check anytime you switch from plain to tiptap. That handles both the initial load and if you switch to plain, add some html, and switch back to tiptap. I think it should say something like "enabling the wysiwyg will change your content [Ok] [Cancel]", or something worded better. It should probably only be like 20 lines of code, what do you think?

I agree

> Actually I'd even go farther and show a diff using that new code in the admin activity module, but maybe that can wait, lol...

that would be actually really useful although not necessary

------

but I still don't understand why it's removing our `div class="slideshow"` that contains the `ul`, I mean, a `ul` can be in a `div`... anyway, from https://github.com/ueberdosis/tiptap/discussions/3792 and https://tiptap.dev/docs/editor/core-concepts/schema maybe it's possible to simply extend the schema?

justinbeaty:
> but I still don't understand why it's removing our `div class="slideshow"` that contains the `ul`, I mean, a `ul` can be in a `div`... anyway, from [ueberdosis/tiptap#3792](https://github.com/ueberdosis/tiptap/discussions/3792) and https://tiptap.dev/docs/editor/core-concepts/schema maybe it's possible to simply extend the schema?

TipTap doesn't know what a `<div>` is because no registered extension has a rule to parse it. A similar example would be if you didn't add the `Link` extension, it would strip out links too. I guess it's like this because it lets the developer be really strict on what is allowed. Like they could have a document that can only contain `<p>` elements, and literally nothing else.

Making a custom `Node.create` is actually the way to extend the schema. I just pushed a new `MahoSlideshow` extension that solves this problem. The nice thing is that it also acts as a validator. So you can't have an `<li>` that contains just text, it must contain an `` optionally wrapped in an `<a>` tag. Anything else is filtered out.

Then in the wysiwyg you'll see a nice preview on how it looks like on the frontend. What will be the ultimate goal is that you can double-click it and add/delete/rearrange slides.

---

> the more I'm reading about tiptap the more everybody loves it so I think we did the right choice here

Yeah, I really like it. I had experimented with it a couple of years ago, but never got this deep into building custom extensions.

fballiano:
> Making a custom `Node.create` is actually the way to extend the schema. I just pushed a new `MahoSlideshow` extension that solves this problem. The nice thing is that it also acts as a validator. So you can't have an `<li>` that contains just text, it must contain an `` optionally wrapped in an `<a>` tag. Anything else is filtered out.

riiiiiight

question, shouldn't we allow for a div to contain anything? it wouldn't be that hard on the previously existing data maybe?

> Then in the wysiwyg you'll see a nice preview on how it looks like on the frontend. What will be the ultimate goal is that you can double-click it and add/delete/rearrange slides.

I couldn't really see it working correctly (cause I've broken my dev environment database a bit...) but what I can see seems great, can you also preview more than the first slide of the slideshow?

justinbeaty:
@fballiano I did see, was just still thinking about it. Here are some thoughts I've gathered:

We could allow a div to contain anything, but there's a few caveats:

1. We'd have to show the div with a dotted border or something, because otherwise divs are invisible.
2. Tiptap isn't going to know what those divs should look like on the frontend, so they'll all just be 100% width.
3. (Not sure about this one) It can interact oddly with paragraphs. For example if you press enter it might "escape" the div. Or if you fix that, you might not be able to escape the div when you actually want to, especially if you're in nested divs.
4. Then there's also stuff like spans, do we allow those? If so, how do you jump in or out of one?
5. It may be complicated for the user who needs the wysiwyg.

There's this quote from the link you posted:

> You'd need a node with a parseHTML that accepts divs but I would warn you that this is probably not what you actually want.
> Tiptap is not an arbitrary HTML editor do not treat it like one thinking that it will work

Which makes sense to me the more I think about it. I really can't find any pre-built extensions that do this, so there might be a reason for it. In any case, I still thought about how we can solve this, because many Tiptap use-cases might be for projects that never had content before, which is not us.

So say we do make an arbitrary content div. Then either:

*That* content also needs to fit the schema. So we still strip all unsupported tags contained within the div. At least this would still work for divs that control layout, but the more ideal solution would be to have custom extensions for creating sidebars, etc.

The div basically becomes an "embedded raw html block", and the user can't use any wysiwyg features in it. This seems to be an approach taken by some other projects. But potentially the entire content might be wrapped in a div, and then it's really confusing to the user because the wysiwyg view is basically just an HTML editor and they won't be sure why.

It's still not a bad idea to be able to embed arbitrary HTML into a page, but really we have static blocks for that.

So, I'm unsure about it. Long term I think we're better off without arbitrary HTML as long as we have enough tools for the user. But I can also see people wanting this, so what about supporting https://github.com/hugerte/hugerte ?

In any case I'll still push the "convert HTML" idea so we can see how that works.

fballiano:
mmm ok I agree with you, I'd say that a 'div' extension that allows to set a id/class and it's shown with the blue outline (kinda like page builder style) could be an interesting concept but I'm not sure if it will solve all the problems with old HTML or just a tiny bit (which would be useless than) :-\

cause the slideshow extension is great, but i'm wordering how many websites use our slideshow syntax, i'd say zero, and if they use a different one... well ok we're saying that they shouldn't be using WYSIWYG in this case right?

I didn't know about hugeRTE, maybe it would have been a quickNeasy migration 🤦🏻 but at this point I wouldn't even support it. if we support more than one RTE we'll have double the code to maintain, people wouldn't know which one to use etc.. so... TipTap scares me but it seems the best one in the open source world... and I couldn't even find a changelog of hugeRTE which is not the best...

🤔

justinbeaty:
I added this warning dialog. So if you load a page with unsupported content, tiptap is disabled by default and you get this confirm box when trying to enable it. I tried to make it work anytime you switch from the regular textarea. It's doable but would be a bit more complicated.

Technically I do get the HTML that is unsupported, but it only shows the last element and so I don't want to mislead the user into thinking that's the only thing wrong. Again with some more work I could get all of the unsupported nodes, but this works for now.

image

fballiano:
tested and works amazing, although I'm still a bit on the fence for not supporting something like this

`<div class="category-image">Hold It - Jet setting just got easier with our innovative travel gear and accessories</div>`

but the modification works perfect

justinbeaty:
> I'm still a bit on the fence for not supporting something like this

It's easy enough to support these things. We just need to identify them. Maybe there's some way we can define our own "mini-schemas" without having to make an extension for everything.

fballiano:
what do you think about this last commit (totally revertable with no problem)?

Screenshot 2025-07-01 alle 18 38 41

justinbeaty:
I changed some code style to match the other extensions, also taking advantage of `element.classList` so we don't have to split / filter / trim class names.

Then I also tried removing the transparent background in the divs, because they get progressively darker. Although maybe you intended this? I'm not sure it's totally necessary.

image


Otherwise this works really well!

justinbeaty:
> what do you think about this last commit of mine? [...] If you like it, 

Do I like it??? It's amazing! I pushed a couple of changes, hope everything is ok.

5f53ec0d060846288c33f169663e0bb85f2e0018: 
* Reused the same src/alt parsing logic from the `MahoImage` extension. 
* No need to do `renderDirectiveImageUrl(...) || src` since the `renderDirectiveImageUrl()` function already handles that.
* `dialog.returnValue` will always exist, so don't need to check for it. And if it's somehow an empty string then we should show that `console.error` anyway.

b219d145768b28130b922b01532dff833bb55d93:
* Move the `let slides` variable up in scope, that way we don't need to store it on the dialog element as it's available in all the functions.

dea98d03cde342adcc821ec67e1304d9f29f57a4:
* IMO we should just load sortable.js through layout xml. We don't load stuff like MahoTree, MahoDialog, widgets dialog, etc through JS.

cd68fa8915290f203bbb74dd8f43046ddab16b4e:
* Moved the `<li>` template up to a `slideTemplate` variable and used our icon helper.
* Attached the event listeners in the loop. That way you can just use `index` directly since JS keeps the correct values in scope.
* Just init sortable once, and ensurue we destroy it when the dialog closes.
* Added URL validation.
* Made the drag handle a bit bigger, and also removed the move cursor from the rest of the li since it wasn't draggable.

0b2e697661287c555ac746d80934469be4a41714: 
* I didn't catch it before, but if you disabled / enabled the editor, the slideshow extension didn't' recognize it.
* The MahoDiv extension must be last so that all other extensions have a chance to "claim" the `<div>`.

f7eab520c5397927d0fba348cdc382fec2aa2cc8:
* The media browser technically shouldn't have a `target_element_id` prop. That's used for inserting whatever image you select into the `<textarea>`, so each time you selected an image for the slideshow it was being appended to the textarea's content.
* It didn't really matter, because once you insert the slideshow it would sync tiptap ->textarea, but still better to fix.

app/code/core/Mage/Catalog/Helper/Output.php
```diff
@@ -115,7 +115,11 @@ public function productAttribute($product, $attributeHtml, $attributeName)
                 $attributeHtml = $this->escapeHtml($attributeHtml);
             }
             if ($attribute->getFrontendInput() == 'textarea') {
-                $attributeHtml = nl2br($attributeHtml);
+                Mage::log($attributeHtml);
```
fballiano, line 118:
```suggestion
```

fballiano:
> > what do you think about this last commit of mine? [...] If you like it,
> 
> Do I like it??? It's amazing! I pushed a couple of changes, hope everything is ok.
> 
> [5f53ec0](https://github.com/MahoCommerce/maho/commit/5f53ec0d060846288c33f169663e0bb85f2e0018):
> 
> * Reused the same src/alt parsing logic from the `MahoImage` extension.
> * No need to do `renderDirectiveImageUrl(...) || src` since the `renderDirectiveImageUrl()` function already handles that.
> * `dialog.returnValue` will always exist, so don't need to check for it. And if it's somehow an empty string then we should show that `console.error` anyway.
> 
> [b219d14](https://github.com/MahoCommerce/maho/commit/b219d145768b28130b922b01532dff833bb55d93):
> 
> * Move the `let slides` variable up in scope, that way we don't need to store it on the dialog element as it's available in all the functions.
> 
> [dea98d0](https://github.com/MahoCommerce/maho/commit/dea98d03cde342adcc821ec67e1304d9f29f57a4):
> 
> * IMO we should just load sortable.js through layout xml. We don't load stuff like MahoTree, MahoDialog, widgets dialog, etc through JS.
> 
> [cd68fa8](https://github.com/MahoCommerce/maho/commit/cd68fa8915290f203bbb74dd8f43046ddab16b4e):
> 
> * Moved the `<li>` template up to a `slideTemplate` variable and used our icon helper.
> * Attached the event listeners in the loop. That way you can just use `index` directly since JS keeps the correct values in scope.
> * Just init sortable once, and ensurue we destroy it when the dialog closes.
> * Added URL validation.
> * Made the drag handle a bit bigger, and also removed the move cursor from the rest of the li since it wasn't draggable.
> 
> [0b2e697](https://github.com/MahoCommerce/maho/commit/0b2e697661287c555ac746d80934469be4a41714):
> 
> * I didn't catch it before, but if you disabled / enabled the editor, the slideshow extension didn't' recognize it.
> * The MahoDiv extension must be last so that all other extensions have a chance to "claim" the `<div>`.
> 
> [f7eab52](https://github.com/MahoCommerce/maho/commit/f7eab520c5397927d0fba348cdc382fec2aa2cc8):
> 
> * The media browser technically shouldn't have a `target_element_id` prop. That's used for inserting whatever image you select into the `<textarea>`, so each time you selected an image for the slideshow it was being appended to the textarea's content.
> * It didn't really matter, because once you insert the slideshow it would sync tiptap ->textarea, but still better to fix.
> 
> Edit: Also I changed the `forEach` loops to `for of`. The latter is the newer syntax and can basically loop over anything. I personally think it's easier to read because it looks like any other loop (the line starts with `for` and just a regular closing bracket instead of `});`). Plus you can `continue` or `break` in it like other loops, etc. Hope that's okay.

wow :-)))))))

app/code/core/Mage/Catalog/Helper/Output.php
justinbeaty: Oops!

fballiano:
I think that when, on this foundation, we make a pagebuilder set of widgets, we may decide to separate the configurations but at the moments that's ok as it is IMHO :-)


> But I did find that `<br>` were added to the short description, even if it was already HTML. I'm not sure if this was an issue with TinyMCE too. But with just simple text it made it look really bad:

is it fix now? cause (if I understood it correctly) that's not happening to me

justinbeaty:
> I think everything is great :-)

Probably too, it's worthwhile checking each CMS page and:
* Converting the content
* Making sure it looks okay
* Updating the sample data with the new `cms_page` and `cms_block` tables.

justinbeaty:
Hey @fballiano sorry I had forgotten it was 4th of July weekend and took a few days off for family, etc. I still just want to take a look at a few things. By the end of the week for sure let's merge it, hopefully sooner. 😁

fballiano:
@justinbeaty family time is always the best time :-)

fballiano:
I totally agree, nothing is set in store here :-)

I'll merge so i'll have time to prepare the release for next week :-)))

fballiano:
I'm noticing something on demo.mahocommerce.com...

tiptap doesn't like this code:

```html
<div class="slideshow">
	<ul>
		<li><a href="{{config path="web/secure/base_url"}}accessories/eyewear.html"><img src="{{media url="wysiwyg/slide-1.webp"}}" alt="An eye for detail - Click to Shop Eye Wear" /></a></li>
		<li><a href="{{config path="web/secure/base_url"}}women.html"><img src="{{media url="wysiwyg/slide-2.webp"}}" alt="Style solutions - covet-worthy styles in travel-friendly fabrics - Click to Shop Woman" /></a></li>
		<li><a href="{{config path="web/secure/base_url"}}men.html"><img src="{{media url="wysiwyg/slide-3.webp"}}" alt="Wing man - hit the runway in stylish separates and casuals - Click to Shop Man" /></a></li>
        </ul>
</div>

<ul class="promos">
	<li>
		<a href="{{config path="web/secure/base_url"}}home-decor.html"> 
			<img src="{{media url="wysiwyg/homepage-three-column-promo-01B.webp"}}" alt="Physical &amp; Virtual Gift Cards" /> 
		</a>
	</li>
	<li>
		<a href="{{config path="web/secure/base_url"}}vip.html"> 
			<img src="{{media url="wysiwyg/homepage-three-column-promo-02.webp"}}" alt="Shop Private Sales - Members Only" />
		</a>
	</li>
	<li>
		<a href="{{config path="web/secure/base_url"}}accessories/bags-luggage.html"> 
			<img src="{{media url="wysiwyg/homepage-three-column-promo-03.webp"}}" alt="Travel Gear for Every Occasion" />
		</a>
	</li>
</ul>
<style type="text/css">
    /* Reset the .std styles */
    .std ul.products-grid li {
        list-style: none;
        margin-left: 0;
    }
    .std ul.products-grid li .product-info {
        padding-bottom: 0 !important;
        min-height: 0 !important;
    }
    .widget-new-products h2 {
        text-align: center;
    }
</style>
{{widget type="catalog/product_widget_new" display_type="new_products" products_count="5" template="catalog/product/widget/new/content/new_grid.phtml"}}
```

and it converts it to:

```
<div class="slideshow">
    <ul>
        <li><a href="{{config path="web/secure/base_url"}}accessories/eyewear.html"><img src="{{media url="wysiwyg/slide-1.webp"}}" alt="An eye for detail - Click to Shop Eye Wear"></a></li>
        <li><a href="{{config path="web/secure/base_url"}}women.html"><img src="{{media url="wysiwyg/slide-2.webp"}}" alt="Style solutions - covet-worthy styles in travel-friendly fabrics - Click to Shop Woman"></a></li>
        <li><a href="{{config path="web/secure/base_url"}}men.html"><img src="{{media url="wysiwyg/slide-3.webp"}}" alt="Wing man - hit the runway in stylish separates and casuals - Click to Shop Man"></a></li>
    </ul>
</div>
<ul>
    <li>
        <p><a href="{{config path="web/secure/base_url"}}home-decor.html"><img src="{{media url="wysiwyg/homepage-three-column-promo-01B.webp"}}" alt="Physical &amp; Virtual Gift Cards"></a></p>
    </li>
    <li>
        <p><a href="{{config path="web/secure/base_url"}}vip.html"><img src="{{media url="wysiwyg/homepage-three-column-promo-02.webp"}}" alt="Shop Private Sales - Members Only"></a></p>
    </li>
    <li>
        <p><a href="{{config path="web/secure/base_url"}}accessories/bags-luggage.html"><img src="{{media url="wysiwyg/homepage-three-column-promo-03.webp"}}" alt="Travel Gear for Every Occasion"></a></p>
    </li>
</ul>
<p>{{widget type="catalog/product_widget_new" display_type="new_products" products_count="5" template="catalog/product/widget/new/content/new_grid.phtml"}}</p>
```

it removes the CSS class from the `ul` and adds the `<p>` to the widget... mmmmmmmmmmm I'm not sure this is ok...

PR #180, opened 2025-05-30
Added admin interface for managing countries, regions, and their translations

fballiano:
Credit: the inspiration from this PR comes from a discussion in OpenMage discussion area and I think it's a pretty useful feature that should have always been present. The code of this PR is 95% generated by AI with some minor adjustments by me.

**Summary**
  • Implement comprehensive admin interface for managing countries, regions, and region names
  • Add full CRUD operations with grid views, edit forms, and mass actions for directory data
  • Integrate directory management into admin navigation under System menu


fballiano:
as far as I could test it should be alright

justinbeaty:
Testing this one now, but am running into a problem creating a new Country. Investigating what could be going wrong.

justinbeaty:
I have a really nice update for this, just forgot I had something to attend today. Will push first thing tomorrow, then continue with reviewing other PRs.

justinbeaty:
Some questions:

1\. I didn't end up requiring the ISO2 and ISO3 code. I'm not sure how much it's required, but I guess let's just allow it to be blank.

2\. Do we want to still block deleting a country / region that has translations? There's an `ON DELETE CASCADE` on the names tables, so we wouldn't have orphaned records.

fballiano:
> I ended up implementing this... 😅 It's one less dependence on Zend_Locale now, which is good. You can add translations for both countries and regions right from the edit page. I realized this was pretty much exactly like the Cart Promo / Coupon grid, which I had just worked on.

it's great! and one less dependency on zend is surely the way to go :-)

the interface works great :-)

> One thing to note, however. Before Zend provided country translations (but not region translations). However, I only added to the database the `en_US` country names. This means that countries are no longer translated by default.
> The solution to this is that we should allow importing of [Unicode CLDR files](https://cldr.unicode.org/), which are the same files Zend used. Alternatively we could include these in the Maho language packs, if we have some way to run sql setup script there.

I think it's a tradeoff we can handle and a language pack is a stardard Maho module so it should be able to add that, i'd use it as a fallback if there's no db translation but I wouldn't do that in this PR anyway

> Either way, I still think it's better to have this functionality, because otherwise adding a new country didn't really mean much if you couldn't add a name as well.

absolutely

> 1. I didn't end up requiring the ISO2 and ISO3 code. I'm not sure how much it's required, but I guess let's just allow it to be blank.

AFAIK they're used only in some shipping method, but anyway, since they're nullable on the DB, it's fine that they're nullable in the backend

> 2. Do we want to still block deleting a country / region that has translations? There's an ON DELETE CASCADE on the names tables, so we wouldn't have orphaned records.

It's totally fine as you developed, the important thing is not to leave orphaned records, the rest is totally fine.

❤️

justinbeaty:
Oh, shoot. Glad you caught this.

luigifab:
I also think that this PR has destroyed install wizard: http.../maho-git/public/index.php/install/wizard/locale/

```
Fatal error:  Uncaught Error: Call to a member function getTableName() on false in maho-git/app/code/core/Mage/Core/Model/Resource.php:274
Stack trace:
#0 maho-git/app/code/core/Mage/Core/Model/Resource/Db/Abstract.php(253): Mage_Core_Model_Resource->getTableName()
#1 maho-git/app/code/core/Mage/Directory/Model/Resource/Country.php(23): Mage_Core_Model_Resource_Db_Abstract->getTable()
#2 maho-git/app/code/core/Mage/Core/Model/Resource/Abstract.php(23): Mage_Directory_Model_Resource_Country->_construct()
#3 maho-git/app/code/core/Mage/Core/Model/Config.php(1478): Mage_Core_Model_Resource_Abstract->__construct()
#4 maho-git/app/Mage.php(417): Mage_Core_Model_Config->getResourceModelInstance()
#5 maho-git/app/code/core/Mage/Core/Model/Resource/Db/Collection/Abstract.php(451): Mage::getResourceModel()
#6 maho-git/app/code/core/Mage/Core/Model/Resource/Db/Collection/Abstract.php(464): Mage_Core_Model_Resource_Db_Collection_Abstract->getResource()
#7 maho-git/app/code/core/Mage/Directory/Model/Resource/Country/Collection.php(30): Mage_Core_Model_Resource_Db_Collection_Abstract->getTable()
#8 maho-git/app/code/core/Mage/Core/Model/Resource/Db/Collection/Abstract.php(110): Mage_Directory_Model_Resource_Country_Collection->_construct()
#9 maho-git/app/code/core/Mage/Core/Model/Config.php(1478): Mage_Core_Model_Resource_Db_Collection_Abstract->__construct()
#10 maho-git/app/Mage.php(417): Mage_Core_Model_Config->getResourceModelInstance()
#11 maho-git/app/code/core/Mage/Core/Model/Locale.php(871): Mage::getResourceModel()
#12 maho-git/app/code/core/Mage/Core/Model/Locale.php(222): Mage_Core_Model_Locale->getCountryTranslationList()
```

PR #181, opened 2025-05-31
Fixed shipping methods configuration to handle array method titles

fballiano:
- add a cart promo rule
- in the conditions try to select "shipping method"
- you'll get this

Screenshot 2025-05-31 alle 13 45 37


justinbeaty (approved):
I was reviewing this, but I got sidetrack realizing the field dependency code was broken here. I have a PR to fix that tomorrow, but this code itself looks good.

fballiano:
> Do you mean the array to string conversion?

yep

> I'm not sure, but I'd assume it's probably present on OM too?

no idea actually

PR #184, opened 2025-06-02
Replaced session management with `symfony/http-foundation`

fballiano:
This implementation
- supports Redis out-of-the-box, no 3rd party modules needed
- removes database based session (IMHO an outdated way of managing sessions)
- defaults to file based session
- migrates all the code in `Mage_Core_Model_Session_Abstract_Varien` to `Mage_Core_Model_Session_Abstract`
- removes `VALIDATOR_SESSION_EXPIRE_TIMESTAMP`, `VALIDATOR_SESSION_RENEW_TIMESTAMP` and `VALIDATOR_SESSION_LIFETIME` that are now managed by symfony
- adds a session cleaning cron (running daily at 3.30) to remove expired session files from filesystem (for redis it's not needed)

To activate redis, add this to your `app/etc/local.xml`:

```xml
<session_save>redis</session_save>
<redis_session>
    <dsn>redis://localhost/9</dsn>
</redis_session>
```

### DIFF
--- a/app/code/core/Mage/Checkout/Model/Session.php
+++ b/app/code/core/Mage/Checkout/Model/Session.php
@@ -6,7 +6,7 @@
  * @package    Mage_Checkout
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright  Copyright (c) 2019-2024 The OpenMage Contributors (https://openmage.org)
- * @copyright  Copyright (c) 2024 Maho (https://mahocommerce.com)
+ * @copyright  Copyright (c) 2024-2025 Maho (https://mahocommerce.com)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
--- a/app/code/core/Mage/Customer/Model/Session.php
+++ b/app/code/core/Mage/Customer/Model/Session.php
@@ -6,7 +6,7 @@
  * @package    Mage_Customer
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright  Copyright (c) 2019-2024 The OpenMage Contributors (https://openmage.org)
- * @copyright  Copyright (c) 2024 Maho (https://mahocommerce.com)
+ * @copyright  Copyright (c) 2024-2025 Maho (https://mahocommerce.com)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
--- a/app/code/core/Mage/Newsletter/Model/Session.php
+++ b/app/code/core/Mage/Newsletter/Model/Session.php
@@ -6,7 +6,7 @@
  * @package    Mage_Newsletter
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright  Copyright (c) 2020-2024 The OpenMage Contributors (https://openmage.org)
- * @copyright  Copyright (c) 2024 Maho (https://mahocommerce.com)
+ * @copyright  Copyright (c) 2024-2025 Maho (https://mahocommerce.com)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 


fballiano:
@justinbeaty I'd really like to merge this one

justinbeaty:
> @justinbeaty I'd really like to merge this one

Hey, sorry I was offline all weekend and getting caught up on other stuff yesterday. Let me take a look at this quick.

app/code/core/Mage/Core/Model/Session/Abstract.php
```diff
@@ -18,9 +25,20 @@
...
+class Mage_Core_Model_Session_Abstract extends Varien_Object
 {
+    public const VALIDATOR_KEY                         = '_session_validator_data';
+    public const VALIDATOR_HTTP_USER_AGENT_KEY         = 'http_user_agent';
+    public const VALIDATOR_HTTP_X_FORVARDED_FOR_KEY    = 'http_x_forwarded_for';
+    public const VALIDATOR_HTTP_VIA_KEY                = 'http_via';
+    public const VALIDATOR_REMOTE_ADDR_KEY             = 'remote_addr';
+    public const VALIDATOR_PASSWORD_CREATE_TIMESTAMP   = 'password_create_timestamp';
+    public const SECURE_COOKIE_CHECK_KEY               = '_secure_cookie_check';
+    public const REGISTRY_CONCURRENCY_ERROR            = 'concurrent_connections_exceeded';
```
fballiano, line 40:
```suggestion
```

@justinbeaty this too, right?

fballiano:
> Hmm, I think there's one weird thing here. We are storing the symfony session object as `$this->symfonySession`, but since there's always multiple session objects (core, adminhtml, etc), then `$this->symfonySession` is actually only stored on the core instance.

mmm but both frontend/backend work with this implementation

> But they do recommend against using the `$_SESSION` super global, and I don't things like the Flash Messages (which are super nice) would work this way.
> 
> I think we need to:
> 
> 1. Store the symfony session in registry.

I'll try to do these

> 2. We should probably write custom `getData()` and `setData()` methods to take advantage of reading from the symfony session object directly instead of using the Varien_Object's methods on `$_SESSION`.

could this be postponed?

fballiano:
were you thinking something like this? https://github.com/MahoCommerce/maho/pull/184/commits/010fd01860c431e52266243233d92124621b4841

fballiano:
done, but on the latest version

fballiano:
if I try to completely remove $_SESSION I can't make it work :-\

justinbeaty:
Also, I removed the legacy fallback stuff like `session_name()`, `session_id()`, etc. If Symfony session were to fail in any case, I don't think we want to be returning anything, but I should double check the API classes.

fballiano:
please fill free to push it here, I had to stop cause I was super tired tonight 😅

fballiano:
never worry about that :-)

justinbeaty:
Pushed some changes. I haven't 100% tested everything, but it should be okay. Will finish testing tomorrow morning then we should merge. I love replacing this old stuff with Symfony components. :)

A few notes:

1\. I also seem to be missing the `password_created_at` field like in #187. I'm guessing this probably is a sample data problem where that column is missing from the sql dump. Because of this, our invalidate session upon password change isn't working, so I can't test to make sure that's working.

2\. I'm fairly certain that if we did have the `password_created_at` field, users would have been logged out upon changing their password. That is because of what I wrote here: https://github.com/MahoCommerce/maho/pull/184#issuecomment-2981905210. Basically the session's timestamp wouldn't be updated, and it would then be older than the password.

3\. I re-added the `useValidateSessionPasswordTimestamp()` method. I know it just always returns true, but classes that extend from this may turn that off if they need to not check this. Like some sort of API session that doesn't use passwords and so sessions shouldn't expire.

4\. Found one potentially nasty bug. Explained in the commit message but will copy paste here:
```
PHP's native session garbage collection never affected Magento because Mage stored
sessions in a place that PHP did not know about. This is why /path/to/mage/var/session
grows to infinity. However NativeFileSessionHandler kindly tells PHP about our custom
path, and so PHP's gc will activate. That means that whatever gc_maxlifetime is set
in php.ini will be when sessions are destroyed. Debian based distros have PHP's native
gc disabled, but other distros have a short default lifetime, around 24 minutes.

We could avoid having a cronjob do this for us and just configure this manually, but
cron is a better solution since no one user's request is the one that causes the gc.
```

5\. I forgot to have checked the SOAP access. Putting a note here so I remember tomorrow.

app/code/core/Mage/Core/Model/Session/Abstract.php
```diff
@@ -38,6 +57,16 @@ class Mage_Core_Model_Session_Abstract extends Mage_Core_Model_Session_Abstract_
...
 
+    /** @var bool Flag true if session validator data has already been evaluated */
+    protected static $isValidated = false;
+
+    /**
+     * Map of session enabled hosts
+     * @example array('host.name' => true)
+     * @var array
+     */
+    protected $_sessionHosts = [];
```
fballiano, line 68:
```suggestion
     */
    protected array $_sessionHosts = [];
```

app/code/core/Mage/Core/Model/Session/Abstract.php
```diff
@@ -18,9 +25,21 @@
  * @method $this setSuccessMessage(string $value)
  * @method $this unsSuccessMessage()
  * @method $this setMessages(Mage_Core_Model_Abstract|Mage_Core_Model_Message_Collection $value)
+ * @method bool|null getSkipEmptySessionCheck()
+ * @method $this setSkipEmptySessionCheck(bool $flag)
  */
-class Mage_Core_Model_Session_Abstract extends Mage_Core_Model_Session_Abstract_Varien
+class Mage_Core_Model_Session_Abstract extends Varien_Object
 {
+    public const REGISTRY_KEY                          = 'symfony_session';
```
fballiano, line 33:
```suggestion
    public const string REGISTRY_KEY                    = 'symfony_session';
```

app/code/core/Mage/Core/Model/Session/Abstract.php
```diff
@@ -60,17 +89,331 @@ class Mage_Core_Model_Session_Abstract extends Mage_Core_Model_Session_Abstract_
...
+
+        return $this;
+    }
+
+    /**
+     * Retrieve cookie object
+     *
+     * @return Mage_Core_Model_Cookie
+     */
+    public function getCookie()
```
fballiano, line 303:
```suggestion
     */
    public function getCookie(): Mage_Core_Model_Cookie
```

fballiano:
actually, i'm wondering, can we add strong types to everything here? since we don't need 3rd party modules for redis etc maybe we're not really breaking compatibility?

fballiano:
> > mmm it could be a problem with an older version of the sample data but I'll re-test it
> 
> It's possible it's fixed in later versions of sample data. Although my guess is that somehow the column isn't there in the sql dump, and it won't be created because `core_resource` table already shows the install / upgrade script ran.

sample data doesn't work like that anymore: https://github.com/MahoCommerce/maho-sample-data it's not a complete dump anymore :-)

justinbeaty:
> @justinbeaty I've added some strong typing, fixing inheriting classes here and there and everything seems ok. I'd merge it

I'll just test SOAP real quick, got sidetracked with some other stuff yesterday.

justinbeaty (approved):
Tested Soap v1, Soap v2, and Soap v2 WS-I.

PR #185, opened 2025-06-02
Fixed gift message bug where quote and item IDs could collide during checkout

### LINKED ISSUE #73: Giftmessage bug in checkout
https://github.com/MahoCommerce/maho/pull/71#issuecomment-2546903077

fballiano:
Fixes #73, documented in https://github.com/MahoCommerce/maho/pull/71#issuecomment-2546903077

### DIFF
--- a/app/code/core/Mage/GiftMessage/Model/Observer.php
+++ b/app/code/core/Mage/GiftMessage/Model/Observer.php
@@ -6,7 +6,7 @@
  * @package    Mage_GiftMessage
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright  Copyright (c) 2019-2024 The OpenMage Contributors (https://openmage.org)
- * @copyright  Copyright (c) 2024 Maho (https://mahocommerce.com)
+ * @copyright  Copyright (c) 2024-2025 Maho (https://mahocommerce.com)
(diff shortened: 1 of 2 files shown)


app/code/core/Mage/GiftMessage/Model/Observer.php
```diff
@@ -87,9 +87,24 @@ public function checkoutEventCreateGiftMessage(Varien_Event_Observer $observer)
...
+                } elseif (str_starts_with($key, 'item_')) {
+                    $entityId = substr($key, 5); // Remove 'item_' prefix
+                } elseif (str_starts_with($key, 'addr_')) {
+                    $entityId = substr($key, 5); // Remove 'addr_' prefix
+                } elseif (str_starts_with($key, 'addritem_')) {
+                    $entityId = substr($key, 9); // Remove 'addritem_' prefix
+                } else {
+                    // Fallback for backward compatibility
+                    $entityId = $key;
+                }
```
justinbeaty, line 106: Since we don't care about which prefix we actually have, how about this?

```suggestion
                // Extract the actual entity ID from the prefixed key
                if (str_contains($key, '_')) {
                    $entityId = substr($key, strrpos($key, '_') + 1);
                } else {
                    // Fallback for backward compatibility
                    $entityId = $key;
                }
```

justinbeaty:
Thank you for not forgetting about this :)

I've tested both the original and the suggested version, and both work perfectly!

PR #186, opened 2025-06-03
Set PHP 8.3 as new minimum version

fballiano:
Development of symfony8 has started, and will require php 8.4, I think it could be time for us to bump our PHP minimum version to 8.3.

This would allow us to correctly require rector and simplify its workflow (cause we were using it with php8.3 directives on php8.2)


justinbeaty (approved):
This seems good to me. :)

PR #188, opened 2025-06-04
Moved category title just below breadcrumbs for better spacing

fballiano:
This PR:
- adds a new `title` block (actually 2, a generic one and a category-specific one)
- puts it right below the breadcrumbs
- removes the title from main-content area
- adds `2columns-left` as default layout for category pages

This makes the design much more modern and cleaner

Screenshot 2025-06-04 alle 15 55 38


justinbeaty:
This looks nice, only question is regarding setting `2columns-left` as default. From my memory, we can have categories that don't have layered navigation because none of the products in the category have attributes that are used in layered nav. Also maybe the search results page, but it doesn't look like you changed that one.

fballiano:
let's remove that part at the moment, the idea was to remove the 3columns layout that it's super ugly in 2025, but it's worth more exploring, just a minute and I'll fix this :-)

fballiano:
done :-)

app/code/core/Mage/Core/Block/Page/Title.php
```diff
@@ -0,0 +1,25 @@
...
+ * Maho
+ *
+ * @package     Mage_Core
+ * @copyright   Copyright (c) 2025 Maho (https://mahocommerce.com)
+ * @license     https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+class Mage_Core_Block_Page_Title extends Mage_Core_Block_Template
+{
+    protected ?string $_title = null;
```
justinbeaty, line 13: Should we ditch the underscore?

app/design/frontend/base/default/template/catalog/category/title.phtml
```diff
@@ -0,0 +1,22 @@
...
+ * @copyright   Copyright (c) 2024-2025 Maho (https://mahocommerce.com)
+ * @license     https://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
+ */
+
+/** @var Mage_Catalog_Block_Category_Title $this */
+?>
+<?php if ($title = $this->getTitle()): ?>
+<div class="page-title category-title">
+    <?php if ($this->getShowRssLink()): ?>
+        <a href="<?= $this->escapeHtml($this->buildRssLink()) ?>" class="link-rss"><?= $this->escapeHtml($this->getRssText()) ?></a>
```
justinbeaty, line 18: `Mage_Core_Model_Url::getUrl()` already escapes, so not needed. :)

```suggestion
        <a href="<?= $this->buildRssLink() ?>" class="link-rss"><?= $this->escapeHtml($this->getRssText()) ?></a>
```

fballiano: wow, what an eye!

justinbeaty:
I pushed a test commit with some changes:

1\. The templates `1column.phtml` (etc) directly call `<?= $this->getChildHtml('category.title') ?>`, but I don't think this makes sense because we have a base title block. So I added the base title block to all pages, and overwrote it in `catalog.xml`. It's only going to show up on catalog pages anyway, because no other page calls `setTitle()`.

2\. I also moved the block from `Mage_Core` to `Mage_Page`, since that's where all other stuff like breadcrumbs are.

Does that make sense? It was easier to push a test commit rather than explain, so feel free to modify it as you see fit.

However, I have one other suggestion before merging. I don't really like that the RSS stuff is duplicated in both `Mage_Catalog_Block_Category_View` and `Mage_Catalog_Block_Category_Title`.  I think we could get rid of the dedicated category title block. I can explain all tomorrow.

fballiano:
> I pushed a test commit with some changes:
> 
> 1. The templates `1column.phtml` (etc) directly call `<?= $this->getChildHtml('category.title') ?>`, but I don't think this makes sense because we have a base title block. So I added the base title block to all pages, and overwrote it in `catalog.xml`. It's only going to show up on catalog pages anyway, because no other page calls `setTitle()`.
> 
> 2. I also moved the block from `Mage_Core` to `Mage_Page`, since that's where all other stuff like breadcrumbs are.
> 
> Does that make sense? It was easier to push a test commit rather than explain, so feel free to modify it as you see fit.

everything makes much more sense :-)

justinbeaty:
I pushed a commit to avoid needing a dedicated Category title block + phtml file.  Here's why I think it's the correct approach:

* We won't need to extend the base Title block for each type of page unless there's a very specific need.
* You can set the title and links in the same way you set the title and links for the head block.
* You don't need to duplicate logic for creating RSS links, etc.
* Full support for multiple links using the `page/template_links` block.

I also used flexbox to move the RSS link over to the side, which I think is probably a bit cleaner:
image

@fballiano what do you think about this approach?

fballiano:
> * We won't need to extend the base Title block for each type of page unless there's a very specific need.

super!

> * You can set the title and links in the same way you set the title and links for the head block.

it's great, but I think we need to separate link content, link title and add a link css class (maybe), cause I'd like to use an icon instead of the "subscribe to RSS" test, I think it would look much nicer.

is that ok?

justinbeaty:
Actually, your approach is much better. :)

justinbeaty:
> actually I think it's too big but.. minimum touch target size if 24px so we're kinda stuck with it I think.

Yeah, it's fine. I guess can't get around it and keep lighthouse score.

Also as a test, I added a title and link to the homepage via layout.xml. Nice to have the flexibility. :)

```xml
    <cms_index_index translate="label">
        <label>CMS Home Page</label>
        <reference name="title">
            <action method="setTitle"><value>My Awesome Homepage</value></action>
        </reference>
        <reference name="title.links">
            <action method="addLink" translate="label title" module="cms">
                <label>Some Link</label>
                <url>https://google.com</url>
            </action>
        </reference>
    </cms_index_index>
```

justinbeaty:
> yeah but it's mainly for the new accessibility act that I still have to make improvements on

Oh right. Would be huge to have out of the box compliance there.

fballiano:
> > yeah but it's mainly for the new accessibility act that I still have to make improvements on
> 
> Oh right. Would be huge to have out of the box compliance there.

there are not many problems, it's just 3-4 things (at least for home, category, product page)

PS: I did a minor css adjustment for mobile

fballiano:
> PS: I did a minor css adjustment for mobile

which don't work well, checking it out

justinbeaty:
> > PS: I did a minor css adjustment for mobile
> 
> which don't work well, checking it out

Could maybe remove the margin-bottom & padding-bottom from the h1 / h2, then:

```css
.page-title {
    display: flex;
    gap: 0.5rem;
    justify-content: space-between;
    align-items: center; // new line
    margin-bottom: 10px; // new line
}
```

But I haven't tested it.

fballiano:
the thing is, when there's the layered, there's too much space between title and layered. but if there's no layered it's too little 😂

like this one
Screenshot 2025-06-06 alle 14 28 00

fballiano:
it' seems perfect now :-)

justinbeaty:
> niceeee thanks! chrome didn't highlight that margin for some reason

Yeah, I run into this all the time. Chrome doesn't seem to extend the highlight on an outer div when some deeper inner div has a margin that extends beyond that parent. I seem to run into it with Magento the most.... 😅

fballiano:
I'd merge it now

PR #189, opened 2025-06-04
Added configurable trailing slash behavior and remove default .html URL suffixes

fballiano:
## Summary

This PR addresses a long-running SEO problem in Magento 1-based projects where URLs can be accessed both with and without trailing slashes (e.g., `example.com/category` and `example.com/category/`), effectively duplicating content in search engines.

This PR introduces configurable trailing slash behavior for URLs and modernizes the default URL structure by removing `.html` suffixes from products and categories.

### 1. Configurable Trailing Slash Behavior
  Added new system configuration in **System > Configuration > Web > URL Options**:
  - **"Redirect to URL with trailing slash"** (default) - Enforces trailing slashes via 301 redirects
  - **"Redirect to URL without trailing slash"** - Removes trailing slashes via 301 redirects
  - **"Do nothing"** - Allows both URL formats (not recommended for SEO)

  **Default behavior:** Add trailing slashes to all URLs

  ### 2. Removed Default URL Suffixes
  - Changed `product_url_suffix` from `.html` to empty string
  - Changed `category_url_suffix` from `.html` to empty string
  - Results in modern URLs: `/category/product/` instead of `/category.html/product.html`

  ### 3. Smart Exclusions
  The trailing slash behavior does NOT apply to:
  - Admin panel URLs (no SEO benefit needed)
  - POST requests (protects forms and checkout)
  - URLs with file extensions (`.xml`, `.json`, etc.)

### URL Generation Enhancement
```php
// All Magento-generated URLs now respect the trailing slash setting
$url = Mage::helper('core/url')->addOrRemoveTrailingSlash($url);
```

### New Helper Methods
**PHP:**
```php
// Safe URL manipulation
Mage::helper('core/url')->addRouteParam($baseUrl, 'id', 1);
Mage::helper('core/url')->setRouteParams($url, ['id' => 1, 'page' => 2]);
```

**JavaScript:**
```js
// Equivalent helpers for frontend
setRouteParams(url, {id: 1, page: 2});
setQueryParams(url, {filter: 'active'});
```

### Potential Breaking Changes
Code that manually concatenates URLs may need updates:

**❌ Problematic:**
```php
$baseUrl = Mage::getUrl('customer/account');
$withParams = $baseUrl . 'id/1/'; // May create: /customer/accountid/1/
```

**✅ Recommended:**
```php
$withParams = Mage::helper('core/url')->addRouteParam($baseUrl, 'id', 1);
// Or ensure proper concatenation:
$withParams = rtrim($baseUrl, '/') . '/id/1/';
```

### CMS Block Considerations
Template URLs should use proper helpers:
```html
<\!-- ❌ Avoid -->
<a href="{{store url=""}}privacy-policy/">Manual URL</a>

<\!-- ✅ Recommended -->
<a href="{{store direct_url="privacy-policy"}}">Helper URL</a>
```

## SEO Benefits

- **Eliminates duplicate content** issues from trailing slash variations
- **Consistent canonical URLs** improve search engine indexing
- **301 redirects preserve** link equity and rankings
- **Configurable behavior** accommodates different SEO strategies

---

🚨 **Important:** Test thoroughly in staging before deploying to production, especially if you have existing server-level redirect rules that might conflict.


justinbeaty:
> should we have a select like -> redirect to "/", redirect without "/"? a "do nothing" option too?

I think this would be ideal.

> I'd actually activate by default the "remove trailing slash"

I'd agree mainly because the default category suffix is set to `.html`, and having something like `https://example.com/men.html/` wouldn't make sense.

It could cause problems if a store owner upgrades their live site without testing, but that's obviously not recommended, so I won't worry there.

justinbeaty:
I found (and fixed) a couple of problems:


1\. By default, Magento constructs URLs *with* trailing slashes. So, `/cart/`, `/customer/account/`, etc. With remove trailing slashes mode, this causes a 301 redirect on almost every page, which isn't good since it adds a round trip to the TTFB for each non cached page. That could be over 100ms depending on server speed.

To fix, I added the same logic in `Mage_Core_Model_Url::getUrl()`. So all URLs generated with that helper class will correctly have or not trailing slashes depending on your preference.

**Note:** some CMS pages / blocks don't use URL blocks correctly, but we can't really do anything about that.
```html
<a href="{{store url=""}}privacy-policy-cookie-restriction-mode/">WRONG</a>

<a href="{{store direct_url="privacy-policy-cookie-restriction-mode"}}">CORRECT</a>
```

---

2\. POST actions to a redirected URL were broken, since 301s don't forward POST data. I was getting errors during checkout:

image

image

To fix, I checked to see if we have a POST request, which is the same way that the `web/url/redirect_to_base` option checks this.

**Note:** the better solution would be to use 307/308 instead of 301/302, but I didn't make that change here. It's probably better to change all 301 / 302 -> 307 / 308 throughout the codebase in one PR.

---

### Plus a few other changes:

I moved the redirect code into `Mage_Core_Controller_Varien_Front`. I think it makes more sense here since `Mage_Core_Controller_Request_Http` is more about parsing the incoming request, as opposed to modifying the response.

I moved the config option from `Catalog / Search Engine Optimizations` to `Web / Url Options` because it affects more than just catalog pages, and we have the similar `Auto-redirect to Base URL` config there:

image

---

### Potential issue:

Because of the change to `Mage_Core_Model_Url::getUrl()`, if we have code like the following, it's going to break if trailing slashes are set to remove.

```php
$baseUrl = Mage::getUrl('customer/account');

// Breaks: https://example.com/customer/accountid/1/
$withSomeParams = $baseUrl . 'id/1/';

// Okay: https://example.com/customer/account/id/1/
$withSomeParams = rtrim($baseUrl, '/') . '/id/1/';

// Best: https://example.com/customer/account/id/1
$withSomeParams = Mage::helper('core/url')->addRouteParam($baseUrl, 'id', 1);
```

I removed the redirects from adminhtml (since there's no SEO issues there), so we only need to worry about this on the frontend. I moved my JS helper functions from `mage/adminhtml/tools.js` to `varien/js.js`, and wrote equivalent ones in PHP.

So I would say the changes to `Mage_Core_Model_Url::getUrl()` are the riskiest, but I think it's necessary to avoid a 301 on every page.

app/code/core/Mage/Core/Helper/Url.php
```diff
@@ -146,6 +210,73 @@ public function removeRequestParam($url, $paramKey, $caseSensitive = false)
...
+    public function addTrailingSlash(string $url): string
+    {
+        $path = parse_url($url, PHP_URL_PATH) ?? '/';
+
+        // Do nothing for url without path
+        if (strlen($path) === 1) {
+            return $url;
+        }
+
+        // Do nothing for pages an extension, such as .html
```
fballiano, line 225:
```suggestion
        // Do nothing for pages with an extension, such as .html
```

fballiano:
@justinbeaty I was a bit worried about breaking things like `$baseUrl . 'id/1/'`, but looking at this PR as a whole I think it makes sense, if the trailing slashes are removed... that has to break...

should we merge it?

justinbeaty:
@fballiano the only way around it would be to not modify the getUrl function, and instead regex the whole HTML to remove slashes in <a> tags, but I don't really like that solution.

I haven't actually checked in frontend blocks / controllers / JS to see if there's any code like this. I'm guessing the most likely place would be in some JS since there was never a setRouteParams helper function there until now.

What do you think about making the default either do nothing, or add trailing slash? I don't say that because trailing slashes are my preference, only because that mode is way less likely to break something. Or if you want remove trailing slash to be default, maybe make the setting do nothing for now, and set it to remove on our local instances / demo site. Then if nothing broken is found, enable it before the next release or something?

justinbeaty:
There's also this JS that could break, because the url might be `/id/foo/`, `/id/foo`, or `/id/foo?q=something`.

```js
url = baseUrl.replace(/\/id\/\w+\//, '/id/theNewId/');
```

I ran into this kind of code more than a few times in adminhtml, which was why I wrote the setRouteParams function in the first place.

fballiano:
It seems there's a problem with this version. I've set it to "add trailing slash" but I can navigate to both http://maho.test/accessories/shoes.html and http://maho.test/accessories/shoes.html/ without being redirected

justinbeaty:
> adding `/` unless there's an extension? but it's a mixed solution that feels weird :-((

Either way we have mixed URLs:

```
// With trailing slash
https://example.com/customer/account/
https://example.com/accessories.html
```

```
// Without trailing slash
https://example.com/customer/account
https://example.com/accessories.html
```

Personally, I think the without trailing is more strange, because some URLs have a file suffix and some don't. With the trailing slashes, `/customer/account/` implies `/customer/account/index.html`.

However, why don't we set `Product / Category URL Suffix` to blank, so that all URLs will either case match? This is what I always did.

> It seems there's a problem with this version. I've set it to "add trailing slash" but I can navigate to both http://maho.test/accessories/shoes.html and http://maho.test/accessories/shoes.html/ without being redirected

Yeah, I can confirm, and will fix. Even if you have enable trailing slash, it should redirect you to `accessories/shoes.html` because `accessories/shoes.html/` makes no sense.

justinbeaty:
Fixed that issue, but I found a new one that exists on main branch, the original PR, and current:

When you have add trailing slash, you can just keep adding slashes and it loads. Although that could just be something to do with nginx.

```
https://store.example.com/about-maho-demo-store////////
```

fballiano:
> When you have add trailing slash, you can just keep adding slashes and it loads. Although that could just be something to do with nginx.

I remember to test this but I could be wrong

fballiano:
> Personally, I think the without trailing is more strange, because some URLs have a file suffix and some don't. With the trailing slashes, `/customer/account/` implies `/customer/account/index.html`.

I see your point as `/` having the same function as `.html` and it's also true that `.html`, nowadays, kinda doesn't make sense anymore (but it's still the default for Maho)

> ```
> // Without trailing slash
> https://example.com/customer/account
> https://example.com/accessories
> ```

somehow to me this always looked cleaner and the final slash always seemed superfluous

at this point I'm confused and it seems we're creating more problems that the ones we solve :-\

fballiano:
if we just had a canonical tag in the page, the whole thing whouldn't have been such a big problem :-\

justinbeaty:
> somehow to me this always looked cleaner and the final slash always seemed superfluous

Yeah, that part is just down to personal preference, and if we have an option for that, everyone is happy. :)

The main problem I've always had with Mage is that only some pages have `.html`. Without it, both options are great!

```
// Both GOOD!

// With trailing slash
https://example.com/customer/account/
https://example.com/accessories/

// Without trailing slash
https://example.com/customer/account
https://example.com/accessories
```

---

> it's also true that .html, nowadays, kinda doesn't make sense anymore (but it's still the default for Maho)

Yes, but no reason we can't just set that to blank, right? It wouldn't affect existing stores as long as they have whatever value already saved in the DB.

---

> at this point I'm confused and it seems we're creating more problems that the ones we solve :-\

But it's worth solving because it's annoyed thousands of developers... 😅  I'm pretty sure the most recent version should work perfectly. 

When I first added the `Mage::helper('core/url')->[add|remove]TrailingSlashes()` methods, I didn't have the `buildUrl` method so I had used string operations. But clearly that didn't work in all cases. The code should be 1,000 times more readable now.

fballiano:
if for categories I'm ok with `https://example.com/accessories/` (because to my old brain a category it's a folder) it's the product page that hurts me a bit with the trailing slash (Zalando for example uses trailing slash for categories, .html for product pages... weird eh?).

but anyway it seems that
- remove .html suffix from category/product
- add trailing slash
- I'd also remove "Use Categories Path for Product URLs" cause it can create duplicate content if a product is in more than one category (and that's another mess with the breadcrumbs...)

what do you think?

fballiano:
> The only reason I suggested making trailing slash the default is because it shouldn't break anything.

and you're totally right, I think it makes sense for it to be the default

> > * I'd also remove "Use Categories Path for Product URLs" cause it can create duplicate content if a product is in more than one category (and that's another mess with the breadcrumbs...)
> 
> Ideally this is the problem we solve with canonical tags. The category name in the URL is better both for SEO (since it has more keywords), and for humans (for the same reason):

yep I know but I knew that in the past years this changed and there was always the debase of "first level folder" is much more valuable than having many keywords but the content is a 3rd level folder, dunno if I could explain it.

that's why zalando uses this format for product page: https://www.zalando.es/tommy-hilfiger-core-corporate-zapatillas-black-to112o0lx-q11.html with no folder but keywords in the product name

so mmmmm

justinbeaty:
> that's why zalando uses this format for product page: https://www.zalando.es/tommy-hilfiger-core-corporate-zapatillas-black-to112o0lx-q11.html with no folder but keywords in the product name

Hmm.. I think this could easily be a config option and we'd only need to change a few lines in `Mage_Catalog_Model_Url`. Basically it is only changing the joining character.

But `Mage_Catalog_Model_Url` is also an absolute mess. I assume every Mage developer has run into issues with the URL rewrite table at one point. I know there are problems because I broke the table several times when working on #81. I remember thinking a lot of the problems could be solved by using a composite primary key instead of an auto-increment one. But Varien collections don't exactly make that easy. 

In any case, everything would still work fine, as demonstrated by just creating a custom URL rewrite:

image

justinbeaty:
I forgot we hadn't merged this yet... 😅 

I pushed one fix for the `setRouteParams()` methods. I'm not sure why I had used `\w` for param values, since that only matches `[A-Za-z0-9_]`. I guess it had just worked in all my cases. But there could be dashes, commas, percent signs, etc.

I updated the default config values. I tested by dropping all rows from `core_config_data` and rebuilding, and it's ok.

I think the first level folder thing could wait for a new PR. 

Otherwise I think we've tested this quite a bit. Depending on when you'll do the release in July, I'd say we can merge it so that if there are any problems we find them while testing other PRs.

fballiano:
I'll recheck this one super soon :-)))

justinbeaty:
> I'll recheck this one super soon :-)))

Yeah, I'll test tomorrow too just since I changed the Regex.

fballiano:
I've tested this and everything seems ok!

I've updated the PR title/description.

:-)

justinbeaty:
I'll test momentarily, just want to double check various usages of the setUrlParam methods.

PR #192, opened 2025-06-05
Fixed field dependence and coupon code toggle on Sales Rule edit page

justinbeaty:
#54 caused two problems with editing a Shopping Cart Price Rule.

1\. The salesrule form set a field dependence like the following:

```php
            ->addFieldDependence(
                $couponCodeFiled->getName(),
                $couponTypeFiled->getName(),
                Mage_SalesRule_Model_Rule::COUPON_TYPE_SPECIFIC, // int(2)
            )
```

The "wanted" value in this case in an integer, but HTML select values are strings. I used JS's `Array.includes()` to compare, but this is a strict equality, so the check failed. All other dependence blocks seemed to set only strings, (i.e. `"2"`) for the wanted value, so I didn't catch this.

2\. The second thing is much more stupid. In `template/promo/salesrulejs.phtml`, the Mage devs toggled the "Coupon Code" input field depending on if the "Use Auto Generation" box was checked. They failed to take into account if the coupon type was set to "No Coupon". However, it worked because their code ran before the dependence block check, which then correctly hid the "Coupon Code" field. There was no particular reason on event ran before the other, so this was very brittle code.

---

A few other improvements:

* Rewrote without PrototypeJS, and removed a useless `eval()`.
* Fixed a bug also present in OM where you cannot save a shopping cart rule if "Use Auto Generation" is checked.

### DIFF
--- a/app/design/adminhtml/default/default/template/promo/salesrulejs.phtml
+++ b/app/design/adminhtml/default/default/template/promo/salesrulejs.phtml
@@ -5,7 +5,7 @@
  * @package     default_default
  * @copyright   Copyright (c) 2006-2020 Magento, Inc. (https://magento.com)
  * @copyright   Copyright (c) 2022 The OpenMage Contributors (https://openmage.org)
- * @copyright   Copyright (c) 2024 Maho (https://mahocommerce.com)
+ * @copyright   Copyright (c) 2024-2025 Maho (https://mahocommerce.com)
  * @license     https://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
  */
 ?>
--- a/app/design/adminhtml/default/default/template/promo/salesrulejs.phtml
+++ b/app/design/adminhtml/default/default/template/promo/salesrulejs.phtml
@@ -9,100 +9,90 @@
  * @license     https://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
  */
 ?>
-<script type="text/javascript">
-//<![CDATA[
-var couponTypeSpecific = '<?= Mage_SalesRule_Model_Rule::COUPON_TYPE_SPECIFIC ?>';
-var tmpButtonsActionsStorage = [];
+<script>
+const couponTypeSpecific = '<?= Mage_SalesRule_Model_Rule::COUPON_TYPE_SPECIFIC ?>';
 
 function disableEnableCouponsTabContent(disable) {
-    var containerId = 'promo_catalog_edit_tabs_coupons_section_content';
-    if($(containerId)){
-        var dataFields = $(containerId).select('input', 'select', 'textarea', 'button');
-        for(var i = 0; i < dataFields.length; i++) {
-            disable ? dataFields[i].disable().addClassName('disabled')
-                : dataFields[i].enable().removeClassName('disabled');
-        }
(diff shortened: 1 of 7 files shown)


app/design/adminhtml/default/default/template/promo/salesrulejs.phtml
```diff
@@ -5,104 +5,94 @@
...
-        if ($('promo_catalog_edit_tabs_coupons_section_content')
-            && $('promo_catalog_edit_tabs_coupons_section_content').visible()
-            && Ajax.activeRequestCount == 0
-        ) {
-            handleCouponsTabContentActivity();
+        if (result?.messages) {
+            setMessagesDivHtml(result.messages);
         }
+    } catch (error) {
+        setMessagesDiv(error.message, 'error');
```
fballiano, line 87:
```suggestion
        setMessagesDiv(error?.message || 'An error occurred', 'error');
```

does something like this make sense?

justinbeaty: Here it isn't needed, since error is a JS `Error` object which always has the message property.

But it could be useful to put it in `mahoFetch` in case a controller sends back only `{ error: true }`. I might as well add that to this PR.

fballiano:
you're a machine, merging in a sec

trabulium:
I seem to have a problem where this doesn't show the coupon code input field when you switch between no coupon and specific coupon code. The following seems to fix it.

`25   function handleCouponsTabContentActivity() {                                                                                                                                                  
26 +      const couponType = document.getElementById('rule_coupon_type').value;                                                                                                                     
27 +      const couponCodeRow = document.getElementById('rule_coupon_code')?.closest('tr');                                                                                                        
28 +                                                                                                                                                                                               
29       if (couponType == couponTypeSpecific) {                                                                                                                                                   
30 +          if (couponCodeRow) {                                                                                                                                                                 
31 +              couponCodeRow.classList.remove('no-display');                                                                                                                                    
32 +              couponCodeRow.style.display = '';                                                                                                                                                
33 +          }                                                                                                                                                                                    
34 +                                                                                                                                                                                               
35           if (document.getElementById('rule_use_auto_generation').checked) {                                                                                                                    
36               disableEnableCouponsTabContent(false);                                                                                                                                            
37               setElementDisable(document.getElementById('rule_coupon_code'), true);                                                                                                             
...                                                                                                                                                                                                
33               setElementDisable(document.getElementById('rule_coupon_code'), false);                                                                                                            
34           }                                                                                                                                                                                     
35       } else {                                                                                                                                                                                  
36 +          if (couponCodeRow) {                                                                                                                                                                 
37 +              couponCodeRow.classList.add('no-display');                                                                                                                                       
38 +              couponCodeRow.style.display = 'none';                                                                                                                                            
39 +          }                                                                                                                                                                                    
40           disableEnableCouponsTabContent(true);                                                                                                                                                 
41       }                                                                                                                                                                                         
42   }  `

trabulium:
Hi Justin. My apologies. It was a weird situation where I seemed to somehow have the changes in [app/design/adminhtml/default/default/template/promo/salesrulejs.phtml](https://github.com/MahoCommerce/maho/pull/192/files#diff-ec1460d84e70d6e2ba7a6a0b6661bbf2b59875839028009f49442b30cde4330c) but not the changes from the other files. I should have updated maho via composer first before messaging here.

PR #193, opened 2025-06-07
Added experimental load-on-intent JavaScript optimization feature

fballiano:
This PR:
  • Implements experimental JavaScript deferral system that loads scripts only on user interaction
  • Adds admin configuration option to enable/disable the feature with appropriate warnings

  Technical Details

  • Modifies HTTP response processing to defer JavaScript execution until user events (click, scroll, hover, etc.)
  • Maintains script execution order and handles both inline and external scripts
  • Includes comprehensive safeguards to prevent conflicts with admin panel and non-HTML responses
  • Only activates for frontend pages when explicitly enabled via system configuration
  • It doesn't support prototypejs/jquery "dom ready" events
  • This feature is always disabled on checkout, for safety reasons


justinbeaty:
> mmmmmmmm it scares me a bit. 3-5 secs is a lot of time for the use to not interact with a page they're seeing no?

I probably wasn't clear, but meant that even if there are no user interactions, scripts will be loaded 3-5 seconds after DOMContentLoaded. So basically `if (userInteraction || enoughTimeHasElapsed) loadScripts()`

fballiano:
no no I got it, what I mean was, how many times the user will not interact in some way with the page in those 3-5 seconds, triggering the load anyway?

fballiano:
this last version should load the scripts in parallel. 

I tried to create a hybrid version with defer for external scripts and converting to text/plain all inline scripts but I can't get it to work properly

fballiano:
I've enabled this feature on https://demo.mahocommerce.com/

fballiano:
yeah no exactly, it would be a simple jsdefer (which is also enabled on the demo), actually... we could make this PR become both things, it would be so to remove the need for something like https://packagist.org/packages/fballiano/openmage-defer-javascripts and it could still be useful

and if we managed to make our theme not dependent on js at all, in that case it could be switched on :-)

PR #195, opened 2025-06-07
Added minimal layout for checkout (no header, no footer)

fballiano:
I think this is pretty interesting:

Screenshot 2025-06-07 alle 18 41 34

At the moment it's enabled by default and it's configurable via:

Screenshot 2025-06-07 alle 18 44 33

This new page layout can also be used in `CMS -> Edit Page`, for example for specific marketing landing pages:

image


justinbeaty:
Actually, one question. We have this in `app.js`:

https://github.com/MahoCommerce/maho/blob/b04c1749c4b7e1442d0acd7e9669c4d602435b3c/public/skin/frontend/base/default/js/app.js#L693-L708

But we don't load `app.js` in minimal layout. Is that JS needed?

fballiano:
> Love this idea! I think some other platforms call it "Focused Checkout".

should we rename it?

> 1. Instead of the `<checkout_onepage_minimal>` handle, I made it `<page_minimal>` and put it in `layout/page.xml`. Because it can be useful for marketing landing pages which are usually also minimal layout. You can now use it in CMS > Edit Page:

niceeee, now I'm wordering, is it the same as empty.phtml?

> You can also set it through layout XML, so I reverted the change in `OnepageController.php`.
> 
> ```
>     <checkout_onepage_index translate="label">
>         <update handle="page_minimal" ifconfig="checkout/options/minimal_layout"/>
>         
>     </checkout_onepage_index>
> ```

great idea

> 2. Instead of removing the header and footer blocks, I just changed the template file. I did this for two reasons:

I tried that but thought it was messier, glad you made it work :-)


> 3. At some point, the logo on the regular layout became tiny (maybe the min touch size update?).

do you always see it like that or it's a specific break point? because I've never seen it like that

> I thought maybe it was just something on my end, but it's caused by not having an explicit width or height set in the system config. So I updated the `getLogoHeight()` function to return `50px` if both width and height are not set.

mmm I'm wondering if it can collide with the CSS (there's a max-height 50px if I remember correctly)

will test now :-)

fballiano:
> Will update comment in a sec, just wanted to mention in case you're itching to press that merge button.. 😅

I'm always itching to press the merge button 😂

justinbeaty:
Since there's some other stuff in `app.js` that you might want on a marketing landing page, I changed it so we only remove that script on the checkout page. For example, if you put a slideshow there, it wouldn't have worked.

Then I just had to wrap some of the stuff in `app.js` in checks for those elements. It seems to work fine.

fballiano:
will quickly test and merge in a bit :-)

fballiano:
The changes in app.js are actually very good in general, let's go!

fballiano:
mmm weirdly enought on https://demo.mahocommerce.com/checkout/onepage/ I get the footer, which I didn't have on my local installation :-\

fballiano:
no actually I have it locally too and didn't see it before

justinbeaty:
Hmm.. I don't have it here:

image


Cache issue maybe?

fballiano:
but now the footer is gone from all pages, I think the footer is getting cached for all pages so it's either shown on all pages or hidden from all pages

fballiano:
committed here: https://github.com/MahoCommerce/maho/commit/cae4be596f5bc76f3ea13bf3302417c11f9956b4 🙏🙏🙏

PR #198, opened 2025-06-08
Added stdin support to `maho translations:missing`

justinbeaty:
One small feature that is missing in the Maho command from the original translation script is the ability to pipe a list of files to check. This is useful to check for missing translations that were introduced in a specific PR.

Usage:
```sh
git diff --name-only main | ./maho translations:missing
```

This will check only the files modified in your branch, instead of showing everything.


justinbeaty:
> wow, this could be a github action!

I didn't even think of that! Only one caveat, if you change one file, it checks the entire file. So you could get errors for a previous missing translation. But then we can either ignore, or might as well fix while you're modifying that particular file.

PR #199, opened 2025-06-11
Added Media Library to CMS admin section

fballiano:
This PR is in preparation for another idea a have... ;-)


fballiano:
@justinbeaty I've removed the controller and some more stuff ;-)

PR #206, opened 2025-06-21
Added region support to SMTP functionality

fballiano:
@trabulium about the region support, I'd prefer a simpler implementation like this one, without the list of regions, because regions are supported by multiple transports but they may have different names or they may change with time.

would this be ok with you?


trabulium:
I am not sure about that. My bad, I didn't read the symfony docs on the component. I have about 5 different projects on in all different languages - Python, C, Flutter, PHP, React etc so my head is everywhere. Trust yourself more than me :)

PR #208, opened 2025-06-25
Added Admin Activity Log functionality

fballiano:


justinbeaty:
Just looking at the code, this is excellent! Very much a needed feature too. Will test it out today.

justinbeaty:
Checking this out now.

app/code/core/Maho/AdminActivityLog/Block/Adminhtml/Activity/View/Form.php
```diff
@@ -0,0 +1,256 @@
...
+        $fieldset->addField('username', 'label', [
+            'label' => Mage::helper('adminactivitylog')->__('Username'),
+            'value' => $activity->getUsername(),
+        ]);
+
+        $fieldset->addField('fullname', 'label', [
+            'label' => Mage::helper('adminactivitylog')->__('Full Name'),
+            'value' => $activity->getFullname(),
+        ]);
+
```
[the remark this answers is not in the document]
fballiano: yep sorry!

app/code/core/Maho/AdminActivityLog/Model/Observer.php
```diff
@@ -0,0 +1,521 @@
...
+
+        // Get all parameters for debugging
+        $allParams = $request->getParams();
+        $actionName = $request->getActionName();
+        $controllerName = $request->getControllerName();
+        $moduleName = $request->getModuleName();
+
+        // Debug log what we're seeing
+        Mage::log("Mass Action Debug - Action: {$actionName}, Controller: {$controllerName}, Module: {$moduleName}", null, 'adminactivitylog_debug.log');
+        Mage::log('Parameters: ' . print_r($allParams, true), null, 'adminactivitylog_debug.log');
```
justinbeaty, line 320: There's a few debug logs still. Should we remove this, or put it behind `Mage::getIsDeveloperMode()` ?

fballiano: nah, let's remove those, sorry

app/code/core/Maho/AdminActivityLog/etc/config.xml
```diff
@@ -0,0 +1,203 @@
...
+                <run>
+                    <model>adminactivitylog/observer::cleanOldLogs</model>
+                </run>
+            </adminactivitylog_clean_old_logs>
+        </jobs>
+    </crontab>
+    <default>
+        <admin>
+            <adminactivitylog>
+                <enabled>1</enabled>
```
justinbeaty, line 192: Do you think it should be enabled by default? The table could potentially get pretty big and I assume only a subset of users will want this feature.

fballiano: My thought was, this is a feature that people wouldn't activate but then complain that it wasn't active when they need to search for something weird that happened on the DB. That's why I thought about enabling it although not everything is logged.

But also, there's only 30 days logged, the rest should be automatically removed

fballiano: I don't want to impose this so let's talk about it, I totally see your point

justinbeaty:
I pushed some changes, as always feel free to revert anything if you'd like. Some minor changes:

1. Moved `cleanOldLogs()` methods to the resource models.
2. Old logs should still be cleaned even if logging is disabled.
3. Moved the `<events>` section to `<adminhtml>` node in config, that way you don't have to manually check, and the events won't even be fired on the frontend.
4. Consolidate all of the `Mage::getStoreConfigFlag()` into the helper, because it's easier to keep track. There were actually two bugs here:
    * Logs were checking the wrong path, so wouldn't be cleaned.
    * Admin login/logout was logged even if disabled.

And then the major change is that I reworked the observer code a bit. The main reason is that saving `core_config_data` was broken. It would save a row for each field on the page even if it wasn't changed, and each row would have the value of the last field on the page.

Hopefully I didn't screw anything up here. I tested a bit, but will continue tomorrow.

justinbeaty:
There's also the possibility of storing sensitive data. I created a new customer from the admin. The problem is that we don't know what the DB columns are when doing this:

image

app/code/core/Maho/AdminActivityLog/etc/config.xml
fballiano: I think it would be ok to only log the main action, but how to do that tho?

But actually order/shipping etc shouldn't even be logged because they're already `historical` in the sense that they've timestamps, they're not modifiable etc...

I think this PR is very useful for all of the data that it's not historically tracked (everything else actually, product change, inventory change, modifying a tax class, modifying a CMS block)

justinbeaty:
> yeah, actually I was thinking, what if we encrypt old_data and new_data?

I just pushed a change that should get the actual database fields for objects. This at least gets rid of the `password` field. Anything else that is logged would already be fields that are in the database. But I think encryption is still probably a good idea because this only works with `Mage_Eav_Model_Entity_Abstract` and `Mage_Core_Model_Resource_Db_Abstract` type objects.

app/code/core/Maho/AdminActivityLog/etc/config.xml
justinbeaty: > I think it would be ok to only log the main action, but how to do that tho?

The only thing I can think of is if we group it by "page visit". So you'd see something like `/catalog/product/save/id/123`, then click on that and see all the model saves that happened during that action. Not really sure the best way to implement that though, maybe using `preDispatch` somehow.

fballiano:
what do you think about this last version (you've to add action_group_id column to the adminactivitylog_activity table)?

justinbeaty:
> what do you think about this last version (you've to add action_group_id column to the adminactivitylog_activity table)?

Let me check it out!

For the encryption, could we put the logic in the model class? That way in the form you can just write: 
```php
$oldData = $activity->getOldData();
```

```php
class Maho_AdminActivityLog_Model_Activity extends Mage_Core_Model_Abstract
{
    #[\Override]
    protected function _construct(): void
    {
        $this->_init('adminactivitylog/activity');
    }

    public function getOldData(): array
    {
        $data = $this->getData('old_data');
        if ($data) {
            $decrypted = Mage::getModel('core/encryption')->decrypt($data);
            if (json_validate($decrypted)) {
                return json_decode($decrypted, true);
            }
        }
        return [];
    }

    // ...
```

justinbeaty:
> what do you think about this last version (you've to add action_group_id column to the adminactivitylog_activity table)?

Wow, that's pretty incredible!

I guess the only strange thing is that the `action` and `entity_type` don't necessarily reflect what's true. Is there some easy way to show like `mixed` or something if they're not all `create`?

image

fballiano:
yeah I don't know how to solve that actually because when I edit a product, the first model that gets updated it's not the product_model sooooo :-\\\

PR #209, opened 2025-06-29
Added simple defer javascript functionality in tendem with load-on-intent

fballiano:
This PR builds on https://github.com/MahoCommerce/maho/pull/193 to make it a bit more comprehensive, adding a simple defer-javascript functionality, kinda the same as https://github.com/fballiano/openmage-defer-javascripts

- It eliminates the need for a dedicated module
- the code is shared between the defer and the loadonintent


app/locale/en_US/Mage_Core.csv
```diff
@@ -444,5 +444,7 @@
...
+"JavaScript Deferral Mode","JavaScript Deferral Mode"
+"Defer Only","Defer Only"
+"Load on Intent","Load on Intent"
+"Disabled:</strong> No deferral (default behavior)<br/><strong>Defer Only:</strong> Moves all script tags to the end of the page, improving initial render performance<br/><strong>Load on Intent:</strong> <strong style="color:red">Highly Experimental!</strong> Defers all scripts until user interaction (click, scroll, hover, etc.) - designed for pure vanilla JavaScript and may not work with PrototypeJS or jQuery","Disabled:</strong> No deferral (default behavior)<br/><strong>Defer Only:</strong> Moves all script tags to the end of the page, improving initial render performance<br/><strong>Load on Intent:</strong> <strong style="color:red">Highly Experimental!</strong> Defers all scripts until user interaction (click, scroll, hover, etc.) - designed for pure vanilla JavaScript and may not work with PrototypeJS or jQuery"
```
justinbeaty, line 450:
```suggestion
"<strong>Disabled:</strong> No deferral (default behavior)<br/><strong>Defer Only:</strong> Moves all script tags to the end of the page, improving initial render performance<br/><strong>Load on Intent:</strong> <strong style="color:red">Highly Experimental!</strong> Defers all scripts until user interaction (click, scroll, hover, etc.) - designed for pure vanilla JavaScript and may not work with PrototypeJS or jQuery","<strong>Disabled:</strong> No deferral (default behavior)<br/><strong>Defer Only:</strong> Moves all script tags to the end of the page, improving initial render performance<br/><strong>Load on Intent:</strong> <strong style="color:red">Highly Experimental!</strong> Defers all scripts until user interaction (click, scroll, hover, etc.) - designed for pure vanilla JavaScript and may not work with PrototypeJS or jQuery"
```

justinbeaty (approved):
Not tested, but looks good to me. Can test tomorrow if you want, or merge if you think it's good.

app/locale/en_US/Mage_Core.csv
fballiano: wa, rookie mistake!

fballiano:
@justinbeaty in my OM module I've a regex to remove html commits, to avoid situations like this:

Screenshot 2025-06-29 alle 20 03 56

should I add it to this one too?

justinbeaty:
@fballiano How about if we only remove comments in the `<head>` tag? Something like this:
```php
if ($pos = strpos($html, '</head>')) {
    $html = preg_replace('/<!--(.*?)-->/s', '', substr($html, 0, $pos)). substr($html, $pos);
}
```

fballiano:
mmmmm I see what you mean but, at the end of the day, scripts can be anywhere in the HTML so I think either we remove the comments or not

could we have a setting for that? is that even something people want to care about?

I mean, in production, what are HTML comments for? can they have a real purpose? it seems they're always used for debug, which maybe can be avoided in production..

justinbeaty:
I could think of two possible edge-cases where removing comments could cause a problem:

1\. People used to write stuff like this:
```html
<script type="text/javascript">
<!--
  // JavaScript code here
//-->
</script>
```

2\. Technically comments are part of the DOM, so it could be possible some JS tries to read data contained within. For example reading a "template" before `<template>` existed, etc.

I'm not sure how likely at all either are, and those people could just not use this feature. But I figure the most likely place there's going to be a comment next to a `<script>` is the head so it's probably good enough.

fballiano:
but the fact that a comment is part of the DOM etc is valid also for the comments in the HEAD hehehe

so maybe we can just leave them there and maybe one day add a "html minify" feature that can be enabled/disable that will take care of that?

fballiano:
then let's merge and forget at the moment :-D

PR #220, opened 2025-06-30
Replaced `Zend_Pdf` with `TCPDF`

fballiano:
WIP


fballiano:
@justinbeaty question, I've search for alternatives to Zend_Pdf:
- tcpdf (similar approach, maintained)
- dompdf (HTML-to-PDF approach, completely different from what we have today, maintained)
- mpdf (last release in 2022, not sure about PHP 8.3+ support)

so I made this experiment with TCPDF but I kinda hate the experience of these PDF generation libraries, migrating to dompdf would require more work and more breaking change but wouldn't it be much more awesome to have simple phtml templates for PDFs instead of what we have now?

justinbeaty:
@fballiano There's also FPDF, which looks like it at least supports PHP 8.2, but not really in active development.

But I like the dompdf idea a lot. One of the headaches of the current zend implementation is that we have to overwrite classes in order to modify how the pdf looks. So it would be really amazing if we could integrate this into the layout xml system. Something like:

* Create a new `Mage_Core_Block_Pdf` class, probably somewhat similar to how `Mage_Core_Block_Abstract` is structured.
* There's an `output` attribute on blocks that defines which method is called.
* You'd create new layout handles for each PDF you want:

```xml
<sales_order_pdf>                                                                                                             
    <block type="core/pdf" name="root" output="toPdf">                                                                             
        <block type="adminhtml/sales_order_pdf_info" template="sales/order/pdf/info.phtml"></block>
        <!-- ... -->
```

* Then there would be an easy method to load a layout handle and get the output.

It would be nice to be able to override individual pieces of the PDF like you can do with HTML. It is definitely a huge departure from the current system, though.

fballiano:
it could be a set of phtml files, once rendered they get passed to dompdf and that's it, it would definitely be easier for developers in the long run, I'll make a test

PR #229, opened 2025-07-09
Replace `Zend_Soap` with `Laminas_Soap`

fballiano:
This is pretty much a drop-in replacement.

For this PR I chose Laminas instead of Symfony because, in the symfony world, there's nothing I could find to manage SOAP.

It would be nice to make the whole Laminas_Soap optional for websites that don't use any SOAP API, but that would make phpstan fail on all SOAP classes, @justinbeaty how would you tackle this? would requiring laminas_soap in our phpstan-plugin work?


justinbeaty:
> It would be nice to make the whole Laminas_Soap optional for websites that don't use any SOAP API, but that would make phpstan fail on all SOAP classes, @justinbeaty how would you tackle this? would requiring laminas_soap in our phpstan-plugin work?

This should definitely be doable with a PHPStan stub file. Let me take a look.

fballiano:
or we could require-dev laminas-soap?

fballiano:
> How about simply installing it here?

but then we can't test phpstan locally, which I always do :-\

justinbeaty:
> the stub is for sure the cleanest option 😊

I forgot stubs are only for fixing PHPDocs on classes that actually do exist.

Instead, we can have a folder that has "mock" classes. I put it at `.phpstan/mock` so it's a bit out of the way. Basically imagine it's an extra `lib/` folder that PHPStan will look for classes. The file tree need to be how composer expects, so `.phpstan/mock/Laminas/Server/Soap.php`. Then inside these classes you can just define the methods you need with the correct parameter and return types.

fballiano:
@justinbeaty question, should the stub be part of maho-phpstan-plugin?

justinbeaty:
> @justinbeaty question, should the stub be part of maho-phpstan-plugin?

It could be, but doesn't have to. I want to say typically they would belong to the main project, or sometimes projects make a stubs repo. The benefit of this is that you can include everything in one PR instead of making two PRs, having to release a new PHPStan plugin version, and updating the version here.

fballiano:
it works perfectly, I would have liked for the mocks to be in the phpstan-plugin repo cause they "belong" more to that plugin in my opinion on for some reason I'm always reluctant to add first-level folders :-\

fballiano:
true, but we hardly tag new releases on the phpstan-plugin and it takes a second for us :-) I only don't know how to do that hahaha

fballiano:
would this work? I could do that no problem

```
To move the .phpstan/mock files to the maho-phpstan-plugin project, you'll need to:

  1. In the maho-phpstan-plugin repository, create a structure like:
  stubs/
  └── Laminas/
      └── Soap/
          ├── Server.php
          └── Exception/
              └── ExceptionInterface.php
  2. Move the mock files from .phpstan/mock/ to the plugin repository, keeping the same namespace structure.
  3. Update the plugin's extension.neon to register the stub files:
  parameters:
      stubFiles:
          - stubs/Laminas/Soap/Server.php
          - stubs/Laminas/Soap/Exception/ExceptionInterface.php
  4. In your main project, update .phpstan.dist.neon:
    - Remove the scanDirectories section that references .phpstan/mock
    - Ensure the maho-phpstan-plugin is included in your includes section
  5. Delete the .phpstan/mock directory from the main project once confirmed working.
```

fballiano:
oki, I don't love this but that's ok :-)

justinbeaty:
> would have been possible to modify the plugin in the vendor folder and then, once working, move the changes and tag?

Not fully, because if you push to GitHub before tagging the release, it fails here. If you make everything work locally, tag a new release, but then we say let's also use this other function in the library, you have to make a new tag again. Or we decide to not some function/class we mocked before, we have to tag again.

What about mocking the classes locally like we did here, and before merging the PR, move it all over to the plugin repo? It's a bit of extra steps setting up the `scanDirectories` each time, but we avoid the extra files here. And we might not be doing optional packages that often.

fballiano:
we shouldn't really have classes extending non-existing classes, I think this is quite an exception because we can't remove SOAP but at the same time not all projects use it.

If you agree on https://github.com/MahoCommerce/maho-phpstan-plugin/pull/6 I'm pro it, but otherwise I'm fine with keeping the mocks here, at the end of the day it's a minor detail :-)

justinbeaty:
> If you agree on [MahoCommerce/maho-phpstan-plugin#6](https://github.com/MahoCommerce/maho-phpstan-plugin/pull/6) I'm pro it, but otherwise I'm fine with keeping the mocks here, at the end of the day it's a minor detail :-)

That PR isn't quite it. I can make the PR to show. We will need two folders though, stubs/mocks for the plugin itself, and stubs/mocks for the main maho repo. So like 

```
├── mock
│   ├── maho
│   └── maho-phpstan-plugin
└── stubs
    ├── maho
    └── maho-phpstan-plugin
```

fballiano:
branch restored ;-)

justinbeaty:
Okay, I changed https://github.com/MahoCommerce/maho-phpstan-plugin/pull/6.

Shall I merge it and tag? @fballiano

fballiano:
everything seems to work perfectly 🎉🎉🎉🎉🎉

PR #230, opened 2025-07-10
Updated TipTap to 3.0 and improved its configuration

fballiano:
This PR adds the possibility to have `class` and `style` attribute to every node

### DIFF
--- a/public/js/mage/adminhtml/wysiwyg/tiptap/extensions.js
+++ b/public/js/mage/adminhtml/wysiwyg/tiptap/extensions.js
@@ -60,13 +60,39 @@ const renderDirectiveImageUrl = (src, directiveObj, directivesUrl) => {
     return src;
 };
 
+const getWidgetTypeForSelection = (state) => {
+    const { from, to } = state.selection
+
+    // If we have a selected maho widget, return the same type
+    const selectedNode = state.doc.nodeAt(from, to);
+    if (selectedNode?.type.name.startsWith('mahoWidget')) {
+        return selectedNode.type.name;
+    }
+
+    // Otherwise traverse parent nodes skipping empty tags
+    const pos = state.doc.resolve(from);
+
+    let cur = pos.parent;
+    while (cur.content.size === 0 && ['paragraph', 'heading'].includes(cur.type.name)) {
+        cur = pos.node(pos.depth - 1);
+    }
+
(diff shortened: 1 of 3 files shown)


justinbeaty:
> but the problem is mahowidget, reading tiptap documentation it seems we should split the widgets into inline widgets and block widget 😱

Yeah, for sure. I have an idea for it. Basically make them `<div>` first, then querySelector for any inside a list of known inline elements. It should work, I'll try it out in a bit.

fballiano:
feel free to modify this PR, zero problems

justinbeaty:
Well similar problem. From the PHPDocs:

> If a value to be serialized is an object, then by default only publicly visible properties will be included. Alternatively, a class may implement [JsonSerializable](https://www.php.net/manual/en/class.jsonserializable.php) to control how its values are serialized to JSON.

But the problem is that a varien object's `$_data` array is protected. But even if it was public, we really only want the keys from there and not other public properties. This would have been fixed by #55 if not for the circular references.

justinbeaty:
The last part took a bit longer, but I actually went back and simplified the whole thing a bit. Just lightly tested so far, but it seems to be working.

We now have `MahoWidgetBlock` and `MahoWidgetInline`. During parse, we first assume they'll be `<span>`, then convert to `<div>` if only `<div>` parents. We can easily support other parent type tags too, like `<main>`, `<section>`, etc when we have them. The easiest way to do this is to just parse it as HTML and use querySelector. Regex would be a mess. Then in extensions.js, we have similar code. We want to find out if a block is allowed in the current selection or not.

Will take a look at the copilot warning now.

fballiano:
@justinbeaty on CMS pages I get this error

```
js.js:140 Uncaught TypeError: Failed to construct 'URL': Invalid URL
    at setRouteParams (js.js:140:11)
    at tiptapWysiwygSetup.setup (setup.js:237:33)
    at tiptapWysiwygSetup.initialize (setup.js:30:14)
    at new tiptapWysiwygSetup (setup.js:15:14)
    at HTMLDocument.<anonymous> (9efe363cc97e29b901ecd0d2d81d5377/:572:42)
```

fballiano:
there's this case too 🤦🏻

Screenshot 2025-07-16 alle 09 41 55

PR #231, opened 2025-07-12
Replaced `Zend_Json` with `json_encode`/`json_decode` and `Zend_Json_Server` with `laminas/laminas-json-server`

fballiano:


justinbeaty:
On main branch:

```php
<?php

Mage::init();

// Simple test
$obj = new Varien_Object();
$obj['foo'] = 'bar';
var_dump(json_encode($obj));


// Nested test
$obj = new Varien_Object();
$obj['child'] = new Varien_Object();
$obj['child']['foo'] = 'bar';
var_dump(json_encode($obj));


// Array test
$results = [];
for ($i = 0; $i <3; $i++) {
    $obj = new Varien_Object();
    $obj['foo'] = 'bar' . $i;
    $results[] = $obj;
}
var_dump(json_encode($results));
```

Results:
```
string(2) "{}"
string(2) "{}"
string(10) "[{},{},{}]"
```

It's because `json_encode` doesn't know about how `Varien_Object` stores data in the `$_data` array.

Those specific examples were fixed with #55, but it didn't work properly when you had nested Varien_Objects (or any class that extends them), which is common with stuff like `$this->getOrder()` returning a `Mage_Sales_Model_Order` object.

I really had wanted to get rid of Zend_Json too. I think we may always need a compatible JSON serializer even if it's not Zend specifically.  Laminas probably works here, right?

justinbeaty:
@fballiano I just checked the example in #70. It wasn't just nested objects, because that actually did work with #55, but it was nested objects with circular references:

```php
$x = new Varien_Object();
$y = new Varien_Object();

$x->setY($y);
$y->setX($x);

json_encode($x, JSON_THROW_ON_ERROR);
```

With #55 it would throw `Uncaught JsonException: Recursion detected`, but the above patch just runs out of memory:

image

---

The circular references happen quite a lot. For example the order object has a reference to the order items, and the order item objects have a reference to the order object.

I think we can just take the toJson code from Zend and put it directly somewhere in lib. I think it's really only like one function, and can be simplified for our use case.

justinbeaty:
Or actually, something like this could also work too. We just need a way to reset the `hasSerialized` state after the call to `json_encode` is done. 

```php
    /**
     * Implementation of JsonSerializable::jsonSerialize()
     *
     * @link https://www.php.net/manual/en/jsonserializable.jsonserialize.php
     */
    #[\Override]
    public function jsonSerialize(): array
    {
        $data = $this->__toArray();
        foreach ($data as $key => $value) {
            if ($value instanceof Varien_Object && $value->hasSerialized) {
                unset($data[$key]);
            }
        }
        $this->hasSerialized = true;
        return $data;
    }
```

justinbeaty:
One more idea, this might be the best solution, but I haven't fully tested yet:

```php
    /**
     * Implementation of JsonSerializable::jsonSerialize()
     *
     * @link https://www.php.net/manual/en/jsonserializable.jsonserialize.php
     */
    #[\Override]
    public function jsonSerialize(): array
    {
        $data = $this->__toArray();
        foreach ($data as $key => $value) {
            if ($value instanceof Varien_Object) {
                $reflection = new ReflectionClass($this);
                if ($reflection->getMethod('jsonSerialize')->getDeclaringClass()->getName() !== 'Varien_Object') {
                    continue;
                }
                if ($reflection->getMethod('toJson')->getDeclaringClass()->getName() !== 'Varien_Object') {
                    $data[$key] = $value->toJson();
                } else {
                    unset($data[$key]);
                }
            }
        }
        return $data;
    }
```

---

Basically, for each Varien_Object child, it checks:
- If that child  has it's own `jsonSerialize()` method. Nothing in Magento ever had this since it was a newer feature, but we will give priority to this method of children have it.
- If there's a `toJson` function that *is not* the varien object one.
- Otherwise we just won't serialize it.

So child classes need to take care of avoiding recursion themselves, but that's sort of reasonable to expect. For example if an Order Item object defines a `jsonSerialize` or `toJson` method, they shouldn't be outputting the order object.

fballiano:
@justinbeaty I've put this code in https://github.com/MahoCommerce/maho/pull/234, it should be easier to test

PR #234, opened 2025-07-16
Fixing regression in jsonSerialize implementation

fballiano:
Please check comments in https://github.com/MahoCommerce/maho/pull/231 to understand the purpose of this PR.


justinbeaty:
The diff looks good to me. Can test shortly, otherwise I expect everything would be fine. If so, then I'm good with merging. :)

PR #241, opened 2025-07-18
Revert "Fixed CodeQL warning 71"

fballiano:
Reverts MahoCommerce/maho#240

### DIFF
--- a/public/js/mage/adminhtml/tools.js
+++ b/public/js/mage/adminhtml/tools.js
@@ -789,7 +789,7 @@ function setMessagesDiv(message, type = 'success', div = null) {
 */
 function setMessagesDivHtml(html, div = null) {
     if (div ??= document.getElementById('messages')) {
-        div.replaceChildren(...xssFilter(html, true));
+        div.innerHTML = xssFilter(html);
     }
 }
 
--- a/public/js/varien/js.js
+++ b/public/js/varien/js.js
@@ -211,20 +211,18 @@ function stripTags(str, removeScriptAndStyleContent = false) {
 
 /**
  * Alternative to PrototypeJS's string.stripScripts() method that also removes event attributes
- * @param {string} str - HTML string to sanitize
- * @param {boolean} returnDOM - If true, returns sanitized DOM nodes; if false, returns HTML string
  */
-function xssFilter(str, returnDOM = false) {
+function xssFilter(str) {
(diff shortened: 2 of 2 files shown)


fballiano:
@justinbeaty I've reverted this change but i've no idea how to fix this :-\

justinbeaty:
The description of the warning is a bit silly:

To fix the issue, ensure that the *xssFilter* function returns text content instead of HTML to *prevent the possibility of XSS*.

But our implementation of xssFilter is just probably not passing what GitHub would consider sufficient. Since we only use xssFilter on the admin, we could consider using a vetted library. I avoided it originally, but if this is something that will keep causing CodeQL warnings it might be worth it.

Maybe something like https://github.com/cure53/DOMPurify/blob/main/dist/purify.min.js ?

fballiano:
yeah I've seen DomPurify but it's about 23kb and it seems a bit overkill but... if you think we've to use it then I'm ok with it

justinbeaty:
Here's one that's much smaller, but seemingly just one developer: https://github.com/julienetie/xsskillah/blob/main/src/xsskillah.js (2.11KB minified)

It would seem this library's approach is to simply not allow [a bunch of tags](https://github.com/julienetie/xsskillah/tree/main?tab=readme-ov-file#filtered-tags), which makes it smaller and faster. If you look at the [places we use xssFilter](https://github.com/search?q=repo%3AMahoCommerce%2Fmaho%20xssfilter&type=code), I'd say this is fine.

fballiano:
nice find, since dompurify does a bunch more stuff and xsskillah only works with HTML it seems it should be fine :-)

justinbeaty:
> nice find, since dompurify does a bunch more stuff and xsskillah only works with HTML it seems it should be fine :-)

Yeah, let's try it out. I think we should move our `xssFilter` function to adminhtml/tools.js, and just call xssKillah from there.

fballiano:
I'm trying xsskillah but I can't make it work and there's zero documentation :-\

fballiano:
ok it works this way

```js
function xssFilter(str) {
    const sanitize = xssKillah();
    const safeNodes = sanitize(str);
    return Array.from(safeNodes).map(n => n.outerHTML || n.textContent).join('');
}
```

but I don't think we've a way of adding js with type=module :-\

PR #242, opened 2025-07-18
Introduced xssKillah

fballiano:
@justinbeaty it seems to work this way, I didn't want to add a new file for this small library, would this be ok with you?


fballiano:
yeah it worked! let's hope we didn't break anything :-\

justinbeaty:
> yeah it worked! let's hope we didn't break anything :-\

Great! I think it will be fine, because those tags for sure aren't needed for admin messages, and would have been stripped from tiptap anyway. Also wasn't really necessary in the uploader instance, just that for sure other devs than Magento ones wrote that.

PR #246, opened 2025-07-22
Added HTTP Security Headers for Enhanced PCI DSS Compliance

fballiano:
**Headers Added**

  - HSTS (Strict-Transport-Security): Prevents SSL stripping attacks on payment pages
  - X-Content-Type-Options: Blocks MIME confusion attacks via file uploads
  - X-XSS-Protection: Provides legacy browser XSS protection
  - Referrer-Policy: Prevents session token leakage to payment gateways

  Key Features

  - Per-website configuration for multi-store flexibility
  - Smart HTTPS detection (HSTS only applies on secure connections)
  - Secure defaults enabled out-of-the-box
  - Granular control - enable/disable headers independently
  - Backward compatible with existing installations

  **Default Values & Rationale**

  **HSTS Configuration**

```
  hsts_enabled = 0
  hsts_max_age = 31536000 (1 year)
```

  - Disabled by default
  - 1-year max-age follows industry standard (Google Chrome preload requirement)

  **Content Security Headers**

```
  content_type_options_enabled = 1
  xss_protection_enabled = 1
```

  - Both enabled provide essential protection with no compatibility issues
  - X-Content-Type-Options prevents upload-based XSS attacks common in ecommerce
  - X-XSS-Protection supports legacy browsers still used by many customers globally

  **Referrer Policy**

`  referrer_policy = "strict-origin-when-cross-origin"`

  - Balances security and functionality for payment gateway integrations
  - Same-origin requests (internal Maho pages) get full referrer for analytics
  - Cross-origin requests (PayPal, Stripe) only get clean domain, preventing token leakage
  - Compatible with all major payment processors

  **Configuration**

System > Configuration > Admin > Security


justinbeaty:
Hm, I am still not sure about enabling by default unless it was a very short expiration, like 24 hours, but that pretty much negates the point of HSTS. I am trying to see if any other framework enables HSTS by default.

fballiano:
in M2 it's not enabled by default but it's 1years https://github.com/mage-os/mageos-magento2/blob/1670d3a164fb324adf717d7b8e1663205e6d8148/app/code/Magento/Store/Model/HeaderProvider/Hsts.php#L15

it kinda makes sense

fballiano:
agree, will change asap but also disable subdomains

justinbeaty (approved):
LGTM!

PR #270, opened 2025-08-03
Use PDO to execute SQL files in sample data installation

empiricompany:
Modify the sample data installation process to use PDO for executing SQL files instead of the mysql command line client.

This change improves compatibility with Docker environments like maho-starter where the mysql client is not available. By using PDO directly, we can execute SQL files without relying on external tools.

This approach maintains the same functionality while making the installation process more portable across different environments.

### DIFF
--- a/lib/MahoCLI/Commands/Install.php
+++ b/lib/MahoCLI/Commands/Install.php
@@ -8,8 +8,6 @@
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
-declare(strict_types=1);
-
 namespace MahoCLI\Commands;
 
 use Exception;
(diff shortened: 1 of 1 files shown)


lib/MahoCLI/Commands/Install.php
```diff
@@ -8,8 +8,6 @@
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
-declare(strict_types=1);
-
```
fballiano, line 12: I'd restore the strict types if possible

lib/MahoCLI/Commands/Install.php
```diff
@@ -173,31 +171,44 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-                    return Command::FAILURE;
+
+            try {
+                // Create PDO connection
+                $dsn = "mysql:host={$dbHost};dbname={$dbName};charset=utf8";
+                $pdo = new \PDO($dsn, $dbUser, $dbPass);
+                $pdo->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
+
+                // Set PDO to allow multiple queries in a single call
+                $pdo->setAttribute(\PDO::ATTR_EMULATE_PREPARES, 0);
```
fballiano, line 183: ATTR_EMULATE_PREPARES shoulnd't be necessary no?

empiricompany:
i use this images:
https://github.com/x86fantini/mahocommerce-docker-starter
with this patch
https://github.com/x86fantini/mahocommerce-docker-starter/compare/main...empiricompany:mahocommerce-docker-starter:main

And from inside the PHP container, there is no MySQL direct client; only PDO is possible.

fballiano:
this patch is good improvement anyway, but in that dockerfile I think it's anyway better to install the mysql client

PR #474, opened 2026-01-01
Fixed unnecessary `GROUP BY` in Flat Category getProductCount()

fballiano:
## Summary

- Removed redundant `GROUP BY` clause from `Mage_Catalog_Model_Resource_Category_Flat::getProductCount()` that was causing unnecessary performance overhead

## Problem

When Catalog Flat Tables are enabled, the `getProductCount()` method included a `GROUP BY category_id` clause on a query that already filters by a single `category_id` via `WHERE`. This forced MySQL to:
- Create temporary tables
- Perform filesort operations
- Use query plan: `"Using where; Using index; Using temporary; Using filesort"`

The non-flat version (`Mage_Catalog_Model_Resource_Category::getProductCount()`) correctly omits the GROUP BY and uses the optimal plan: `"Using index"`

## Solution

Removed the unnecessary `GROUP BY` clause to match the non-flat implementation.

## Thanks

Thanks to @loekvangool for discovering this issue.
See: https://github.com/OpenMage/magento-lts/issues/5206

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Category.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Category.php
@@ -522,12 +522,9 @@ class Mage_Catalog_Model_Resource_Category extends Mage_Catalog_Model_Resource_A
                 ['main_table' => $productTable],
                 [new Maho\Db\Expr('COUNT(main_table.product_id)')],
             )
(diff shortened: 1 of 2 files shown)


loekvangool:
Nice.

PR #555, opened 2026-02-14
Added `content versioning` for CMS pages, CMS blocks, and blog posts

mageaustralia:
## Summary

- Adds `Maho_ContentVersion` module — automatic version snapshots of CMS pages, CMS blocks, and blog posts on every save (admin panel, API, or CLI)
- Versions tab added to all three entity edit forms in admin, with preview and one-click restore
- CMS blocks and blog posts converted to tabbed layouts for consistency with CMS pages
- Configurable retention: max versions per entity (default 20) and max age in days (default 90)

## Key Design Decisions

- **Observer-based** (`model_save_before`) — captures ALL saves regardless of source
- **Polymorphic table** — single `content_version` table with `entity_type` discriminator
- **Inline cleanup** — expiry runs after each insert, no cron dependency
- **Reversible restores** — restoring a version auto-snapshots the current state first
- **Sandboxed preview** — HTML content rendered in `<iframe sandbox="">` to prevent XSS

## Files Changed (26 files)

- **New module**: `app/code/core/Maho/ContentVersion/` (model, service, observer, controller, blocks, helper)
- **Tab conversion**: CMS blocks and blog posts now use tabbed edit forms
- **Layout**: Versions tab injected via `contentversion.xml` layout update
- **Translations**: Full `en_US` CSV with all 26 strings

## Test plan

- [x] Edit a CMS page in admin — verify version row created in Versions tab
- [x] Edit same page again — verify version number increments
- [x] Click Preview — opens in new tab with admin layout, shows snapshot data
- [x] Click Restore on an older version — content restored and new version created
- [x] Edit a CMS block and blog post — Versions tab appears and works
- [x] Verify blog post edit has Content, Meta, and Versions tabs
- [x] Set max_versions to 3 in config, create 4+ versions — oldest pruned


mageaustralia:
Adds versioning for CMS Pages, Blocks and Blogs - 
image

Also converted Blog to have tabs to follow the same pattern as CMS pages:

image

mageaustralia:
I just pushed the API user detection + directive processing cherry-pick (commit 490013947) - that's to work when the API stuff goes live but won't interfere before then. All good to go. 👍

fballiano:
Hi @mageaustralia, I'm a bit against the directive processing since it's not even executed in the "real" view/edit of the content (and it would modify the "content" that the user is trying to understand if it's worth reverting to), we should maybe have the rich text editor to render the preview, but i think we're just overthinking and it's ok to have it "raw" at the moment.

And since the merging of APIPlaform may be months away I'm not really into merging the current_api_user since at the moment it's not matching any code, yes we'll have to remember to reintroduce it but we've been on a years long path of removing dead code that I think it's better not to have it and add it later.

Sorry but being a pain 😅

**EDIT**: my mistake on the preview breaking, sorry!

PR #578, opened 2026-02-23
Rewrote the API on API Platform: REST v2, GraphQL, JWT auth, granular permissions and OpenAPI docs

mageaustralia:
## Summary

Adds a complete REST + GraphQL API layer to Maho, built on Symfony / API Platform running alongside Maho via `rest.php`. Maho is used as a library (models, config, DB) while Symfony owns the HTTP layer, with no changes to the existing front controller. The new API ships as **beta**: fully tested and hardened, with the contract open to small refinements based on integrator feedback during its first release cycle.

Full documentation: **https://mahocommerce.com/api/v2/**

### What's included

- **~140 REST endpoints (212 operations)** covering products (incl. bundle/configurable/grouped/downloadable sub-resources), categories, orders, invoices, shipments, credit memos, customers, addresses, carts and guest carts, checkout, CMS, blog, reviews, wishlists, newsletter, coupons, gift cards, inventory, tax, directory, stores, and EU revocation requests
- **Dual protocol**: every resource is exposed over both REST (JSON or JSON-LD/Hydra) and GraphQL, sharing the same providers/processors so behavior never diverges
- **JWT authentication** (`lcobucci/jwt`, HS256, strict issuer/audience/validity constraints) with three grants: `customer`, `client_credentials`, and legacy `api_user`; tokens are revocable (logout, password change) and users are re-validated against the DB on every request
- **Granular permission system**: resources declare `resource/operation` permissions via `#[Maho\Config\ApiResource]`, compiled into a registry at `composer dump-autoload` and managed through a new admin role/user editor; service accounts can also be scoped to specific stores
- **Guest cart flow** with secure masked IDs, gift messages, gift cards, and full checkout
- **HTTP caching**: ETag + 304 support, public/private Cache-Control tiers, tagged response caches with event-driven invalidation
- **Idempotency keys** for safe mutation retries: DB-backed replay with 24h TTL, 409 on concurrent duplicates, automatic reclaim of reservations abandoned by crashed requests
- **Security hardening**: all protocols disabled by default (404 at the entry point), CORS origin allowlist validation, fail-closed same-origin checks on the admin GraphQL console, rate limiting, provider-agnostic CAPTCHA events, timing-safe comparisons, encrypted auto-generated JWT signing secret
- **OpenAPI 3.1** spec always available at `/api/docs.json`; browsable Swagger UI / ReDoc / GraphiQL as an opt-in extra
- **Extensibility**: third-party modules add resources with one attribute, and can extend every core DTO through `api_*_dto_build` events in a dedicated `api` event area

### Architecture

```
web server -> /api/* -> rest.php -> Symfony kernel (API Platform)
                                        |
                                 Mage::init() (library mode)
                                        |
                          Providers/Processors use Mage models
```

Resources live in the owning module (`Catalog/Api/`, `Sales/Api/`, `Checkout/Api/`, ...) and are discovered by convention, with no central registration file.

### Compatibility and rollout

- **Merging this exposes nothing.** Every protocol (REST v2, GraphQL, admin GraphQL, legacy REST, SOAP, XML-RPC, JSON-RPC) defaults to **off** and returns 404 until enabled in System > Configuration > Services > API Platform.
- The legacy Magento 1 API code paths are preserved and still work once their protocol is enabled.
- **Dependency posture**: the API core (api-platform/core, framework/security bundles, nelmio/cors-bundle, webonyx/graphql-php, lcobucci/jwt, symfony/expression-language, ~19 MB in vendor) is required but only ever loaded on `/api/*` requests; heavyweight or niche pieces are opt-in `suggest` packages: laminas-soap / laminas-xmlrpc / laminas-json-server for the legacy protocols, symfony/twig-bundle + symfony/asset for the browsable Swagger UI.

### Testing

- Dedicated `Api` Pest suite: 565 tests / 1890 assertions exercising REST reads and writes, GraphQL, auth, permissions, caching, idempotency, and CORS
- Full CI matrix green: MySQL, MariaDB, PostgreSQL, SQLite across PHP 8.3 / 8.4 / 8.5, plus CodeQL, PHPStan level 6, php-cs-fixer, and rector

## Release notes checklist

Items to carry into the release announcement:

- [ ] **Action required for existing API integrations**: all API protocols now default to off. Stores using the legacy SOAP / XML-RPC / JSON-RPC / REST APIs must re-enable them under System > Configuration > Services > API Platform after upgrading, and install the matching laminas package for SOAP/XML-RPC/JSON-RPC (`composer require laminas/laminas-soap`, etc.)
- [ ] **Web server configuration**: the new API URLs are served only through `rest.php`; nginx/Caddy users must add the documented rewrite rules (Apache `.htaccess` ships with them)
- [ ] **REST v2 + GraphQL are beta**: field names and shapes may receive small refinements in the next release based on feedback
- [ ] **JWT signing secret** is auto-generated and stored encrypted on first use; no manual setup needed
- [ ] New cache type **API Data** appears in Cache Management; new cron job `apiplatform_idempotency_cleanup` runs daily at 03:00


mageaustralia:
> maybe we should check out,

More like 1000% need to fix. Interestingly, I'm certain I had dealt with many of these issues already. I'll revise it all though!

fballiano:
## File-type custom options not supported via API

While reviewing this PR in the context of the [Magento 2 PolyShell vulnerability](https://sansec.io/research/magento-polyshell) (CVE-2025-54263 through CVE-2025-54267, [fix commit](https://github.com/magento/magento2/commit/796c4ce195cee0814ac92e5a19fc2ecfa79dae69)), I noticed that products with **file-type custom options** cannot be properly added to cart through the new REST/GraphQL API.

The file option handler (`Mage_Catalog_Model_Product_Option_Type_File::_validateUploadedFile()`) only reads from `$_FILES`, which is always empty in a JSON API context. This means:
- **Non-required file options**: silently ignored (product adds to cart without the file)
- **Required file options**: blocks the add-to-cart entirely ("Please specify the product required option")

This is a missing feature rather than a security issue — the API simply cannot handle file uploads for custom options yet.

### Security considerations for future implementation

When this gets implemented (likely by accepting base64-encoded file data in the options payload), these are the pitfalls to avoid — this is exactly where Magento 2 got PolyShelled:

1. **Only process file data when the option type in the database is actually `file`** — never trust the user input for determining the option type
2. **Validate the file extension** against both the option's allowed extensions and the global forbidden extensions list (`catalog/custom_options/forbidden_extensions`)
3. **Validate actual file content** (e.g. `getimagesizefromstring()` for images, `finfo` mime detection for other types) — do not trust the declared MIME type
4. **Write to a non-web-executable directory**, or ensure server config blocks PHP execution in the target directory
5. **Run the same validation pipeline** that `_validateUploadedFile()` uses for `$_FILES`-based uploads

fballiano:
### Dependency cleanup & JWT library consolidation

Cleaned up `composer.json` to remove 7 redundant dependencies that were already being pulled in transitively:

- `symfony/http-kernel` — required by `symfony/framework-bundle`
- `symfony/routing` — required by `symfony/framework-bundle`
- `symfony/serializer` — required by `api-platform/core`
- `symfony/property-info` — required by `api-platform/core`
- `symfony/property-access` — required by `api-platform/core`
- `symfony/yaml` — pulled in transitively by routing/serializer

This also migrates from `firebase/php-jwt` to `lcobucci/jwt` (5.x). Since `league/oauth2-server` already depends on `lcobucci/jwt`, we were shipping two separate JWT libraries that do the same thing. By consolidating on `lcobucci/jwt`, we drop one dependency entirely.

Net result: the API Platform feature branch went from adding **16 new entries** in `composer.json` down to **9**.

fballiano:
@mageaustralia CrudResource seems like an awesome thing to have, I wanted to convert as much as possible to it but...

⏺ The analysis is sobering. Most resources are too complex for CrudResource:

  ┌───────────────────────────┬─────────┬────────────────────────────────────────────────────────────┐
  │         Resource          │ Verdict │                            Why                             │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ BlogCategory              │ YES     │ Read-only, flat scalars, standard store filter             │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ CmsBlock                  │ PARTIAL │ Content directive/sanitizer, status mapping, store access  │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ CmsPage                   │ PARTIAL │ Same as CmsBlock + identifier lookup, root_template rename │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ BlogPost                  │ PARTIAL │ Same as CMS + image file handling in processor             │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ Country, Currency, Region │ NO      │ Not model-backed, assembled from runtime state             │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ Store, StoreConfig        │ NO      │ App-state assembly, config reads                           │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ Newsletter, Contacts      │ NO      │ Workflow endpoints, not CRUD                               │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ Review                    │ NO      │ Two-model save (review + rating), aggregation              │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ StockUpdate               │ NO      │ Raw SQL for performance, bulk operations                   │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ Coupon                    │ NO      │ Two-model save (coupon + rule), condition trees            │
  ├───────────────────────────┼─────────┼────────────────────────────────────────────────────────────┤
  │ GiftCard                  │ NO      │ Domain methods (adjustBalance), email notifications        │
  └───────────────────────────┴─────────┴────────────────────────────────────────────────────────────┘

  Only BlogCategory is a clean conversion. The CMS resources and BlogPost could work with hooks, but they'd still need custom Provider/Processor subclasses for content processing,
  sanitization, and store access validation — the savings would be modest (~80-100 lines each, but you'd still need ~40 lines of hook code).

what do you think?

mageaustralia:
It seems that ModuleApiDiscovery::discover() only scans BP . '/app/code/...' (project root), so on a Composer-based Maho project where core modules live under vendor/mahocommerce/maho/app/code/..., it never sees them. It only finds whatever local additions/overrides the project has overridden into its own app/code/.

What's your thoughts on this setup?

Adopt the same pattern you already use for config discovery. In maho-composer-plugin's AttributeCompiler, you solved the exact equivalent problem for etc/config.xml files: scan both project root and vendor at build time (during composer dump-autoload), compile the merged result into vendor/composer/maho_attributes.php, and at runtime just require that pre-built file.

 So: add a sibling ApiResourceCompiler to the same plugin.

 At build time (post-autoload-dump hook) it would:

1.  Walk BP/app/code/*/*/*/Api/ (project root — finds local overrides and community modules like MageAustralia/FilterablePages/Api/)
2.  Walk BP/vendor/mahocommerce/maho/app/code/*/*/*/Api/ (vendor — finds all the core Mage_Catalog_Api_Product / Mage_Sales_Api_Order resources in their natural module homes)
3. Merge the two lists, with project root winning on collisions (so local overrides work)
4. Write the result to vendor/composer/maho_api_resources.php as a plain PHP array

 At runtime, ModuleApiDiscovery::discover() stops doing any filesystem walking and just requires that pre-built file.

  **Three potential wins:**
  - Deployment-agnostic - running via composer vs git > app/code/core/ both work identically, the plugin compiles from whatever filesystem layout it finds.
  - Faster  ie: no runtime directory traversal; the file is included once and cached in opcache like any other PHP file.
  - Override-friendly — putting a file at app/code/core/Mage/Catalog/Api/Product.php overrides the vendor one, which is exactly what you'd expect. The override rule is explicit in the compiler (project root wins) rather than implicit in alphabetical classmap ordering.

It seems you already did this exact thing for configs. So it seems like this continues the pattern you already have and use?

What do you think?

fballiano:
Hey @mageaustralia I think I've managed to solve it using part of the routines of the composer plugins.

And actually this is the reason why I'm working a lot on PHP attributes and consequently on the symfony/routing migration, because that would allow for a tighter integration in this PR too and maybe some optimizations/coderemoval.

It's taking a lot of time but there's no way around it I think :-(

fballiano:
## URL restructure: new REST namespaced under `/api/rest/v2`

Just pushed `fd6206b562` reorganising the new API URL space. Summary for reviewers:

### What changed
- **New REST**: `/api/<resource>` → `/api/rest/v2/<resource>` (e.g. `/api/products` → `/api/rest/v2/products`)
- **GraphQL**: unchanged at `/api/graphql` and `/api/admin/graphql`
- **Docs**: unchanged at `/api/docs`
- **Legacy `Mage_Api`**: untouched — `/api/rest`, `/api/soap`, `/api/v2_soap`, `/api/xmlrpc`, `/api/jsonrpc` keep their original handlers

### Why
Two reasons:
1. **Symmetry with legacy paths.** Magento 1 already used a protocol-namespaced URL space (`/api/rest`, `/api/soap`, `/api/v2_soap`). Putting the new REST at bare `/api/<resource>` was the only protocol *not* under a protocol prefix, which read as inconsistent. `/api/rest/v2/...` slots in alongside the existing siblings.
2. **Future-proof versioning before clients pin URLs.** Doing this now is cheap; doing it after launch means coordinating with every integrator.

### Why GraphQL is *not* versioned
GraphQL evolves the schema in place — clients only fetch the fields they ask for, so additive changes don't break anyone. The convention is to deprecate fields with `@deprecated` rather than version the URL. There's no legacy GraphQL to coexist with either, so `/api/graphql/v1` would be solving a non-problem. If a breaking schema rewrite is ever needed, `/api/graphql/v2` can be added then.

### Implementation note
Used API Platform's built-in `api_platform.defaults.route_prefix = '/rest/v2'` (one line in `Kernel.php`) — covers all 36 resource DTOs without touching their `uriTemplate` declarations. GraphQL/docs/entrypoint routes are registered separately by API Platform and aren't affected by `route_prefix`, which is exactly what we want.

### Files touched
- `Kernel.php` (route prefix), `public/.htaccess` (rewrite ordering), `HttpCacheListener.php` / `Observer.php` / `InvoiceProvider.php` (3 hardcoded paths), `ApiPermissionRegistry.php` (comment)
- `docs/API.md` — rewrote Nginx/Apache/Caddy config blocks to route REST + GraphQL + docs to `rest.php` while letting legacy paths fall through
- 50 test files — bulk URL update (preserved `/api/graphql`, `/api/docs`, legacy `/api/rest/...` fixtures, `/api/soap`, `/api/invalid_type`)

### Verified
- `vendor/bin/php-cs-fixer fix --dry-run` clean
- `vendor/bin/phpstan analyze` `[OK] No errors`
- `php -l` on modified files clean

End-to-end test run still pending against a properly configured server.

fballiano:
## API protocols are now opt-in (disabled by default)

Pushed `8c534505ce`. Every API protocol — new and legacy — defaults to **disabled** in `config.xml`. Operators must consciously enable each one in **System > Config > Services > API Platform > API Protocols**.

### Toggles (all default 0)

| Config path | Protocol | URL |
|---|---|---|
| `apiplatform/protocols/rest_v2` | New REST | `/api/rest/v2/*` (also gates `/api/docs`) |
| `apiplatform/protocols/graphql` | Public GraphQL | `/api/graphql` |
| `apiplatform/protocols/admin_graphql` | Admin GraphQL | `/api/admin/graphql` |
| `apiplatform/protocols/legacy_rest` | Legacy Magento 1 REST | `/api/rest/*` |
| `apiplatform/protocols/soap` | Legacy SOAP v1 | `/api/soap` |
| `apiplatform/protocols/v2_soap` | Legacy SOAP v2 | `/api/v2_soap` |
| `apiplatform/protocols/xmlrpc` | Legacy XML-RPC | `/api/xmlrpc` |
| `apiplatform/protocols/jsonrpc` | Legacy JSON-RPC | `/api/jsonrpc` |

### Enforcement
- **New protocols + `/api/docs`**: a Symfony `kernel.request` listener (`ProtocolToggleListener`, priority 200) returns 404 before the firewall or any controller runs.
- **Legacy SOAP / v2_SOAP / XML-RPC / JSON-RPC**: gated in `Maho_ApiPlatform_IndexController::forwardToLegacy()`.
- **Legacy REST**: gated at the top of `public/api.php` after `Mage::init()`.

A disabled protocol returns a clean 404 — indistinguishable from one that was never deployed. No 503s, no leakage of "disabled" status.

### Why
Security-by-default. The vast majority of Maho installs probably don't need SOAP/XML-RPC/JSON-RPC, and a fresh install shouldn't expose those endpoints by accident. New REST and GraphQL are also off by default — you turn on what you actually integrate with, nothing more.

### Migration note for upgraders
**This is a behaviour change for sites with active legacy API integrations.** A site upgrading from OpenMage with active SOAP integrations will see them break post-upgrade until `apiplatform/protocols/soap` (or whichever protocol they use) is set to `1`. We discussed and skipped a smart upgrade script that would auto-detect existing `api_user` records — explicit re-enable is the safer/clearer path.

### Helper API
```php
Mage::helper('apiplatform')->isProtocolEnabled(Maho_ApiPlatform_Helper_Data::PROTOCOL_REST_V2);
```

The redundant `apiplatform/general/enabled` flag (and `Helper::isEnabled()`) is removed — nothing read it.

### Tests
`tests/Pest.php` enables all protocols once at suite start (writes `core_config_data` + flushes config cache) so the existing API test suite continues to pass against the out-of-process web server.

### Verified
- `vendor/bin/php-cs-fixer fix --dry-run` clean
- `vendor/bin/phpstan analyze` `[OK] No errors`
- `php -l` clean on all modified files

mageaustralia:
Hit a small UX/observability issue with the GraphQL stack worth flagging.

When a GraphQL named query expects an IRI as its `id` argument (e.g. `orderOrder(id: ID!)`, `categoryCategory(id: ID!)`, `customerCustomer(id: ID!)`) and the caller passes a bare integer instead of a full IRI, ApiPlatform's `IriConverter::getResourceFromIri()` throws `InvalidArgumentException("No route matches '1'")`. Nothing in the GraphQL executor chain treats this as a client/input error, so it bubbles to the request handler and comes back as a generic `{"error":"internal_server_error","message":"An internal error occurred","code":500}`.

Reproducer (no auth needed — fails before the security check):

```bash
curl -X POST -H 'Content-Type: application/json' \
  --data '{"query":"{ orderOrder(id:1){ _id } }"}' \
  https://your-maho/api/graphql
```

Same shape on `categoryCategory(id:N)`, `customerCustomer(id:N)`, etc. Passing a real IRI works correctly:

```bash
curl -X POST -H 'Content-Type: application/json' \
  --data '{"query":"{ orderOrder(id:\"/api/rest/v2/orders/1\"){ _id } }"}' \
  https://your-maho/api/graphql
```

→ returns the expected 403 (auth check ran), proving the resource side is fine.

Two concerns:
1. Per the GraphQL spec, validation errors should land in the `errors[]` array with a 200 HTTP status and a useful `message` (e.g. "Invalid IRI '1' for argument 'id'"), not as an HTTP 500. Most consumers see "the server is broken" instead of "fix your input".
2. Real internal errors are now indistinguishable from user-input errors in production logs and HTTP traces — observability loss when something genuinely regresses.

Possible places this could land:
- ApiPlatform itself — extend the GraphQL `ErrorHandler` to translate `IriConverter`'s `InvalidArgumentException` to a proper field error. Probably the right home long-term.
- Maho-side EventListener under `Maho/ApiPlatform/symfony/EventListener/` that wraps the GraphQL request and converts the exception. Faster ship, scoped to Maho.
- Schema-level — make `id` arg use a custom type with format validation before the resolver runs.

Happy to put up a PR on the Maho-side listener path if that's the preferred direction — but checking first since you know the GraphQL stack better than I do.

(For context: surfaced while wiring a POS terminal that calls `orderOrder(id:1)` from JS. A wrapper-resource workaround sidesteps this for our use case, but the underlying bug applies to anyone who naively passes an int from a JS client.)

fballiano:
Hey @mageaustralia, does my last commit handle your issue as you were expecting?

mageaustralia:
Set up a fresh instance of this branch today (on a PostgreSQL backend) and ran into a few issues worth flagging:

**1. `/api/*` returns 500 on a cold cache - the router fails to build.** (backend-independent)
`Kernel::configureRoutes()` does:
```php
$routes->import('%kernel.project_dir%/Controller/', 'attribute');
```
Symfony's routing loader doesn't resolve the `%kernel.project_dir%` placeholder (the services loader in `configureContainer()` does, which is why that path works), so it looks for a literal `%kernel.project_dir%/Controller/` directory, fails, and the entire `/api` router (REST + GraphQL) 500s. On a fresh instance with no pre-warmed cache, every API request fails. Using `$this->getProjectDir() . '/Controller/'` fixes it.

**2. JWT secret isn't auto-generated on first boot.**
On a fresh instance `JwtService::getSecret()` throws ("JWT secret not configured" / "must be at least 32 characters"), so admin token generation fails and every secured endpoint returns 403 even for a logged-in admin. The docblock says the kernel auto-generates and persists a secret on first boot, but that didn't happen here; setting `apiplatform/oauth2/secret` manually worked around it.

**3. `api_user.api_key` is too small for the issued token.**
Creating an API key fails because the value exceeds `api_user.api_key varchar(40)`. On Postgres this is a hard error (`SQLSTATE 22001`); on MySQL it would be silently truncated (corrupting the key). The column needs widening.

**4. `security` expressions require `symfony/expression-language`.** (backend-independent)
Resources using `security: "is_granted(...)"` fail with *"The symfony/expression-language library must be installed to use the security attribute"*; it isn't declared as a dependency.

#1 and #4 are backend-independent; #2 and #3 surfaced on the Postgres backend. Happy to submit commits/PRs for these, or leave them with you, whichever you prefer.

mageaustralia:
One more from the same Postgres testing, and this one is project-wide (related to #3 above):

**Case-sensitive search on Postgres.** Maho's ORM `like` filter (`addAttributeToFilter(..., 'like' => ...)`) maps to `LIKE` in the Pgsql adapter's condition map:
```php
'like'  => '{{fieldName}} LIKE ?',
'nlike' => '{{fieldName}} NOT LIKE ?',
```
Postgres `LIKE` is case-sensitive, whereas MySQL's default-collation `LIKE` is not. So any case-insensitive search silently regresses on Postgres: e.g. searching `sunglasses` returns nothing, while `Sunglasses` matches "Aviator Sunglasses". This affects everything built on the `like` filter (quick/catalog search, admin grid filters, autocompletes, etc.), not a single feature.

Switching the Pgsql adapter's `like`/`nlike` map entries to `ILIKE` / `NOT ILIKE` restores MySQL-equivalent behaviour. Trade-off worth noting: `ILIKE` won't use a plain btree index, so a `lower()` functional index (or `citext`) may be worth it on hot paths.

hirale:
> Set up a fresh instance of this branch today (on a PostgreSQL backend) and ran into a few issues worth flagging:
> 
> **1. `/api/*` returns 500 on a cold cache - the router fails to build.** (backend-independent) `Kernel::configureRoutes()` does:
> 
> ```
> $routes->import('%kernel.project_dir%/Controller/', 'attribute');
> ```
> 
> Symfony's routing loader doesn't resolve the `%kernel.project_dir%` placeholder (the services loader in `configureContainer()` does, which is why that path works), so it looks for a literal `%kernel.project_dir%/Controller/` directory, fails, and the entire `/api` router (REST + GraphQL) 500s. On a fresh instance with no pre-warmed cache, every API request fails. Using `$this->getProjectDir() . '/Controller/'` fixes it.
> 
> **2. JWT secret isn't auto-generated on first boot.** On a fresh instance `JwtService::getSecret()` throws ("JWT secret not configured" / "must be at least 32 characters"), so admin token generation fails and every secured endpoint returns 403 even for a logged-in admin. The docblock says the kernel auto-generates and persists a secret on first boot, but that didn't happen here; setting `apiplatform/oauth2/secret` manually worked around it.
> 
> **3. `api_user.api_key` is too small for the issued token.** Creating an API key fails because the value exceeds `api_user.api_key varchar(40)`. On Postgres this is a hard error (`SQLSTATE 22001`); on MySQL it would be silently truncated (corrupting the key). The column needs widening.
> 
> **4. `security` expressions require `symfony/expression-language`.** (backend-independent) Resources using `security: "is_granted(...)"` fail with _"The symfony/expression-language library must be installed to use the security attribute"_; it isn't declared as a dependency.
> 
> #1 and #4 are backend-independent; #2 and #3 surfaced on the Postgres backend. Happy to submit commits/PRs for these, or leave them with you, whichever you prefer.

Hi, @mageaustralia 
### The sequel to #4: `security.expression_language` gets pruned even when the library *is* installed

Your #4 is the first half of this. Once `symfony/expression-language` is declared (which clears the "must be installed" error), a second failure shows up on any resource using `security: "is_granted(...)"`:

- api-platform aliases `api_platform.security.expression_language` → `security.expression_language` and injects it into `ResourceAccessChecker` via `nullOnInvalid()` — a *soft* reference (see api-platform's `config/security.php`).
- In a kernel that doesn't register `access_control` / the expression voter, `ResourceAccessChecker` is the **only** consumer of `security.expression_language`. With no hard reference, `RemoveUnusedDefinitionsPass` drops SecurityBundle's (private) `security.expression_language` and resolves the argument to `null`.
- `ResourceAccessChecker::isGranted()` guards that null (throws the "must be installed" `LogicException`), **but `ResourceAccessChecker::usesObjectVariable()` does not** — it calls `$this->expressionLanguage->parse(...)` directly. `AccessCheckerProvider` calls `usesObjectVariable()` for every operation carrying a `security:` expression, so each one 500s with "call to a member function parse() on null". The prune is baked into the compiled container, so it fails on every request, not just a cold cache.

Two layers to it:

1. **Functional fix — belongs in the kernel (it's the container doing the pruning).** Anchor the existing service so the removal pass can't drop it. A small compiler pass is enough, and it preserves SecurityBundle's expression cache — re-declaring the service would drop its `cache.security_expression_language` argument:

   ```php
   // Kernel::build()  (CompilerPassInterface + PassConfig imported)
   $container->addCompilerPass(
       new class implements CompilerPassInterface {
           public function process(ContainerBuilder $container): void
           {
               if ($container->hasDefinition('security.expression_language')) {
                   $container->getDefinition('security.expression_language')->setPublic(true);
               }
           }
       },
       PassConfig::TYPE_BEFORE_REMOVING,
   );
   ```

2. **Upstream robustness (api-platform/core).** `ResourceAccessChecker::usesObjectVariable()` arguably should null-guard the way `isGranted()` does, so a missing/pruned expression language degrades gracefully instead of fatally. Filed upstream as api-platform/core#8215 (with a one-line fix proposed) — orthogonal to this PR.

fballiano:
@mageaustralia @hirale I should have addressed all the remaining points

fballiano:
## Code review pass — 35 fixes (commit e73a21df53)

I ran an in-depth, multi-pass review of the new REST + GraphQL layer (10 review cycles, 3 reviewers each: logic / security / conventions). Found and fixed **35 issues across 34 files**. Everything passes **PHPStan level 6** and **PHP-CS-Fixer** clean. Summary below.

### 🔒 Security
- **Idempotency TOCTOU** (`IdempotencyListener`): two concurrent requests on an expired key could both reclaim it and double-execute. Added a `created_at < cutoff` guard to the reclaim `UPDATE` so only one wins.
- **JWT type check** (`JwtService`): an `assert($parsed instanceof Plain)` is a no-op when `assert.active=0` (typical in prod), letting a non-`Plain` token skip the `SignedWith` constraint. Replaced with a hard `instanceof` guard that throws.
- **CORS** (`Kernel`): `allow_headers` omitted `X-Idempotency-Key`, so browser preflight stripped it and silently disabled cross-origin replay protection. Added to both arrays.
- **Path traversal** (`Cms/MediaProcessor`, `Cms/MediaProvider`): containment checks used a bare `str_starts_with` prefix match, so a sibling dir (`<root>_other`) passed. Switched to a trailing-separator boundary.
- **SSRF / DNS-rebinding** (`ProductMediaProcessor`): hardened the remote-image fetch by reconstructing the URL from parsed parts pinned to the validated IP (closes the TOCTOU window).
- **Token forgery** (`AccountTokenService`): pipe-delimited HMAC payload let an email containing `|` shift parse offsets. Switched to a JSON payload.
- **Guest order token** (`OrderProvider`): the one-time access token wasn't consumed on the GraphQL path (replayable indefinitely) and was echoed back in the response. Now consumed + omitted, matching REST.
- **State leak** (`ProductQueryHandler`): GraphQL search used `getSingleton()` for catalog/search layers, leaking prior-request query/category text across FPM workers. Switched to per-request models.

### 🐛 Correctness
- **Partial-update reset** (the biggest class): several DTO flag/store fields defaulted to a concrete value (`true`/`false`/`[0]`) and were written unconditionally, so a partial `PUT` omitting them **silently reset** the model. Made nullable + create-only defaults: `manage_stock` (Product/StockUpdate), category `isActive`/`includeInMenu`, CMS page/block `isActive`/`stores`, blog post `isActive`/`stores`, and customer default-billing/shipping pointers.
- **Cart totals**: reset the totals-collected flag before `collectTotals()` in the gift-card and merge-cart mutations; use the billing address for virtual-cart discount/tax totals (they don't live on the shipping address).
- **Order placement** (`OrderService`/`OrderProvider`): preserve the original domain exception message instead of re-wrapping it generically; build the guest DTO before consuming its token.
- **Customer**: guard a null billing address in create-account-from-order; bridge the REST body into forgot/reset-password handlers (were GraphQL-only → always 400 over REST); fix a camelCase GraphQL arg-key mismatch (`firstName` vs `firstname`) that silently dropped name updates; defer the email save into the address try-block to avoid a partial commit; cap GraphQL order-list pagination.
- **Money**: default to the store's current currency instead of a hardcoded `'AUD'`.
- **StoreConfig**: dispatch the `dto_build` event before caching so observer mutations are persisted.

### 📐 Conventions
- Replaced native `json_decode`/`json_encode` on **user input** with `Mage::helper('core')->jsonDecode/jsonEncode`, returning HTTP 400 on malformed JSON (`ContactFormProcessor`, `OrderProcessor`, `ProductProvider`, `Kernel`, `ModuleApiDiscovery`).

### Notes / out of scope
- The five recurring bug classes (partial-update reset, REST↔GraphQL auth/token parity, unsafe JSON on user input, FS traversal boundaries, GraphQL arg-casing) were each swept exhaustively and confirmed converged in the final cycle.
- Left as-is deliberately: the custom-option `File.php` `secret_key` derivation (pre-existing core pattern, mirrored correctly), `public/rest.php` pre-boot `json_encode` (Mage helper unavailable before init), and files touched only by merge commits.
- Pre-existing: the `tests/Api/V2/Write/CmsPageTest` suite fails identically on a clean baseline (API routes not compiled in the test env) — a test-infra condition, not introduced here.

fballiano:
## In-depth review — 36 issues found and fixed

Ran a multi-pass review (7 cycles, three independent reviewers per cycle: logic / security / conventions) over the REST + GraphQL API surface. **36 issues fixed** in 30 files; the final cycle returned zero valid findings. Pushed as `2d8c657977`. PHPStan (level 6) and php-cs-fixer both pass clean.

### Critical — auth & request-breaking
- **DefaultDenyListener** — auth bypass via a junk `Bearer` header, plus a double-registration that ran before the firewall and 401'd every authenticated REST request.
- **IdempotencyListener** — ran before the firewall, so the user scope was always null and idempotency was silently disabled for all requests.
- **JwtService** — nullable `resource_id` triggered a `TypeError` that 500'd every `client_credentials`/`api_user` login; also a first-install secret-generation race across workers.
- **OAuth2Authenticator** — tokens missing a `jti` skipped the revocation/blacklist check.
- **AuthenticationTrait** — admins were spuriously 403'd on every REST write.
- **GraphQL placeOrder** — missing shipping-method validation (allowed fabricated/free shipping).

### Security & access control
- Cart-merge IDOR; login merge re-implemented with a weaker ownership guard → routed through `CartService::mergeCarts()`.
- Newsletter subscribe IDOR; reset-password / guest-order token enumeration via distinct error messages.
- Public **coupon-validate** and **GraphQL login** endpoints had no-op rate limiters (missing `system/rate_limit` defaults) — coupon/credential enumeration was unthrottled.
- Information leaks: plaintext `api_key` flashed to admin session (`UserController`), raw SQL in delete errors (`ModelPersistenceTrait`), and raw exception messages in admin GraphQL responses (now gated behind `showDebug()` + logged).
- Missing form-key on the gift-card balance `save` action.

### Logic & data integrity
- **OrderProcessor** — quote saved *before* shipping-method validation, persisting a bogus method on failure.
- `Money` — hardcoded 5-currency table → `Locale::formatCurrency()`.
- `ProductCustomOptionProcessor` — Pgsql `ON CONFLICT` corruption (identity columns in update list) + orphaned option rows on type change.
- `BundleOptionProcessor` — stale product registry feeding the wrong store scope.
- `WishlistProcessor` — disabled products re-added as duplicates.
- `InvoiceProvider` — API users wrongly 403'd on invoices/PDF.
- `CreditMemoProcessor` — comment saved after commit, re-triggering post-save observers on a refunded memo.
- `ProductQueryHandler` — category depth computed against the absolute tree level instead of the parent's.
- Plus fixes in `CartMapper`, `AccountTokenService` (HKDF-scoped signing key), `ProductMediaProcessor`, and `CustomerProcessor`.

### Verified clean
Guest-cart masked-ID logic (`CartService`), HTTP cache tiers / cross-store CDN poisoning (`HttpCacheListener`), store-scope authorization, the generic Crud provider/processor path, order placement (`GET_LOCK` serialization), token blacklist (durable, fail-closed), and the Symfony infrastructure / base classes / schema + config XML. Documented design choices (admin/API-user reads intentionally not per-store filtered; early-boot native `json_encode` in `rest.php`) were confirmed and excluded.

fballiano:
## Standardized money representation across the API

Cleaned up how monetary values are represented so REST and GraphQL behave identically. Pushed as `8c200f318c` + `201fc5a2b9`.

### The rule
**Money is always a plain number, and every response names its `currency` code exactly once.**

No localized "formatted" strings are returned over the API; clients format the numeric value with the currency code in their own locale.

### Why
- Each cart, order, product, invoice, or credit memo response is single-currency, so attaching a currency to every individual amount (`grandTotal`, `subtotal`, `taxAmount`, …) would just repeat the same code 6+ times per response for no added information.
- The DTOs are shared between REST and the auto-generated GraphQL operations, so keeping money as plain numeric fields keeps both protocols consistent with no special-casing. The only divergence was hand-rolled objects in the custom GraphQL resolvers, which is what this fixes.
- Plain numbers are the idiomatic JSON shape and match the existing price fields, so nothing in the numeric-price contract changes.

### Changes
- Removed the unused `Money` value object class (it was never instantiated; GraphQL used ad-hoc arrays, the price DTOs used bare floats).
- Dropped the `{ value, formatted }` money objects from the GraphQL cart, order, and product resolvers; they now return plain numbers like the DTOs.
- Added a single `currency` field to the read DTOs that lacked one (`Product`, `Invoice`, `CreditMemo`); `Cart` and `Order` already had it.
- Added `currency` to the standalone GraphQL payloads that have no enclosing cart/order object (place order, credit memo, order summary, gift card balance, available shipping methods).

Net effect: a client reads `currency` once and every money value under it is a plain number, identical in REST and GraphQL. PHPStan (level 6) and php-cs-fixer pass clean.

fballiano:
## Code review — fixes applied (commit 006573e)

Ran a multi-reviewer pass (security / logic / conventions) over the API Platform layer, verified each finding against the code, and pushed fixes for the genuine ones.

### Security — authorization & rate limiting
The Resource `security` expressions only gate `ROLE_ADMIN or ROLE_API_USER`, so the registered **granular** API-user permissions were never enforced on several write paths — an api-user token with only `products/read` could perform privileged writes. Added the missing per-operation checks (admin tokens bypass via early-return, so admin flows are unaffected):

- `CreditMemoProcessor` → `credit-memos/create`
- `ShipmentProcessor` → `shipments/create`
- `StockUpdateProcessor` → `inventory/write`
- `GiftCardProcessor` → `giftcards/create` (create) + `giftcards/write` (balance adjust)
- `AuthTokenProcessor` — added IP rate limiting to the token **refresh** endpoint (the issue endpoint already had it)
- `CartService::merge` — hardened the guest-cart ownership guard so a non-guest cart with an unset/zero `customer_id` can no longer be absorbed into another account (IDOR)

### Conventions
- Replaced raw `json_encode`/`json_decode` with `Mage::helper('core')->jsonEncode/jsonDecode` across the new API providers and cart services (`LayeredFilterProvider`, `ProductProvider`, `CartMapper`, `CartMutationHandler`, `CartService`), per the AGENTS.md JSON-handling convention.

### Verification
- phpstan (level 6), php-cs-fixer, and rector all clean on the 10 changed files.
- No tests affected — the write tests authenticate with admin tokens, which bypass the new per-permission checks.

### Reviewed and intentionally not changed (false positives)
A number of flagged "logic" issues did not hold up under inspection, notably:
- `ProductProcessor` categoryIds `!== []` → `!== null`: **rejected** — the DTO field is non-nullable (`array = []`), so that change would wipe all category assignments on any update omitting `categoryIds` (data loss). The current guard is correct.
- `ProductTierPriceProvider` "operator precedence bug": the `(int)` cast already applies to the parenthesized `??`, so it's correct.
- `ApiExceptionListener` 401/403 "inversion": expired/invalid tokens are handled earlier as authentication errors; 403 at that block is correct.
- `GraphQlPermissionListener` exception catch: `jsonDecode` throws native `\JsonException` (via `JSON_THROW_ON_ERROR`), so the catch is right.
- `File.php` required-option "bypass": the `$_FILES` fallthrough already enforces `getIsRequire()`.
- `ProductMediaProcessor` IPv6 SSRF and `StoreContextListener` query/header precedence: not exploitable as described.

fballiano:
## In-depth review pass — fixes pushed (`8839b39`)

Reviewed the API platform across logic, security, and conventions. Every finding was verified against the actual code before acting (several reviewer reports turned out to be false positives on closer reading). Pushed fixes for the confirmed issues:

### Correctness
- **Product write fields were silently ignored.** The `Post`/`Put` operations declare a `normalizationContext` but **no `denormalizationContext`**, so writes are ungrouped and API Platform deserializes every property *except* those marked `writable: false`. That meant `isActive`, `visibility`, `stockQty`, `barcode`, and `categoryIds` never populated from the request body — the processor logic reading them was dead code, and `POST/PUT /products` quietly dropped those fields. Removed `writable: false` from exactly those five inputs (verified against the fields the processor consumes; `status`/`createdAt`/`updatedAt` stay read-only).
- **Auth refresh** now issues the new JWT before revoking the old one, so a generation failure can't strand the customer with no valid token.
- **`getCustomerCart`** is now scoped to the current store (`setSharedStoreIds`), matching the token-grant path — avoids loading another store's cart in multi-store setups.
- **Customer quick-create** uses the store's default country instead of a hardcoded `AU`.

### Security
- **Idempotency** now excludes all `/auth/` endpoints (previously only `/auth/token`), so `/auth/refresh` and `/auth/logout` responses — which carry JWT material — are never stored or replayed, and rotation/revocation always executes.
- **GraphQL permission analysis** guards `collectTopLevelFields` against cyclic fragment spreads (`A → B → A`), which `Parser::parse()` accepts and which would exhaust the stack before validation runs.
- **GraphQL `DELETE_PREFIXES`** added so destructive mutations require a resource's `delete` permission *when it defines one*, mirroring REST's `ApiUserVoter::resolveOperation` (falls back to `write` otherwise — consistent and future-proof).
- **Cart merge** rejects a guest cart owned by a different customer regardless of the `customer_is_guest` flag.
- **Product media SSRF guard** explicitly rejects private/reserved IPv6 literals (previously only blocked as an IPv4-resolver side effect).

### Conventions
- **Rate limiting** no longer reads the client IP in the trait; it resolves identity through core via `RateLimitScope::Ip`, honoring "core owns request identity." Verified no bucket collisions across all 12 call sites; behavior is preserved (and slightly safer — the old unknown-IP fail-open was effectively a bypass).

### Verification
`composer lint` clean (PHP-CS-Fixer, Rector, PHPStan L6). Backend integration tests pass (`RateLimitDefaultsTest`, `ResourceGraphQlSecurityTest`). `dump-autoload` recompiles the API permission registry cleanly.

⚠️ **One item to confirm in CI/a server env:** the product-write fix should be validated by `tests/Api/V2/Write/ProductWriteTest.php` (esp. `isActive:false → status:'disabled'`, category/visibility assertions). I couldn't run the live API suite locally — the dev server returned `301` trailing-slash redirects on every endpoint (incl. unmodified GET reads) and the test helper uses `follow_location => false`, so all cases failed on the redirect rather than on the code.

fballiano:
## In-depth review pass — 18 issues fixed (commit 142a515)

Ran a multi-cycle review (logic / security / conventions) over the non-test production code. Each finding was verified against the actual code before fixing. All changed PHP files pass PHPStan (level 6) and php-cs-fixer.

### Auth & security
- **client-credentials & api-user grants** (`AuthTokenProcessor`): both leaked valid `client_id`/username existence via distinct error messages + timing (hash check skipped on miss). Now constant-time with a dummy bcrypt hash and a uniform error.
- **API activity log** (`ActivityLogTrait`): wrong key `api_user_id` → `consumer_id`; the API user identity was silently dropped from every logged write.
- **Contact form** (`ContactFormProcessor`): email-keyed rate limit only — a single client could rotate emails to spam the owner / bomb third parties via auto-reply. Added an IP-scoped cap.
- **.htaccess**: TRACE/TRACK 405 block sat after the `[L]` `/api/*` rewrites, so those methods reached `rest.php`. Moved before all rewrites.

### Logic & correctness
- `CartService::mergeCarts()` — null guard for invalid masked IDs.
- `OrderService` — gift-card revalidation moved inside the placement `GET_LOCK`.
- `CartProcessor` & GraphQL `CartMutationHandler` — configurable variants added by child SKU never got their fulfillment type set (parent-SKU mismatch); now uses ID-snapshot new-item detection.
- `CartMutationHandler` — null-deref on gift-card apply/remove cart reload; and gift-card removal left a stale discount (now reuses `CartService::removeGiftcard`).
- `CrudProcessor` — fall back to the in-memory model when a post-save reload returns empty.
- `StockUpdateProcessor` — direct-SQL stock writes never refreshed `cataloginventory_stock_status`; added reindex in single + bulk paths.
- `GroupedProductLinkProcessor` — DELETE silently returned success on an unparseable child id; added a positive-id guard (matches `ConfigurableSetupProcessor`).

### Conventions
- `Mage::log()` levels, `declare(strict_types=1)` placement, `ADMIN_RESOURCE` const in place of legacy `_isAllowed()`, Maho SPDX line, and a comment on the pre-bootstrap `json_encode`.

### Reviewed and intentionally left as-is
- `/auth/refresh` requiring a non-expired token (the safer design); admin/API tokens using the guest group in catalog cache keys (deterministic, no cross-group contamination); the unconfirmed-account login message (pre-existing storefront-consistent UX tradeoff); `StoreConfigProvider` re-dispatching its event on cache hits (required so observers run on every request); `StoreContext` loose `in_array` (not a bug under PHP 8.3).

fballiano:
## In-depth review pass — fixes pushed (`6648a70`)

Ran a multi-cycle review (parallel logic / security / conventions reviewers, each finding adversarially verified). Conventions came back clean; the actionable logic & security findings are now fixed. Quality gate green: **PHP-CS-Fixer**, **PHPStan (level 6)**, **Rector**, XML, and locale-CSV sorting all pass.

### Concurrency (CRITICAL)
- **Double-refund race** — credit-memo creation (REST `CreditMemoProcessor` + GraphQL `OrderMutationHandler`) now runs inside a per-order `GET_LOCK` with an under-lock `canCreditmemo()` re-check.
- **Duplicate-cart race** — `CartService::getCustomerCart()` serializes load-or-create with a per-customer `GET_LOCK`.

### Security
- **SSRF / TLS** — image-URL upload pins TLS verification to the original hostname (`verify_peer` + `peer_name`) while staying IP-pinned with redirects disabled; HTTPS source images now validate correctly instead of silently failing.
- **Upload hardening** — temp-file extension allowlist (incl. AVIF) applied before the magic-byte check (which also now accepts AVIF); temp file cleaned up on rename failure.
- **Path traversal** — `..` segments stripped from the CMS media `folder` param before the realpath boundary checks.
- **Token freshness** — `api_user` `allowed_store_ids` is re-read from the DB rather than trusted from the JWT claim, so store-scope changes apply immediately.
- **Rate limiting** — public guest-order-token lookup (REST + GraphQL) is throttled per-IP via a new `system/rate_limit/guest_order_lookup` setting (default 30/hour).

### Correctness
- Masked cart IDs normalized to lowercase at the lookup chokepoint (collation-independent).
- POS dummy emails use `random_bytes` (no collision 500s); concurrent-registration unique-constraint failures return a clean "already exists" error.
- GraphQL login surfaces unconfirmed-email as a distinct 403 (matching REST).
- Fulfillment type matched by product id (child→parent) on configurable cart merges.

### Conventions
- Class-docblock summaries given trailing periods across new API files; removed a stray `@package` tag from `lib/Maho/Config/ApiResource.php`.

A few flagged items were verified as already-mitigated or by-design and intentionally left unchanged: the fake POS `freeshipping` method (rejected by `placeOrder` rate re-validation), guest-token one-time-use, the pre-place `collectTotals()->save()` (guarded by the under-lock `is_active` re-check), and `MAHO_STORE_ID` (admins already have all-store access; gated by the admin-ID HMAC).

fballiano:
Updated after maintainer review. Several of the original fixes were reverted as out of scope or not matching codebase conventions; the branch now contains only the changes below. All quality gates green: PHPStan level 6 (no errors), php-cs-fixer, rector.

**Kept**

Critical
- Coupon writes were ungated for scoped API users on REST and GraphQL (root cause: `Coupon` used the vanilla `ApiResource` so `coupons` was never registered as a permission). Now registered with read/create/write/delete and enforced in the processor.
- Newsletter `item_query` allowed any logged-in customer and fell through to an ownership-less provider, exposing every subscriber's email and customer id. Restricted to admin/API users; customers keep `newsletterStatus`.
- GraphQL permission listener only gated POST, so query-string GraphQL over GET bypassed per-resource checks. Now gates GET and POST.

Warning
- Newsletter subscribe lacked an IP-level rate limit (the per-email bucket was trivially bypassed).
- `X-Order-Token` missing from CORS allow_headers, breaking cross-origin guest order confirmation.
- Clarified the refresh-token error for non-customer tokens; removed a redundant legacy admin router XML block.

**Reverted (after review, no longer on the branch)**
- Per-order lock around shipment creation: serializing concurrent operations is not the codebase convention, and the raw `GET_LOCK` was MySQL-only.
- Inactive-store guard in `StoreContextListener`: it ran before auth and would have blocked admin API callers from targeting inactive stores.
- `imageUrl` product-media 10MB download cap.
- Gift card balance currency conversion in the cart GraphQL handler.
- PostgreSQL adapter `raw_query` error-path change.

**Still open for maintainer decision (unchanged)**
1. JWT signing secret stored plaintext in `core_config_data`. Encrypting at rest would close DB-only exposure but re-couples to the crypt key (deliberately avoided here) and needs a rotation migration that invalidates outstanding tokens.
2. Login returns 403 "account not confirmed" before checking the password, enabling single-request enumeration of unconfirmed accounts. This mirrors the existing storefront login behaviour, so any change should be platform-wide.

Note: four pre-existing lock sites in this PR (`CreditMemoProcessor`, `OrderService`, `OrderMutationHandler`, `CartService`) use the same MySQL-only raw `GET_LOCK` and will not run on the PostgreSQL/SQLite adapters.

fballiano:
All the documentation is now at https://mahocommerce.com/api/v2/

fballiano:
## Review round — fixes pushed (`20cae47`…`9c70976`)

Ran another multi-angle review pass (8 parallel finder angles — line-by-line, removed-behavior, cross-file tracing, reuse, simplification, efficiency, altitude, conventions — with every candidate independently re-verified). 17 candidates → 11 held up; all fixed. Quality gates green locally: PHP-CS-Fixer, Rector, PHPStan level 6, Api suite (565 tests / 1890 assertions), Backend suite.

### Fixed
**Performance (N+1s on list endpoints)**
- Order lists loaded a payment collection per order; shipment list-all loaded a full order + track collection + item collection per shipment (~3 queries/row). Both now batch-preload per page following the existing `_preloaded_items` pattern.

**Correctness / consistency**
- `searchProducts` (GraphQL) ignored the requested `storeId` while `getCategories` honored it — store-scoped pricing/visibility/name overrides came from the default store.
- Cart vs place-order address paths had drifted: duplicated camelCase→snake_case mapping, reversed region code/name lookup precedence, normalization on one path only, and two cart endpoints not resolving region at all. Single source of truth now: `CartService::mapAddressInput()`.
- A request killed mid-flight (OOM, fatal, max_execution_time) pinned its idempotency key as 409 for the full 24h TTL. Abandoned in-progress reservations are now reclaimable after 10 minutes (Stripe-style lock window); fresh ones still 409.
- `API_STORE_CONFIG` (24h TTL) and `API_COUNTRIES` (1h) had no invalidation — config edits served stale API data until expiry. Config/store saves now clean both tags; `API_LAYERED_FILTERS` registered on the `api_data` cache type.

**Cleanup**
- Removed never-thrown `ConflictException`/`AuthorizationException` and four uncalled `StoreContext` statics; blog providers reuse `singleItemPaginator()`.

### Reviewed, intentionally unchanged
- Legacy SOAP/XML-RPC default-off on upgrade (deliberate default-deny posture, documented in system config) — existing integrators upgrading should be warned in release notes.
- Gift cards with a blank expiration are now never-expiring (matches the form's "Leave empty for no expiration" note; the old default-lifetime behavior contradicted it).

Refuted during verification (no action): admin-session CSRF (fail-closed Origin check covers the only cookie-authenticated endpoint), cart currency fallback (unreachable — quote always persists its currency), Postgres ILIKE change (MySQL-parity-correct).

mageaustralia:
Congrats on getting this merged, Fabrizio. I never quite imagined it was going to be 500 commits to get it completed when started. Really happy to finally see this land, though.

PR #824, opened 2026-04-12
Changed default observer instantiation from `singleton` to `model`

### LINKED ISSUE #818: Default Observer instantiation to new instance and remove type parameter
## Summary

The `type` parameter on `#[Maho\Config\Observer]` (which toggles between `singleton` and `model` instantiation) should be removed, with observers always getting a fresh instance per dispatch.

## Rationale

- A fresh instance per dispatch is the safer default — no risk of leaked state between event handlers
- Several observer classes extend `DataObject` or `Mage_Core_Model_Abstract` which carry mutable state, making shared instances a subtle bug vector
- None of the ~20 current `type: model` usages actually depend on fresh-instance semantics — but none of the `singleton` usages depend on shared state either
- The singleton optimization (avoiding repeated instantiation) is negligible
- There is no compelling use case for wanting shared state across dispatches of the same observer

## Changes needed

- Remove `type` parameter from `lib/Maho/Config/Observer.php`
- Update dispatch logic to always instantiate a new instance (equivalent of current `model` behavior)
- Remove `type: model` from the ~20 observer declarations that currently specify it
- Update compilation logic if it handles `type`
- Run `composer dump-autoload`

fballiano:
## Summary

- Changed the default `type` parameter on `#[Maho\Config\Observer]` from `'singleton'` (shared instance) to `'model'` (fresh instance per dispatch)
- Updated dispatch logic in `App.php` and `Config.php` to default to `getModel()` instead of `getSingleton()`
- Added explicit `type: 'singleton'` to the 3 observer classes that rely on shared state across dispatches: `AdminActivityLog`, `CatalogRule`, `CatalogInventory`
- Removed now-redundant `type: 'model'` from ~21 observer declarations that were explicitly setting what is now the default

## Rationale

A fresh instance per dispatch is the safer default — no risk of leaked state between event handlers. Several observer classes extend `DataObject` or `Mage_Core_Model_Abstract` which carry mutable state, making shared instances a subtle bug vector.

An audit of all 54 observer classes confirmed only 3 rely on singleton state across dispatches, and those are now explicitly opted in.

## Test plan

- [x] `composer dump-autoload` compiles successfully
- [x] Compiled attributes show 199 model / 44 singleton — singletons only from the 3 expected classes
- [x] `vendor/bin/php-cs-fixer fix --dry-run` — 0 issues
- [x] `vendor/bin/phpstan analyze` — 0 errors
- [x] `composer test` — 1845 tests passed

Closes #818

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Compare/Item.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Compare/Item.php
@@ -161,7 +161,7 @@ class Mage_Catalog_Model_Product_Compare_Item extends Mage_Core_Model_Abstract
      *
      * @return $this
      */
-    #[Maho\Config\Observer('customer_login', area: 'frontend', type: 'model')]
+    #[Maho\Config\Observer('customer_login', area: 'frontend')]
     public function bindCustomerLogin()
     {
         $this->_getResource()->updateCustomerFromVisitor($this);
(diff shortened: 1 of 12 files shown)


fballiano:
This is a better default for platform safety, so I'll merge this

PR #826, opened 2026-04-13
Stub empty summary on zero-review products to eliminate N+1 queries

### LINKED ISSUE #825: Perf: Mage_Review_Model_Review::appendSummary fires N queries per zero-review product on category pages
## Problem

`Mage_Review_Model_Review::appendSummary()` (core/Mage/Review/Model/Review.php:170) is advertised as a batched review-summary loader for product collections. It batches via `addEntityFilter($entityIds)` — but only for products that **have** review summary rows. Zero-review products never get `rating_summary` populated by the batch, which makes every subsequent consumer fall through a slow-path single-PK lookup.

The silent fallback lives in `Mage_Review_Block_Helper::getSummaryHtml()`:

```php
if (!$product->getRatingSummary()) {
    Mage::getModel('review/review')
        ->getEntitySummary($product, Mage::app()->getStore()->getId());
}
```

`getEntitySummary()` calls `$product->getReviewSummary()`, which in `Mage_Catalog_Model_Product::getReviewSummary()` does:

```php
$this->_reviewSummary[$storeId] = Mage::getModel('review/review_summary')
    ->setStoreId($storeId)
    ->load($this->getId());
```

Every zero-review product in the grid triggers one `SELECT * FROM review_entity_summary WHERE entity_pk_value = N` that returns zero rows — an \`AFF: 0\` query — for every product, every render.

## Impact observed

A typical category page  with 36 products in the grid. Query log before any fix:

\`\`\`
1009× SELECT \`review_entity_summary\`.* FROM \`review_entity_summary\` WHERE ((\`review_entity_summary\`.\`entity_pk_value\`=N)) AND ((store_id = 1))
\`\`\`

36 products × ~28 `Block/Helper::getSummaryHtml()` calls per product (product card renders the block in several templates: grid, alternatives block, mega-menu featured, etc.) ≈ 1008 wasted queries returning AFF: 0.

Each query is <1ms so it's invisible in timing profilers — but that's 1000+ round-trips to MySQL per page render. At concurrency this chews through the connection pool and DB buffer cache unnecessarily.

This hits **every Maho install** that has any zero-review product appearing alongside reviewed products in a grid — which is effectively every real store.

## Recommended patch

In `Mage_Review_Model_Review::appendSummary()`, after the existing `foreach` that populates reviewed items, stub an empty Summary model on every product that wasn't matched. Consumers check truthiness of \`rating_summary\`, so an empty Summary (no \`rating_summary\` value, no \`reviews_count\`) behaves identically for display — but `hasData('rating_summary')` is now true, so `Block/Helper` skips the fallback.

\`\`\`php
public function appendSummary(\$collection)
{
    \$entityIds = [];
    foreach (\$collection->getItems() as \$item) {
        \$entityIds[] = \$item->getId();
    }

    if (!count(\$entityIds)) {
        return \$this;
    }

    \$summaryData = Mage::getResourceModel('review/review_summary_collection')
        ->addEntityFilter(\$entityIds)
        ->addStoreFilter(Mage::app()->getStore()->getId())
        ->load();

    /** @var Mage_Review_Model_Review_Summary \$summary */
    foreach (\$summaryData as \$summary) {
        if ((\$item = \$collection->getItemById(\$summary->getEntityPkValue()))) {
            \$item->setRatingSummary(\$summary);
        }
    }

    // Products with zero reviews aren't returned by the batch, so stub an
    // empty Summary on them. Otherwise Mage_Review_Block_Helper falls back
    // to getEntitySummary() and fires a single-row SELECT per product (AFF: 0)
    // for every zero-review product in the grid.
    \$emptySummary = Mage::getModel('review/review_summary');
    foreach (\$collection->getItems() as \$item) {
        if (!\$item->hasData('rating_summary')) {
            \$item->setRatingSummary(\$emptySummary);
        }
    }

    return \$this;
}
\`\`\`

## Verification

On the same page after applying the patch: \`review_entity_summary\` query count drops from **1009 → 1** (just the batch SELECT with IN-list). Total page query count dropped from **6884 → 533** in one change (other fixes we made accounted for the rest). No visible behavioural change — zero-review products still render without a rating badge, identical to before.

## Notes

- The empty `Mage::getModel('review/review_summary')` created outside the loop is shared by reference across all unmatched items. That's fine — consumers only read from it; they don't mutate.
- An even cleaner fix would be to check `hasData('rating_summary')` in `Mage_Review_Block_Helper::getSummaryHtml()` instead, but the current approach keeps the fix localised to the batch loader (one place) rather than every caller.
- Found while profiling category pages on a ~30k-product Maho dev install.

fballiano:
## Summary

- Fixes #825
- `appendSummary()` now stubs an empty `review/review_summary` model on products not matched by the batch query
- Prevents `Mage_Review_Block_Helper::getSummaryHtml()` from firing individual SELECT queries for each zero-review product (~1000 redundant queries on a typical 36-product category page)

## Test plan

- [ ] Load a category page containing products with and without reviews
- [ ] Verify zero-review products still render without a rating badge (no visual change)
- [ ] Verify reviewed products still display correct rating summaries
- [ ] Confirm query log no longer shows per-product `review_entity_summary` SELECT queries for zero-review products

### DIFF
--- a/app/code/core/Mage/Review/Model/Review.php
+++ b/app/code/core/Mage/Review/Model/Review.php
@@ -190,6 +190,13 @@ class Mage_Review_Model_Review extends Mage_Core_Model_Abstract
             }
         }
 
+        $emptySummary = Mage::getModel('review/review_summary');
+        foreach ($collection->getItems() as $item) {
+            if (!$item->hasData('rating_summary')) {
+                $item->setRatingSummary($emptySummary);
+            }
+        }
+
         return $this;
     }
 
--- a/app/code/core/Mage/Review/Model/Review.php
+++ b/app/code/core/Mage/Review/Model/Review.php
@@ -190,10 +190,9 @@ class Mage_Review_Model_Review extends Mage_Core_Model_Abstract
             }
         }
 
-        $emptySummary = Mage::getModel('review/review_summary');
         foreach ($collection->getItems() as $item) {
             if (!$item->hasData('rating_summary')) {
-                $item->setRatingSummary($emptySummary);
+                $item->setRatingSummary(Mage::getModel('review/review_summary'));
             }
         }
 


fballiano:
@mageaustralia does this work for you? :-)

PR #827, opened 2026-04-13
Consolidate 'adminhtml' config area into 'admin'

### LINKED ISSUE #814: Consolidate 'adminhtml' config area into 'admin'
## Summary

The codebase has two confusingly named top-level config areas: `<admin>` and `<adminhtml>`. After removing the dead `AREA_ADMIN` constant in #813, the `<admin>` section still exists in ~25 config.xml files and is actively used for routing, fieldsets, and other configuration. This proposes fully merging `<adminhtml>` into `<admin>` to eliminate the confusion.

`admin` is the better name — it's shorter, cleaner, and the `html` suffix is a Magento-era artifact that adds no clarity.

**Note:** #813 removed the `AREA_ADMIN` constant because it was dead code at the time. This change would reintroduce `AREA_ADMIN` as the *primary* constant, and `AREA_ADMINHTML` would become deprecated instead.

## Current state

The `<admin>` top-level config section is used for:
- **Routers** (23 modules): Every module that adds admin controllers defines `<admin><routers>` — this is how the admin router discovers which modules handle which URLs
- **Fieldsets** (4 modules): `Bundle`, `Catalog`, `Customer`, `Downloadable` define `<admin><fieldsets>` for dataflow field mapping
- **Design/web config**: `Adminhtml` and `Core` modules define admin design and web settings under `<admin>`

The `<adminhtml>` section is used for events, layout updates, and translate config — the things that are actually loaded at runtime via `loadArea('adminhtml')`.

The router system bridges the two: in `Mage/Core/etc/config.xml`, the admin router is configured with `<area>admin</area>`, which tells `collectRoutes()` to read from `admin/routers`. Meanwhile, admin controllers load the `adminhtml` area for events/layout/design.

## Proposed change

1. Reintroduce `AREA_ADMIN` as the primary area constant; deprecate `AREA_ADMINHTML`
2. Move all `<adminhtml>` content (events, layout updates, translate config) into `<admin>` across all config.xml files
3. Update PHP code that references area `'adminhtml'` to use `'admin'`
4. Update `loadArea()`, event dispatching, layout loading, and any other area-aware code to use `admin`
5. Routers, fieldsets, and other config already under `<admin>` stay in place
6. Add a **compatibility shim** in config loading that auto-merges any remaining `<adminhtml>` content into `<admin>`, so third-party modules using the old pattern still work
7. Log a deprecation notice when the shim activates, so module authors know to update

### PHP files requiring changes

- Area constants and references throughout the codebase (`'adminhtml'` → `'admin'`)
- `Mage_Core_Model_App::loadArea()` and area-related logic
- Event observer area filtering
- Layout and design area loading
- Any code using `Mage_Core_Model_App_Area::AREA_ADMINHTML` or string `'adminhtml'` as an area identifier

### Not affected

- Store config paths like `admin/security/use_form_key` (under `<default><admin>`, not top-level `<admin>`)
- Class aliases like `admin/user` (model alias for `Mage_Admin_Model_User`)
- The `Mage_Admin` module itself (handles admin users/roles/ACL)

## Pros

- **Eliminates a long-standing source of confusion** — developers no longer need to know that routing goes in `<admin>` but events go in `<adminhtml>`
- **Simpler mental model** — one area name for admin backend: `admin`
- **Cleaner name** — `admin` is shorter and the `html` suffix was never meaningful
- **Compat shim protects third-party modules** — existing extensions keep working without changes, with deprecation notices guiding authors to update
- **Cleaner config caching** — one fewer top-level section to cache separately

## Cons

- **Large diff** — touches many config.xml files and PHP files
- **Third-party module impact** — even with the compat shim, module authors will eventually need to update their config.xml files
- **Risk of edge cases** — any code that hardcodes `'adminhtml'` as a config area root (beyond what we have identified) could break silently
- **Compat shim maintenance** — adds temporary complexity that needs to be removed in a future version

fballiano:
Closes #814

## Summary

- Merges the two confusingly named top-level config areas — `<admin>` (routers, fieldsets) and `<adminhtml>` (events, layout, translate) — into a single `<admin>` area
- Adds `AREA_ADMIN` as the primary constant; `AREA_ADMINHTML` is deprecated and aliased to `AREA_ADMIN`
- Renames `<adminhtml>` to `<admin>` in all 60 config.xml files, merging children where both sections existed
- Updates all ~67 observer PHP attributes from `area: 'adminhtml'` to `area: 'admin'`
- Updates all PHP code referencing the admin area (constants, `setArea()`, design config arrays)

## Backward compatibility for third-party modules

- **XML config**: A compatibility shim in `Config::_migrateAdminhtmlConfig()` auto-merges any `<adminhtml>` content into `<admin>` at config load time, with a deprecation log notice
- **PHP observer attributes**: The `AttributeCompiler` in `maho-composer-plugin` normalizes `area: 'adminhtml'` → `'admin'` at compile time (separate PR needed for the plugin repo)
- **Design paths**: `Package::areaToDesignDir()` maps area code `'admin'` to filesystem directory `'adminhtml'`, so templates/skins/layouts resolve correctly without renaming directories
- **Route names**: The router name `'adminhtml'` (from `<admin><routers><adminhtml>`) is unchanged — code checking `getRouteName()` uses the literal string, not the area constant

## Note

The `AttributeCompiler` change in `vendor/mahocommerce/maho-composer-plugin/` needs a separate PR to that repo.

## Test plan

- [x] All 1845 tests pass
- [x] PHPStan clean (no errors)
- [x] php-cs-fixer clean
- [x] `composer dump-autoload` compiles attributes correctly under `'admin'` area
- [ ] Manual: verify admin panel loads, login, dashboard, product edit, system config
- [ ] Manual: verify PDF generation renders with admin theme

### DIFF
--- a/app/code/core/Mage/Admin/Model/Observer.php
+++ b/app/code/core/Mage/Admin/Model/Observer.php
@@ -19,7 +19,7 @@ class Mage_Admin_Model_Observer
      *
      * @param \Maho\Event\Observer $observer
      */
-    #[Maho\Config\Observer('controller_action_predispatch', area: 'adminhtml', id: 'auth')]
+    #[Maho\Config\Observer('controller_action_predispatch', area: 'admin', id: 'auth')]
     public function actionPreDispatchAdmin($observer)
     {
         /** @var Mage_Admin_Model_Session $session */
--- a/app/code/core/Mage/Admin/etc/config.xml
+++ b/app/code/core/Mage/Admin/etc/config.xml
@@ -73,7 +73,7 @@
             </emails>
         </admin>
     </default>
-    <adminhtml>
+    <admin>
         <layout>
             <updates>
                 <admin>
@@ -81,5 +81,5 @@
                 </admin>
             </updates>
         </layout>
-    </adminhtml>
+    </admin>
 </config>
--- a/app/code/core/Mage/AdminNotification/Model/Observer.php
+++ b/app/code/core/Mage/AdminNotification/Model/Observer.php
@@ -15,7 +15,7 @@ class Mage_AdminNotification_Model_Observer
     /**
      * Predispath admin action controller
      */
-    #[Maho\Config\Observer('controller_action_predispatch', area: 'adminhtml')]
+    #[Maho\Config\Observer('controller_action_predispatch', area: 'admin')]
     public function preDispatch(\Maho\Event\Observer $observer)
     {
         if (Mage::getSingleton('admin/session')->isLoggedIn()) {
--- a/app/code/core/Mage/AdminNotification/etc/config.xml
+++ b/app/code/core/Mage/AdminNotification/etc/config.xml
@@ -44,7 +44,7 @@
             </adminnotification_setup>
         </resources>
     </global>
-    <adminhtml>
+    <admin>
         <translate>
             <modules>
                 <Mage_AdminNotification>
@@ -61,7 +61,7 @@
                 </adminnotification>
             </updates>
         </layout>
-    </adminhtml>
+    </admin>
     <default>
         <system>
             <adminnotification>
--- a/app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Search/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Search/Grid.php
@@ -93,7 +93,7 @@ class Mage_Adminhtml_Block_Sales_Order_Create_Search_Grid extends Mage_Adminhtml
             ->addAttributeToSelect('gift_message_available')
             ->addStoreFilter()
             ->addAttributeToFilter('type_id', array_keys(
-                Mage::getConfig()->getNode('adminhtml/sales/order/create/available_product_types')->asArray(),
+                Mage::getConfig()->getNode('admin/sales/order/create/available_product_types')->asArray(),
             ))
             ->addAttributeToFilter('status', [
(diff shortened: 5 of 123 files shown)


fballiano:
After deeper analysis, we've decided not to proceed with this consolidation. Here's why:

**The naming mismatch problem goes both ways.** Consolidating into `<admin>` (as this PR does) creates a permanent mismatch: the area code is `admin` but the filesystem directories (`app/design/adminhtml/`, `public/skin/adminhtml/`), the router name, route name, block/model alias groups, and controller prefixes all remain `adminhtml`. This requires a mapping layer (`areaToDesignDir()`) that lives forever.

We explored the reverse direction (consolidating into `<adminhtml>`), but that requires moving the router/fieldset config and updating every place the framework reads `admin/routers` — introducing its own set of backward compatibility concerns.

**Either direction trades one confusion for another.** The current split (`<admin>` for routers/fieldsets, `<adminhtml>` for events/layout/translate) is admittedly confusing, but it's a well-understood confusion that the entire Magento 1 ecosystem has dealt with for 15+ years. Consolidating introduces a *new* inconsistency that would surprise developers in different ways, plus adds backward-compat shims that need to be maintained indefinitely for third-party modules.

**The cost/benefit doesn't justify the churn.** This touches 60+ config files, 67+ observer attributes, and multiple core framework methods — all for a naming cleanup that doesn't enable new functionality. The risk of subtle breakage in third-party modules outweighs the clarity gained.

PR #828, opened 2026-04-13
Added HTML validation for WYSIWYG textarea mode

### LINKED ISSUE #776: Feature: HTML Validation of CMS Contents and Attributes
## Description

Currently, it's very easy to introduce HTML errors into any entity where the WYSIWYG editor can be used (CMS blocks, CMS pages, product descriptions, category descriptions, etc.). Errors can occur at any time when modifications are made to the HTML code (WYSIWYG is completely disabled or needs to be hidden, e.g. to set classes or when working extensively with legacy code). However, there's no way to directly validate the HTML artifact being edited in Maho. You either have to copy the finished HTML and validate it manually (e.g., here: https://validator.w3.org/#validate_by_input) or save it and then validate the entire page afterward, by which time the error may have already occurred.

It would be fantastic if the code could be validated directly on the specific entity where changes have been made, for example by clicking a button (`Validate HTML`). The W3C also offers an API for this: https://validator.w3.org/docs/api.html

The following settings might be useful for this feature:
* `Enabled` (`Yes`/`No` - Those who don't need or want the feature can disable it.)
* `Checker Instance URL` (default: `https://validator.w3.org/nu/` - You can also run a checker locally and adjust the URL accordingly.)
* `Prevent saving when HTML errors are present` (`Yes`/`No`)
  * This could be some kind of strict mode (which I'd love to have, but others might prefer not to), which validates changes made to the HTML entity automatically (only if there are changes) on save - and don't even save the faulty code. Instead, the HTML errors from the checker are returned and are shown to the user (like on manual `Validate HTML` button click).

This feature could be useful for any Maho instance, so I'm simply proposing it for the core here. :)

fballiano:
## Summary

- Adds a **"Validate HTML"** button on editor textareas when TipTap is disabled or toggled to source mode
- Sends content to the W3C Nu validator (`validator.w3.org/nu/`) via a backend proxy to avoid CORS
- Wraps the HTML fragment in a minimal document skeleton and adjusts line numbers in the response
- Displays validation results (errors/warnings with highlighted source extracts) in a Maho Dialog

## Changes

- `WysiwygController.php` — new `validateHtmlAction()` backend proxy
- `Wysiwyg/Config.php` — adds `validate_html_url` to editor config
- `Editor.php` — renders the "Validate HTML" plugin button (last in the row)
- `tools.js` — `validateHtmlContent()` function for the client-side logic

Closes #776

## Test plan

- [ ] Set WYSIWYG to "Enabled by Default", toggle to source mode → "Validate HTML" button should appear
- [ ] Set WYSIWYG to "Disabled by Default" → button should appear next to textarea
- [ ] Set WYSIWYG to "Disabled Completely" → button should appear next to textarea
- [ ] Click "Validate HTML" with valid HTML → dialog shows "No errors or warnings found"
- [ ] Click "Validate HTML" with invalid HTML (e.g. unclosed tags, `<br/>`) → dialog shows errors/warnings with highlighted extracts
- [ ] Click "Validate HTML" with empty textarea → dialog shows "no content" message
- [ ] Verify button is hidden when TipTap editor is active

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Helper/Form/Wysiwyg.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Helper/Form/Wysiwyg.php
@@ -34,6 +34,16 @@ class Mage_Adminhtml_Block_Catalog_Helper_Form_Wysiwyg extends \Maho\Data\Form\E
                     'onclick'  => "catalogWysiwygEditor.open('$wysiwygUrl', '{$this->getHtmlId()}')",
                 ])->toHtml();
         }
+        if ($this->getEntityAttribute()->getIsWysiwygEnabled()) {
+            $validateUrl = Mage::getSingleton('adminhtml/url')->getUrl('*/cms_wysiwyg/validateHtml');
+            $html .= Mage::getSingleton('core/layout')
+                ->createBlock('adminhtml/widget_button', '', [
+                    'label'    => Mage::helper('cms')->__('Validate HTML'),
+                    'type'     => 'button',
+                    'class'    => 'validate-html',
+                    'onclick'  => "validateHtmlContent('{$this->getHtmlId()}', '$validateUrl');",
+                ])->toHtml();
+        }
         return $html;
     }
 
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Helper/Form/Wysiwyg.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Helper/Form/Wysiwyg.php
@@ -41,6 +41,7 @@ class Mage_Adminhtml_Block_Catalog_Helper_Form_Wysiwyg extends \Maho\Data\Form\E
                     'label'    => Mage::helper('cms')->__('Validate HTML'),
                     'type'     => 'button',
                     'class'    => 'validate-html',
+                    'style'    => 'margin-left: 4px;',
                     'onclick'  => "validateHtmlContent('{$this->getHtmlId()}', '$validateUrl');",
                 ])->toHtml();
         }
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Helper/Form/Wysiwyg.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Helper/Form/Wysiwyg.php
@@ -31,6 +31,7 @@ class Mage_Adminhtml_Block_Catalog_Helper_Form_Wysiwyg extends \Maho\Data\Form\E
                     'type'     => 'button',
                     'disabled' => $this->getDisabled() || $this->getReadonly(),
                     'class'    => 'btn-wysiwyg',
+                    'style'    => 'margin-right: 4px;',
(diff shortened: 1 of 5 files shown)


fballiano:
@tmewes can you give this a check?

tmewes:
@fballiano I've tested the current version. It's definitely going in the right direction, thank you so much! :)

Unfortunately, if WYSIWYG is set to "Disabled Completely", the "Validate HTML" button doesn't appear for product and category textarea attributes.

fballiano:
@tmewes can you check this last version?

tmewes:
@fballiano Now the button is displayed and working. :)

Just a cosmetic note:
If WYSIWYG is set to "Disabled by Default", everything looks fine:
grafik
If WYSIWYG is set to "Disabled Completely", the "Validate HTML" button is the only one and has a margin-left now:
grafik
Maybe it's better to give the "WYSIWYG Editor" button a margin right instead?

I also continued testing around and found something that (I think) should be addressed directly in the first version.
grafik
The root cause here are the quotes, but this could be automatically solved when all `{{}}` directives are either rendered (maybe too complex?), replaced by a placeholder or are removed (could theoretically lead to false positives, e.g. a `Must be non-empty` HTML error for `img` elements) when sent to the HTML validator.

fballiano:
super nice catch @tmewes, PR updated! 🤞

tmewes:
@fballiano I tested the current state, everything validates fine now. Great! <3

I also checked (as a last test to not forget anything) also the "Enabled by Default" case and found something that I at least want to report.
On CMS pages and CMS block edit pages, when the HTML is in a faulty state for TipTap and disables itself by default, the current state looks like this:
grafik
When I click on "Show / Hide Editor", the alert is showing as expected:
grafik
After clicking "OK", the TipTap editor shows up:
grafik
Then clicked on "Show / Hide Editor" again to disable it and now the additional buttons including the "Validate HTML" button are available:
grafik

Would it be possible to add the "Validate HTML" button directly, so that is not needed to show and hide the TipTap editor to use it (also possible in another PR)?

fballiano:
@tmewes mmm that's a tricky situation, when the wysiwyg editor is enabled by deafult, it autoloads only in case the HTML is compatible with tiptap (which is a bit picky) so, if there's something in the HTML that would get lost by tiptap, then it doesn't load and it shows that alert... but it's not an HTML validation problem, it's something tiptap related... so i don't know... it's correct not to show the HTML validation button because tiptap by default always produces valid HTML...

it's a bit of a edge case...

tmewes:
@fballiano Okay, then I have an idea: When we ensure (in another PR of course), that all CMS contents (page, blocks), product textarea attributes and category textarea attributes are fine with TipTap when installing Maho freshly - with and without sample data - this problem should never occur 'natively'. Currently, even when installing a fresh Maho instance, this problem occurs. I tested without sample data and only 'home' loads the TipTap editor by default when loading the page. All others have probems loading TipTap.
grafik

What do you think? Could we get the 'default' (always pre-installed) and sample data TipTap-ready?

fballiano:
@tmewes makes sense and fortunately less important :-)

so let's merge this in the meanwhile?

tmewes:
@fballiano Yes, please! 😄

PR #839, opened 2026-04-15
Fix configurable product pricing inputs rendering with spurious single quotes

### LINKED ISSUE #838: Admin configurable product pricing inputs render with spurious single quotes ('value' or '')
## Summary

On the admin product edit page for a **configurable** product, every
pricing input rendered inside the _Super product attributes configuration_
tab shows literal single quotes around its value:

- An option with **no price adjustment** → `value="''"` (two quotes, user-visible as `''`)
- An option with price **10** → `value="'10'"`

Same for the corresponding `{{html_id}}` substitution: rendered ids become
`id="'...'"` instead of `id="..."`.

## Environment

- Maho `dev-main` @ `e44366669d0d535c005fd84089b5527c31d9e968`
- Affected file: `public/js/mage/adminhtml/product.js` (Product.Configurable)
- Affected helper: `public/js/varien/js.js` (Template / HANDLEBARS_PATTERN)

## Root cause

`Product.Configurable` creates its templates via `createTemplateFunction`
which uses `Template.HANDLEBARS_PATTERN`:

```js
// public/js/varien/js.js:806
static HANDLEBARS_PATTERN = /(^|.|\r|\n)({{(.*?)}})/;
```

```js
// public/js/mage/adminhtml/product.js:322
createTemplateFunction(template) {
    return new Template(template, Template.HANDLEBARS_PATTERN)
        .evaluate.bind(new Template(template, Template.HANDLEBARS_PATTERN));
}
```

Meanwhile the source templates **wrap every placeholder as `'{{key}}'`**
(surrounding single quotes) — both the in-constructor replacement of
`__id__` and the raw HTML in `catalog/product/helper/gallery.phtml` /
`catalog/product/edit/super/config.phtml`:

```js
// product.js:283-284
document.getElementById(idPrefix + 'attribute_template').innerHTML
    .replace(/__id__/g, "'{{html_id}}'")
```

```html
<!-- catalog/product/edit/super/config.phtml value_template -->
<input id="__id___pricing" type="text" ...
       value="'{{pricing_value}}'"/>
```

`HANDLEBARS_PATTERN` only consumes one character before `{{` (into the
`before` capture group), not the surrounding quotes. After `evaluate()`
the leading quote is preserved and the trailing quote — which is not
part of any match — is simply left in place. Hence the literal `''` or
`'10'` in the output.

The class even assigns a quote-aware pattern in its constructor:

```js
// product.js:274
this.templatesSyntax = new RegExp(
    '(^|.|\\r|\\n)(\'{{\\s*(\\w+)\\s*}}\')', "");
```

but `templatesSyntax` is never read — `createTemplateFunction` ignores it
and uses the bare HANDLEBARS_PATTERN. Looks like leftover from the
Prototype→Maho-Template refactor: the old Prototype template class used
a quote-aware pattern so the shipped templates assumed the surrounding
quotes would be consumed.

## Minimal repro

```js
// Uses the exact Maho Template class from public/js/varien/js.js
class Template {
    static HANDLEBARS_PATTERN = /(^|.|\r|\n)({{(.*?)}})/;
    constructor(t, p) { this.template = String(t); this.pattern = new RegExp(p, 'g'); }
    evaluate(data = {}) {
        return this.template.replaceAll(this.pattern, function() {
            const before = arguments[1] ?? '';
            const expr = (arguments[3] ?? '').trim();
            if (expr === '') return before;
            let v = data;
            for (const p of expr.split('.')) v = v?.[p];
            v ??= '';
            return before + v;
        });
    }
}

const tmpl = `<input value="'{{pricing_value}}'"/>`;
const fn = new Template(tmpl, Template.HANDLEBARS_PATTERN).evaluate
    .bind(new Template(tmpl, Template.HANDLEBARS_PATTERN));

console.log(fn({}));                    // <input value="''"/>
console.log(fn({ pricing_value: 10 })); // <input value="'10'"/>
```

## Expected

```
<input value=""/>
<input value="10"/>
```

## Suggested fix

Make `createTemplateFunction` use the quote-aware pattern the class
already declares (`this.templatesSyntax`), e.g.:

```diff
     createTemplateFunction(template) {
-        return new Template(template, Template.HANDLEBARS_PATTERN).evaluate.bind(new Template(template, Template.HANDLEBARS_PATTERN));
+        // Templates wrap placeholders as '{{key}}'; the pattern must consume the surrounding quotes.
+        const pattern = /(^|.|\r|\n)('{{\s*(.*?)\s*}}')/;
+        return new Template(template, pattern).evaluate.bind(new Template(template, pattern));
     }
```

Alternatively the template strings themselves could drop the surrounding
quotes and switch to `HANDLEBARS_PATTERN`, but that ripples into every
`_template` div in the affected phtml files plus the two `__id__` →
`'{{html_id}}'` replacements in the constructor.

I'll send a PR if that helps.

fballiano:
## Summary

- The `__id__` placeholder in `Product.Configurable` templates was replaced with `'{{html_id}}'` (surrounding quotes included), causing literal `'` characters to appear in rendered `id` attributes and adjacent output after template evaluation
- Removing the surrounding quotes from the two `.replace(/__id__/g, ...)` calls lets `HANDLEBARS_PATTERN` work correctly without a custom regex
- Intentional display quotes in text content (e.g. `<strong>'{{label}}'</strong>`) are preserved: the leading `'` is the natural "before" character consumed and re-emitted by the pattern, and the trailing `'` is a plain literal — both survive unchanged

Fixes #838

## Test plan

- [ ] Open a configurable product in the admin
- [ ] Go to the _Super product attributes configuration_ tab
- [ ] Verify pricing inputs show clean values (e.g. `10`, not `'10'`)
- [ ] Verify generated element IDs contain no spurious single quotes

### DIFF
--- a/public/js/mage/adminhtml/product.js
+++ b/public/js/mage/adminhtml/product.js
@@ -282,10 +282,10 @@ Product.Configurable = class {
         /* Generation templates */
         this.addAttributeTemplate = this.createTemplateFunction(
             document.getElementById(idPrefix + 'attribute_template').innerHTML.replace(/__id__/g,
-            "'{{html_id}}'").replace(/ template no-display/g, ''));
+            '{{html_id}}').replace(/ template no-display/g, ''));
         this.addValueTemplate = this.createTemplateFunction(
             document.getElementById(idPrefix + 'value_template').innerHTML.replace(/__id__/g,
-            "'{{html_id}}'").replace(/ template no-display/g, ''));
+            '{{html_id}}').replace(/ template no-display/g, ''));
         this.pricingValueTemplate = this.createTemplateFunction(document.getElementById(idPrefix + 'simple_pricing').innerHTML);
         this.pricingValueViewTemplate = this.createTemplateFunction(document.getElementById(idPrefix + 'simple_pricing_view').innerHTML);
 


postadelmaga:
Tested on a Maho project that previously reproduced the spurious-quote rendering on the Super product attributes configuration tab.

Procedure:
1. Applied this PR's diff to `public/js/mage/adminhtml/product.js`.
2. Cache flush + hard reload.

Results on a configurable product with multiple variants:
- Pricing inputs render cleanly. Empty-adjustment variants show `value=""` instead of `value="''"`; variants with `+10`/`+20` render `value="10"`/`value="20"` instead of `value="'10'"`/`value="'20'"`.
- Generated element `id` attributes no longer contain spurious single quotes.
- Intentional display quotes in text content (e.g. `<strong>'{{label}}'</strong>`) are preserved as claimed.

LGTM.

fballiano:
grazie!

PR #841, opened 2026-04-15
Fix validateSerializedObject crashing on JSON attribute values (#840)

### LINKED ISSUE #840: Mage_Core_Helper_String::isSerializedArrayOrObject treats valid JSON as PHP-serialized, breaking admin save when any attribute stores JSON
## Summary

`Mage_Core_Helper_String::isSerializedArrayOrObject()` returns `true`
for valid JSON strings starting with `{` or `[`. Its only in-tree caller,
`validateSerializedObject()`, then runs PHP `unserialize()` on that
JSON, which fails with a converted-to-exception warning:

```
Warning: unserialize(): Error at offset 0 of N bytes
  in Mage/Core/Helper/String.php on line 577
```

Because `Mage_Eav_Model_Entity_Attribute_Backend_Abstract::validate()`
calls `validateSerializedObject()` on every attribute value during admin
product validation, **any EAV attribute whose value happens to be a
JSON object or array crashes the validate endpoint**. Maho's
`form.js::_validate()` treats the resulting MahoError as a hard failure
and sends the browser to `BASE_URL` (admin dashboard) via
`_processFailure()`. End user perception: "Save does nothing — I end up
on the dashboard."

In the wild this is trivially triggered by any extension that stores
attribute data as JSON. I hit it with the
Boi_ConfigurableSwatches `media_gallery_swatches` attribute, a text
column holding `{"<imageId>":"<swatchOptionId>",...}`.

## Environment

- Maho `dev-main` @ `e44366669d0d535c005fd84089b5527c31d9e968`
- File: `app/code/core/Mage/Core/Helper/String.php`

## Root cause

```php
// app/code/core/Mage/Core/Helper/String.php:555-563
public function isSerializedArrayOrObject($data)
{
    if (is_string($data) && json_validate($data)) {
        return $data[0] === '{' || $data[0] === '[';
    }
    $pattern =
        '/^a:\d+:\{(i:\d+;|s:\d+:\".+\";|N;|O:\d+:\"\w+\":\d+:\{\w:\d+:)+|^O:\d+:\"\w+\":\d+:\{(s:\d+:\"|i:\d+;)/';
    return is_string($data) && preg_match($pattern, $data);
}
```

The method's name + caller contract is about **PHP-serialized** data
(`a:N:{...}` / `O:N:"Class":...`), yet the JSON branch unconditionally
returns true for any valid JSON object/array and short-circuits before
the PHP-serialize regex runs. The caller:

```php
// app/code/core/Mage/Core/Helper/String.php:571-582
public function validateSerializedObject($str)
{
    if (!$this->isSerializedArrayOrObject($str)) {
        return true;
    }
    if (!unserialize($str, ['allowed_classes' => false])) {
        return false;
    }
    return true;
}
```

`unserialize('{"a":1}')` always fails. Combined with Maho's
warning-to-exception converter, this becomes a thrown Exception through
the whole validate path.

## Repro

```php
<?php
require '/app/vendor/autoload.php';
Mage::app();
$helper = Mage::helper('core/string');
$value = '{"394953":"468","394954":"273"}';      // 31 bytes of valid JSON
var_dump($helper->isSerializedArrayOrObject($value));  // bool(true)  — WRONG
var_dump($helper->validateSerializedObject($value));   // fatal: Warning: unserialize(): Error at offset 0 of 31 bytes
```

Admin-side: any product whose form data (or any EAV attribute row)
contains JSON for a text attribute causes `saveAction`'s AJAX validate
(`admin/catalog_product/validate/`) to respond
`{error: true, message: "Warning: unserialize(): ..."}`. `mahoFetch`
throws on `error: true`; `form.js::_validate()` catches and runs
`_processFailure()` → `location.href = BASE_URL` → dashboard. The
actual form submit is never posted.

## Expected

JSON strings must not be flagged as PHP-serialized: JSON cannot encode
PHP objects, so the allow-listing hardening
`validateSerializedObject()` aims to apply has no target and running
`unserialize()` on JSON is always wrong.

## Suggested fix

```diff
     public function isSerializedArrayOrObject($data)
     {
-        if (is_string($data) && json_validate($data)) {
-            return $data[0] === '{' || $data[0] === '[';
+        // JSON is not PHP-serialize; unserialize() will always fail on it.
+        // Callers that validate serialized PHP data must skip JSON.
+        if (is_string($data) && $data !== ''
+            && ($data[0] === '{' || $data[0] === '[')
+            && json_validate($data)
+        ) {
+            return false;
         }
         $pattern =
             '/^a:\d+:\{(i:\d+;|s:\d+:\".+\";|N;|O:\d+:\"\w+\":\d+:\{\w:\d+:)+|^O:\d+:\"\w+\":\d+:\{(s:\d+:\"|i:\d+;)/';
         return is_string($data) && preg_match($pattern, $data);
     }
```

Sibling UX issue: `form.js::_processFailure()` hard-redirects to
`BASE_URL` on any mahoFetch validate error, silently hiding the
validation message the server returned in the same response body. Might
deserve its own ticket — happy to file it.

Happy to send a PR.

fballiano:
## Summary

`validateSerializedObject` was calling `unserialize()` on any value that `isSerializedArrayOrObject` returned `true` for — including valid JSON objects and arrays. Since `unserialize()` always fails on JSON, this threw a warning-turned-exception on every EAV attribute whose value was stored as JSON (e.g. `media_gallery_swatches`), crashing the admin product validate endpoint and silently redirecting to the dashboard instead of saving.

The suggested fix in #840 patches `isSerializedArrayOrObject` to return `false` for JSON, but that would break `Mage_Sales_Model_Quote_Item` (lines 521–522) and `Mage_Dataflow`, which rely on that method returning `true` for JSON so `UnserializeArray::unserialize()` is called to decode it.

The correct fix is in `validateSerializedObject` itself: skip `unserialize()` when the value is valid JSON. JSON cannot encode PHP objects, so the `allowed_classes` security check has no target there anyway.

## Changes

- `app/code/core/Mage/Core/Helper/String.php`: skip `unserialize()` in `validateSerializedObject` when value is valid JSON
- `tests/Backend/Unit/Core/Helper/StringSerializationTest.php`: add test cases for `validateSerializedObject` with JSON input

## Test plan

- [ ] `composer test -- --testsuite=Backend` passes
- [ ] Admin product save works for products with JSON-valued EAV attributes (e.g. `media_gallery_swatches`)

Fixes #840

Co-authored-by: postadelmaga <postadelmaga@users.noreply.github.com>

### DIFF
--- a/app/code/core/Mage/Core/Helper/String.php
+++ b/app/code/core/Mage/Core/Helper/String.php
@@ -574,6 +574,11 @@ class Mage_Core_Helper_String extends Mage_Core_Helper_Abstract
             return true;
         }
 
+        // JSON cannot encode PHP objects — unserialize() must never run on it
+        if (json_validate($str)) {
+            return true;
+        }
+
         if (!unserialize($str, ['allowed_classes' => false])) {
             return false;
         }
--- a/tests/Backend/Unit/Core/Helper/StringSerializationTest.php
+++ b/tests/Backend/Unit/Core/Helper/StringSerializationTest.php
@@ -322,3 +322,25 @@ describe('isSerializedArrayOrObject detects both JSON and legacy formats', funct
         expect($this->helper->isSerializedArrayOrObject(''))->toBeFalse();
     });
 });
+
+describe('validateSerializedObject does not call unserialize on JSON', function () {
+    beforeEach(function () {
+        $this->helper = Mage::helper('core/string');
+    });
+
+    it('passes JSON object without error', function () {
+        expect($this->helper->validateSerializedObject('{"394953":"468","394954":"273"}'))->toBeTrue();
+    });
+
+    it('passes JSON array without error', function () {
+        expect($this->helper->validateSerializedObject('[1,2,3]'))->toBeTrue();
+    });
+
+    it('passes valid PHP-serialized array', function () {
+        expect($this->helper->validateSerializedObject(serialize(['key' => 'val'])))->toBeTrue();
+    });
+
+    it('passes plain string', function () {
+        expect($this->helper->validateSerializedObject('just a string'))->toBeTrue();
+    });
+});


postadelmaga:
Tested on a Maho project that reproduced this bug against an EAV attribute storing a JSON object as its backend value.

Procedure:
1. Applied this PR's diff to `app/code/core/Mage/Core/Helper/String.php`.
2. `composer dump-autoload -o && ./maho cache:flush`.

Verification via `Mage::helper('core/string')->validateSerializedObject(...)` called from PHP:
- `{"394953":"468","394954":"273"}` → `true` ✓ (was throwing before)
- `[1,2,3]` → `true` ✓
- `serialize(['k' => 'v'])` → `true` ✓ (PHP-serialized path untouched)
- `'just a string'` → `true` ✓

Admin configurable product edit page renders 200 OK, save action no longer redirects to the dashboard. LGTM.

Issue #857, opened 2026-04-22
Add explicit defaults to TYPE_TIMESTAMP columns (~124 vulnerable to MySQL explicit_defaults_for_timestamp=OFF quirk)


fballiano:
## Context

Follow-up from #856. When we audited `TIMESTAMP_INIT_UPDATE` usage we also scoped a broader check: `TYPE_TIMESTAMP` columns declared with no `'default'` key in the options array. On MySQL with `explicit_defaults_for_timestamp = OFF` (not the default on MySQL 8.0+, but still present on older configs and some managed hosts), such `TIMESTAMP NOT NULL` columns receive a silently-injected `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`. That's engine- and config-specific behavior we want to eliminate for cross-engine (MySQL / MariaDB / PgSQL / SQLite) parity.

#856 handled the 13 explicit `TIMESTAMP_INIT_UPDATE` columns. This issue tracks the remaining ~124 columns that don't explicitly declare a default and are therefore vulnerable to the quirk.

## Audit summary

| Classification | Count | Examples | Recommended fix |
|---|---|---|---|
| **C** — insert-once timestamps | 43 | `created_at`, `added_at`, `applied_at`, `install_date` | `'default' => TIMESTAMP_INIT` |
| **U** — mutation timestamps | 32 | `updated_at`, `modified_at`, `last_update` | `'default' => TIMESTAMP_INIT` + ensure `_beforeSave()` sets the value |
| **N** — nullable / business | 49 | `last_login_at`, `expires_at`, `completed_at`, `processed_at`, `scheduled_at`, `performed_at` | `'nullable' => true, 'default' => null` — **requires per-column review**, N is a semantic change |
| **Total** | **124** | | |

### Affected modules (top offenders)

- **Mage_Sales** — ~38 columns across orders, quotes, invoices, shipments, creditmemos (`app/code/core/Mage/Sales/sql/sales_setup/install-1.6.0.0.php`)
- **Mage_Log** — ~12 columns (visitor tracking, customer activity logs)
- **Mage_Newsletter**, **Mage_PayPal**, **Mage_Dataflow** — 5–6 each
- **Mage_Catalog**, **Mage_Cms**, **Mage_Core**, **Mage_Cron**, **Mage_Customer**, **Mage_Downloadable**, **Mage_ProductAlert** — ~4 each
[...]

PR #858, opened 2026-04-22
Fix TIMESTAMP_INIT_UPDATE cross-engine parity (fixes #856)

### LINKED ISSUE #856: Audit TIMESTAMP_INIT_UPDATE usage: cross-engine parity and redundant server-side auto-update
## Background

While reviewing #855 (which fixes the silent MySQL auto-update on `admin_user.created`), we noticed a broader cross-engine inconsistency and some likely tech debt around the `Maho\Db\Ddl\Table::TIMESTAMP_INIT_UPDATE` sentinel.

### 1. PgSQL doesn't honor `TIMESTAMP_INIT_UPDATE`

The MySQL adapter emits `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP` for this sentinel. The PgSQL adapter (`lib/Maho/Db/Adapter/Pdo/Pgsql.php:2337-2338`) collapses both `TIMESTAMP_INIT` and `TIMESTAMP_INIT_UPDATE` to plain `CURRENT_TIMESTAMP` — PostgreSQL has no equivalent column-level `ON UPDATE` syntax; it would require a BEFORE UPDATE trigger. So on PgSQL installs, the auto-bump-on-update behavior never fires.

### 2. In practice no feature relies on server-side auto-update

The three columns currently using `TIMESTAMP_INIT_UPDATE`:

| Table.column | PHP writes it? | PHP reads it? |
|---|---|---|
| `core_flag.last_update` | ✅ `Mage/Reports/Model/Resource/Report/Abstract.php:55` | ✅ admin report UI |
| `payment_restriction.updated_at` | ❌ | ❌ |
| `catalog_category_dynamic_rule.updated_at` | ✅ `Mage/Catalog/Model/Resource/Category/Dynamic/Rule.php:78` + `_beforeSave()` | ❌ |

Only `payment_restriction.updated_at` actually depends on the server-side auto-bump to get populated — and nothing reads it, so the work is wasted. The other two are set explicitly from PHP.

## Proposed direction

1. **Replace `TIMESTAMP_INIT_UPDATE` with `TIMESTAMP_INIT`** on all three columns, and ensure `_beforeSave()` sets the value in PHP (pattern already in use for `catalog_category_dynamic_rule`). Works identically on MySQL, MariaDB, PgSQL, SQLite.
2. **Either remove `TIMESTAMP_INIT_UPDATE` from `Maho\Db\Ddl\Table`**, or keep it but add a docblock warning that it's MySQL-only and models should prefer `_beforeSave()` for cross-engine correctness.
3. **Audit `TYPE_TIMESTAMP` columns without an explicit default** (the scope-note follow-up from #855) — these are vulnerable to MySQL's `explicit_defaults_for_timestamp = OFF` quirk that silently injects `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`.

## Priority

Low / tech debt. No user-facing bug, but:
- Prevents future contributors from reaching for `TIMESTAMP_INIT_UPDATE` and silently shipping MySQL-only behavior.
- Closes a cross-engine parity gap.
- Gives #855's audit scope-note a home.

fballiano:
## Summary

- `TIMESTAMP_INIT_UPDATE` is a MySQL-only sentinel (emits `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`). PgSQL and SQLite adapters silently downgrade it to plain `CURRENT_TIMESTAMP`, so `updated_at` auto-bump never fires on those engines.
- Move the update-on-save responsibility into PHP (`_beforeSave()`) so all 13 affected columns behave identically across MySQL / MariaDB / PgSQL / SQLite.
- Deprecate the constant with `@deprecated` PHPDoc + runtime `E_USER_DEPRECATED` notice (emitted by all 3 adapters) so third-party modules keep working but get a migration signal.

## What changed

**Schema (13 columns across 11 scripts)** — replace `TIMESTAMP_INIT_UPDATE` with `TIMESTAMP_INIT` in:
`core_flag.last_update`, `payment_restriction.updated_at`, `catalog_category_dynamic_rule.updated_at`, `cataloglinkrule_rule.updated_at`, `feedmanager_destination/feed/dynamic_rule.updated_at`, `customersegmentation_segment/segment_customer.updated_at`, `customer_segment_email_sequence.updated_at`, `blog_post/category.updated_at`, `maho_paypal_vault_token.updated_at`.

**Models (5)** — add/update `_beforeSave()` to set `updated_at` via `Mage::app()->getLocale()->formatDateForDb('now')`: `Mage_Payment_Model_Restriction`, `Maho_CatalogLinkRule_Model_Rule`, `Maho_Paypal_Model_Vault_Token`, `Maho_CustomerSegmentation_Model_Segment`, `Maho_CustomerSegmentation_Model_EmailSequence`. Also fix 6 existing models to use `formatDateForDb('now')` per CLAUDE.md.

**Migrations (7 new `maho-26.5.0.php` scripts)** — use `modifyColumn()` guarded by `instanceof \Maho\Db\Adapter\Pdo\Mysql` to drop the `ON UPDATE CURRENT_TIMESTAMP` clause on existing MySQL installs. PgSQL/SQLite never emitted the clause so the migration is a no-op there.

**Deprecation** — `@deprecated` PHPDoc on `Maho\Db\Ddl\Table::TIMESTAMP_INIT_UPDATE`, runtime `E_USER_DEPRECATED` notice in the Mysql/Pgsql/Sqlite adapter branches. Sentinel still functional for backwards compat.

## Out of scope — follow-up #857

The related audit of `TYPE_TIMESTAMP` columns without explicit defaults (~124 columns exposed to MySQL's `explicit_defaults_for_timestamp=OFF` quirk) is deferred to #857.

## Test plan

- [x] `vendor/bin/php-cs-fixer fix` — clean
- [x] `vendor/bin/phpstan analyze` — clean
- [x] `composer test` — 1928 passed, 1 risky, 1 skipped
- [ ] Manual MySQL: `SELECT TABLE_NAME, COLUMN_NAME, EXTRA FROM information_schema.columns WHERE TABLE_SCHEMA = DATABASE() AND EXTRA LIKE '%on update%'` should return empty after migration
- [ ] Save-cycle: edit & save a blog post, feed, customer segment, payment restriction on PgSQL/SQLite — confirm `updated_at` bumps via PHP

Fixes #856

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Category/Dynamic/Rule.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Category/Dynamic/Rule.php
@@ -74,7 +74,7 @@ class Mage_Catalog_Model_Resource_Category_Dynamic_Rule extends Mage_Core_Model_
     #[\Override]
     public function save(Mage_Core_Model_Abstract $rule)
     {
-        $now = Mage::getSingleton('core/date')->gmtDate();
+        $now = Mage::app()->getLocale()->formatDateForDb('now');
         $rule->setUpdatedAt($now);
         if (!$rule->getCreatedAt()) {
             $rule->setCreatedAt($now);
--- a/app/code/core/Mage/Catalog/sql/maho_setup/maho-25.6.0.php
+++ b/app/code/core/Mage/Catalog/sql/maho_setup/maho-25.6.0.php
@@ -40,7 +40,7 @@ $table = $installer->getConnection()
     ], 'Created At')
     ->addColumn('updated_at', Maho\Db\Ddl\Table::TYPE_TIMESTAMP, null, [
         'nullable'  => false,
-        'default'   => Maho\Db\Ddl\Table::TIMESTAMP_INIT_UPDATE,
+        'default'   => Maho\Db\Ddl\Table::TIMESTAMP_INIT,
     ], 'Updated At')
     ->addIndex(
         $installer->getIdxName('catalog/category_dynamic_rule', ['category_id']),
--- a/app/code/core/Mage/Catalog/sql/maho_setup/maho-26.5.0.php
+++ b/app/code/core/Mage/Catalog/sql/maho_setup/maho-26.5.0.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Maho
+ *
+ * @package    Mage_Catalog
+ * @copyright  Copyright (c) 2026 Maho (https://mahocommerce.com)
+ * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+/** @var Mage_Catalog_Model_Resource_Setup $this */
+$installer = $this;
+$installer->startSetup();
+
+// Drop MySQL-only `ON UPDATE CURRENT_TIMESTAMP` clause on updated_at columns that were originally
+// declared with TIMESTAMP_INIT_UPDATE. Value is now managed explicitly in PHP via _beforeSave()
+// for cross-engine parity (see issue #856).
+if ($installer->getConnection() instanceof \Maho\Db\Adapter\Pdo\Mysql) {
+    $installer->getConnection()->modifyColumn(
+        $installer->getTable('catalog/category_dynamic_rule'),
+        'updated_at',
+        [
+            'type'     => Maho\Db\Ddl\Table::TYPE_TIMESTAMP,
+            'nullable' => false,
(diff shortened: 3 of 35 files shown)


mageaustralia:
@fballiano  - oh man. I feel guilty about this one because of the extra work involved to do it right. I just had a very specific issue and was trying to get past that and was surprised to see it's been a bug since forever that I had never known about and wasn't raised in OpenMage either! 

Nice extra finds that you discovered and fixed. I'm trying harder to ensure we consider sqlite / postgres in all changes.

PR #878, opened 2026-05-02
Tax RuleController: log caught exceptions and avoid singleton load

fballiano:
## Summary
- `setUsedModuleName('Mage_Tax')` in `_construct()` so `$this->__()` resolves against the Mage_Tax locale, then replace the verbose `Mage::helper('tax')->__('...')` calls with `$this->__('...')`. All affected strings already exist in `app/locale/en_US/Mage_Tax.csv`.
- Use `Mage::getModel('tax/calculation_rule')` instead of `Mage::getSingleton(...)` when loading the rule for save — a singleton retains state across calls and risks leaking data between operations.
- Replace `addError($e->getMessage())` with `addException($e, $msg)` in `saveAction` and `deleteAction` catch blocks so the underlying exception is also written to `exception.log` (user-visible message is unchanged).

Inspired by OpenMage/magento-lts#5525 — thanks to @Hanmac for the original work. Skipped the `_getSession()` migration to stay consistent with Maho's prevailing `Mage::getXXX(...)` accessor style.

## Test plan
- [ ] Save a new tax rule — success message appears
- [ ] Save a tax rule with a duplicate (rate / customer class / product class) — duplicate-error message appears, rule is not created
- [ ] Trigger a save error (e.g. force a DB exception) — generic error appears in admin, full exception is logged to `var/log/exception.log`
- [ ] Delete an existing tax rule — success message appears
- [ ] Delete a non-existent rule (manipulated id) — "rule no longer exists" message appears
- [ ] Verify breadcrumbs / title / messages all render in English (no untranslated keys)

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Tax/RuleController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Tax/RuleController.php
@@ -19,6 +19,14 @@ class Mage_Adminhtml_Tax_RuleController extends Mage_Adminhtml_Controller_Action
     public const ADMIN_RESOURCE = 'sales/tax/rules';
 
     /**
+     * @inheritDoc
+     */
+    protected function _construct()
+    {
+        $this->setUsedModuleName('Mage_Tax');
+    }
+
+    /**
      * Index action
      *
      * @return $this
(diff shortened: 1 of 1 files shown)


Hanmac:
Why aren't you using this function? 🤔 
https://github.com/MahoCommerce/maho/blob/3d5d0d99d57432435254c727a145c6c2b6f91705/app/code/core/Mage/Adminhtml/Controller/Action.php#L84

fballiano:
@Hanmac since the _helper (and similar) methods are deprecated i lean towards keeping consistent with using all Mage::xxx methods :-)

Hanmac:
> @Hanmac since the _helper (and similar) methods are deprecated i lean towards keeping consistent with using all Mage::xxx methods :-)

The MR i was referring also wanted to remove the `_getHelper` functions. Especially ones that just pipe the parameters.
But I think keeping `_getSession` for the controllers is fine. 🤷‍♂️

fballiano:
it is, cause it has to call the getSession with the name of the required session, but i think that for a developer is easier to have one unique way of doing thing. just a matter of preference at the end of the day .-)

Hanmac:
For me, `_getSession` is a easy way to not care if I'm in a Frontend or Backend Controller

But that's just preference

PR #898, opened 2026-05-06
MahoTree: node metadata primitive + category tree ID/count toggles

### LINKED ISSUE #897: Adminhtml: Show Category ID in Category Tree
# Description

Currently, the ID of a category can only be seen in the headline when clicking on it:
Image

It would be very useful (and I already got feedback from people working with and updating product and category attributes everyday) to see the category ids directly in the category tree. Something like that would be totally fine, but I'm open and thankful to any type of implementation. :)
Image

fballiano:
## Summary

Adds a generic node-metadata primitive to `MahoTree`, and uses it to ship #897 — letting users opt into displaying each category's ID alongside the existing product count in the admin category tree.

## What users see

A second row appears under `Collapse All / Expand All`:

- `[ ] Show ID` (default **off**) — appends `[id]` to each node
- `[x] Show count` (default **on**) — appends `(count)` to each node (matches the historical behavior, no regression)

When both are on, rows render as **`Name [id] (count)`**, with `[id]` and `(count)` styled smaller and muted gray so they read as secondary metadata. Hovering shows tooltips:
- Hover the name → `ID: <n>`
- Hover `[id]` → `Category ID`
- Hover `(count)` → `Products count`

State persists across sessions via `localStorage`. Toggling is instant — no server round-trip.

## Why this matters

@tmewes pointed out the bulk attribute-update workflow:
1. `Expand All` so all categories are loaded in the DOM
2. Paste a category ID, hit `Ctrl+F`, jump to the row
3. Click → edit attributes

A tooltip alone wouldn't satisfy step 2 because browser find skips `title=""` text. With `Show ID` on, the ID is in the DOM and `Ctrl+F` finds it. Default users (who don't need this) see no visual change.

## The MahoTree primitive

Rather than hardcoding category-specific fields and a custom DOM walker in `category.js`, the rendering of "muted metadata chips next to a tree node" is now a first-class MahoTree feature, available to any tree consumer.

**Node contract:**
```js
{
    id: 10,
    text: 'New Arrivals',
    meta: [
        { key: 'id',    value: '[10]', title: 'Category ID',    visible: false },
        { key: 'count', value: '(14)', title: 'Products count', visible: true  },
    ],
}
```

**Tree config:**
```js
new MahoTree(div, {
    metaOverrides: { id: true, count: false }, // initial overrides by key
});
```

**Runtime API:**
```js
tree.setMetaVisible('id', true);   // toggles visibility across all existing and future nodes
```

`MahoTreeNode.renderMeta()` is auto-called from `updateAttributes()`, so newly-constructed nodes (lazy load, post-save breadcrumb appends) pick up the current visibility state without any consumer wiring.

CSS classes `.maho-tree-meta` and `.maho-tree-meta-item` style the cluster generically — small, muted, with `cursor: help` on each chip.

A future CMS page tree, attribute-set tree, store hierarchy tree, etc. can opt in by emitting their own `meta` array — no JS work to mirror.

## Catalog wiring

- `Mage_Adminhtml_Block_Catalog_Category_Abstract::buildNodeMeta($node)` emits the catalog-specific `meta` array (id default-hidden, count default-visible).
- `buildNodeName()` is now just an escape — no count baked in.
- `category.js` reads localStorage and wires the two checkboxes to `tree.setMetaVisible(...)`. The previous custom DOM walker, `onLoad` lazyload hook, and `category_name` / `product_count` raw fields are all gone.
- Checkbox tree (product/customer category assignment) and widget chooser inherit the meta rendering automatically — their existing `(count)` display now lives in the muted meta cluster instead of being baked into the label. Same content, clearer visual separation.

## Other small bits

- Generic `title` attribute support on `MahoTreeNode` — any tree node can opt into a label-level tooltip.
- Generic `onLoad` callback on the tree's lazyload options — useful beyond this PR.
- Toolbar pipe separators (`|`) removed in favour of a flex-gap layout for visual consistency between the action and toggle rows.

Closes #897

## Test plan

- [ ] Open Catalog → Manage Categories: hover any node → tooltip reads `ID: <n>`.
- [ ] Toggle `Show ID` on → `[id]` appears after each name; tooltip on `[id]` reads `Category ID`.
- [ ] Toggle `Show count` off → counts disappear from labels; toggle back on → counts return.
- [ ] Toggle states persist across page reloads (localStorage).
- [ ] With `Show ID` on, `Ctrl+F` finds a category by its ID number.
- [ ] Add / edit / delete a category → newly-rendered nodes pick up the active toggle format.
- [ ] Expand a deep subtree (lazy load) → newly-loaded children pick up the active toggle format.
- [ ] Categories with `&` or other HTML entities in their names render correctly (no `&amp;` shown literally).
- [ ] Open a product → Categories tab → checkbox tree shows `Name (count)` (count now muted; content unchanged).
- [ ] Insert a Catalog Category Link widget into a CMS page → widget chooser tree shows the same muted-count format.


fballiano:
@tmewes what do you think about this?

tmewes:
@fballiano Thank you for the super fast PR! This is a good first step and I understand this solution to save space which we don't have for the category tree. :D

Let me describe a process/use case I saw in out content team:
They got a list with category IDs and related attributes to update. There are a lot of categories with a lot of sub categories and some of their names are nearly the same.
What they do to update the category attributes is:
* Click on `Expand All` above the category tree (we implemented the `- ID: [id]` example from the issue description as a rewrite), so all categories are shown incl. theur IDs as searchable text 
* Copy a category ID from the list to clipboard
* Search for this category (Ctrl + F)
* Click on the category
* Update the related attributes

Do you have any ideas how to solve this, so the IDs can be found in the category tree?

fballiano:
@tmewes can you please checkout this new version?

tmewes:
@fballiano Thank you for the huge update! I tested the latest state and I really like it. :)

grafik

This will solve the problem. Thank you! <3

PR #834, opened 2026-04-14
Migrated routing to`symfony/routing` with `#[Route]` attributes

fballiano:
Closes #809

## Summary

Replaces the legacy `Router_Standard`/`Router_Admin`/`Router_Install` routing system with Symfony Routing. All controller actions declare URL patterns via `#[Maho\Config\Route]` PHP attributes, compiled at `composer dump-autoload` time into opcached PHP arrays. Runtime uses Symfony's `CompiledUrlMatcher` and `CompiledUrlGenerator` — no `RouteCollection` is built per request.

### New components

- **`#[Maho\Config\Route]` attribute** (`lib/Maho/Config/Route.php`) — declares URL patterns on controller action methods. Repeatable for multiple paths/methods.
- **`RouteCollectionBuilder`** (`lib/Maho/Routing/`) — accessor for the opcached compiled data dumped at `composer dump-autoload` (`vendor/composer/maho_url_matcher.php`, `maho_url_generator.php`, `maho_attributes.php`). Handles admin frontName resolution and reverse lookups for URL generation and forward dispatch.
- **`ControllerDispatcher`** (`lib/Maho/Routing/`) — dispatches matched Symfony routes. Honors before/after override chains for both admin (`admin/routers/adminhtml/args/modules`) and frontend (`frontend/routers/<frontName>/args/modules`), preserving M1 chain semantics for third-party modules. Validates `_adminFrontName` at match time so forged admin URLs fall through to noroute.
- **`Mage_Core_Model_Controller_Front_Observer`** — single observer on `controller_front_dispatch_before` that runs the six pre-dispatch steps: base URL check, trailing slash, store resolution, DB URL rewrite, config rewrite, HTTPS enforcement. Each step short-circuits the chain if a redirect has been set.

### What was removed

- `Mage_Core_Controller_Varien_Router_Standard`
- `Mage_Core_Controller_Varien_Router_Admin`
- `Mage_Install_Controller_Router_Install`
- `Mage_Core_Model_Url_Rewrite_Request` + its `core/url_rewrite_request` alias and `<request_rewrite>` config node
- Legacy `_checkBaseUrl()` from `Mage_Core_Controller_Varien_Front`
- XML catch-all route definitions (`<use>standard</use>` / `<use>admin</use>`) from all module `config.xml` files
- `web/routers` config entries for admin/standard/install
- `Front::getRouterByRoute()` / `Front::getRouterByFrontName()` (unused after migration)

### What was kept

[...]

PR #903, opened 2026-05-07
Captcha + login fixes (W3C validity, route GET, login fatal)

### LINKED ISSUE #902: Altcha: name attribute causes an HTML validation info
# Description

I recognized an info message in the W3C Markup Validation Service in all my Maho instances, which is related to the altcha widget:
Image
This message is also included in the [validation result of the Maho demo](https://validator.w3.org/nu/?doc=https%3A%2F%2Fdemo.mahocommerce.com%2F#textarea).

Is it possible to remove the `name` attribute without any side effect/problems?

fballiano:
Three small frontend fixes uncovered while addressing #902.

## 1. Remove invalid `name` attribute from `<altcha-widget>` (fixes #902)
W3C Nu HTML Checker: `The name attribute is never allowed on the altcha-widget element`. Per the [Altcha docs](https://altcha.org/docs/website-integration) and `Widget.svelte`, the prop only names the hidden input rendered inside the widget's own DOM — no deduplication, registry, or other side effect. Maho positions the widget as `position: fixed` outside every form (`public/js/maho-captcha.js`) and injects its own `<input name="maho_captcha">` via `setHiddenInputValue()`, so removing the attribute is safe — Altcha falls back to the default `'altcha'` for its unused internal input.

## 2. Restore GET on `/captcha/index/challenge` (regression from #834)
The Symfony routing migration in #834 restricted the route to `POST` only. The Altcha widget fetches with `GET` by default (`Widget.svelte:525`) and exposes no public attribute to switch — every Maho frontend captcha was failing with `405 Method Not Allowed`. Drop the `methods` restriction to restore pre-#834 behavior.

## 3. Drop int sentinel from `Mage_Customer_Block_Form_Login::$_username` (regression from #901)
#901 narrowed the property to `string|int`, but `customer/session::getUsername(true)` returns `null` on a fresh login page visit (no prior failed-login flash), causing a fatal TypeError on `/customer/account/login`. The `int` in the union only existed to host the `-1` "not yet fetched" sentinel — the value is conceptually always a string. Switch to empty string as both default and sentinel; "not fetched" and "fetched empty" produce the same observable result, and the lazy cache still avoids re-reading the session.

## Test plan
- [ ] Re-run W3C validation on a Maho frontend page → `name` info notice on `<altcha-widget>` is gone.
- [ ] Load any captcha-protected frontend page → `GET /captcha/index/challenge/` returns `200 OK` (no 405).
- [ ] Visit `/customer/account/login` on a fresh session → no fatal, form renders.
- [ ] Submit a captcha-protected form (e.g. contact form) → verification passes server-side.

### DIFF
--- a/app/code/core/Mage/Customer/Block/Form/Login.php
+++ b/app/code/core/Mage/Customer/Block/Form/Login.php
@@ -115,8 +115,6 @@ class Mage_Customer_Block_Form_Login extends Mage_Core_Block_Template
 
     /**
      * Retrieve username for form field
-     *
-     * @return string
      */
     public function getUsername(): string
     {
--- a/app/code/core/Mage/Customer/Block/Form/Login.php
+++ b/app/code/core/Mage/Customer/Block/Form/Login.php
@@ -19,7 +19,7 @@
  */
 class Mage_Customer_Block_Form_Login extends Mage_Core_Block_Template
 {
-    private string|int $_username = -1;
+    private ?string $_username = null;
 
     /**
      * Set redirect URL before rendering login form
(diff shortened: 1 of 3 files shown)


fballiano:
@tmewes this should solve the problem ;-)

tmewes:
@fballiano Thank you! :)

PR #907, opened 2026-05-08
Added ignore list for HTML validator messages

### LINKED ISSUE #906: Bug: HTML validation of contents containing only headlines > 1 results in warning
# Description

Some hours ago, a new W3C HTML validation rule was added to warn if a given document contains headlines and h1 is missing.

Message: `This document has heading elements but none of them has a computed heading level of 1.`
Commit: https://github.com/validator/validator/commit/1c14c41462e2fba7b7dc09bb9f94e5e2f192e99e
Issue: https://github.com/validator/validator/issues/2095

Sadly this effects the new `Validate HTML` button introduced in #828. When e.g. the contents of the CMS page `Home page` (untouched demo/preinstalled data) this warning is triggered:
Image

Can this warning be ignored for the validation or could the skeleton potentially be adjusted in such a way that this error does not occur?
In CMS content, it is always possible for only partial content to be present. When editors receive a warning in such cases - where there is nothing to actually fix - it becomes frustrating very quickly.

fballiano:
## Summary

Adds an admin-configurable ignore list for messages emitted by the W3C Nu HTML validator (the "Validate HTML" button on editor / textarea fields).

Motivated by a recent validator change (commit [`1c14c41`](https://github.com/validator/validator/commit/1c14c41462e2fba7b7dc09bb9f94e5e2f192e99e)) that warns whenever a document has heading elements but no `h1`. CMS pages and blocks are fragments — often legitimately starting at `h2` or `h3` — so that warning becomes a permanent false positive editors cannot fix. Hard-coding suppression of one specific message is brittle (wording can change, and rules drift), so this is a generic mechanism users can extend as needed.

## Changes

- New setting **System → Configuration → Content Management → HTML Validator → Ignore List** (textarea, empty by default; config path `cms/html_validator/ignore`).
- Each non-empty line is a substring; if a validator message contains it, the message is dropped before being returned to the dialog.
- Lines starting with `#` are treated as comments.
- The dialog shows `N message(s) hidden by ignore list.` when filtering kicks in, so editors aren't left wondering where a familiar warning went.
- Setting lives in its own group (not under "WYSIWYG Options") because the Validate HTML button appears wherever the editor element is used — catalog products, categories, newsletter and email templates, etc., not just CMS WYSIWYG.

Default behaviour is unchanged (empty list = no filtering).

Closes #906

## Test plan

- [ ] Set ignore list to empty → validating CMS Home page shows the new "computed heading level of 1" warning.
- [ ] Add `none of them has a computed heading level of 1` to the ignore list → re-validate; warning is hidden, dialog shows "1 message hidden by ignore list."
- [ ] Add a `# comment line` → ignored.
- [ ] Substrings with leading/trailing whitespace and mixed line endings (CRLF) → trimmed correctly.
- [ ] CMS block starting with `h3` is unaffected (no spurious warnings introduced).
- [ ] Validate HTML on a catalog product description → ignore list applies there too.


fballiano:
@tmewes can you give this one a test?

tmewes:
@fballiano Thank you!! Works as expected. :)

Before:
grafik

After:
grafik

PR #938, opened 2026-05-24
Fix fatal error opening shipment view from the global Shipments grid

claudiobianco:
The Shipments grid links to /admin/sales_shipment/view/shipment_id/X/... That URL has no #[Maho\Config\Route] attribute, so it falls through to the legacy dispatcher, which sets routeName='admin' from the frontName. Mage_Adminhtml_Sales_ShipmentController::viewAction() then forwards to sales_order_shipment::view; the forward keeps the routeName 'admin', so getFullActionName() resolves to 'admin_sales_order_shipment_view'. The layout XML only declares the adminhtml_-prefixed handle, so the handle is never loaded and getBlock('sales_shipment_view') returns false. The controller then calls updateBackButtonUrl() on the false return, which raises a fatal Error and blanks the page.

Apply two complementary fixes:

* Controller: guard the getBlock() return with a null-check so the call site never crashes if the handle is missing for any future reason.
* Layout XML: alias 'admin_sales_order_shipment_view' to load the existing 'adminhtml_sales_order_shipment_view' handle, so the page actually renders when reached via the legacy URL.

### DIFF
--- a/.github/workflows/phpdoc.yml
+++ b/.github/workflows/phpdoc.yml
@@ -24,6 +24,9 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v6
 
+      - name: Strip SPDX lines from PHP source
+        run: find app lib -name "*.php" -exec sed -i '/ \* SPDX-/d' {} +
+
       - name: Generate phpDoc
         run: phpdoc -c .github/phpdoc.xml
 
--- a/app/code/core/Mage/Adminhtml/controllers/Sales/Order/ShipmentController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Sales/Order/ShipmentController.php
@@ -116,11 +116,9 @@ class Mage_Adminhtml_Sales_Order_ShipmentController extends Mage_Adminhtml_Contr
 
             $this->loadLayout();
 
-            /** @var Mage_Adminhtml_Block_Sales_Order_Shipment_View|false $block */
+            /** @var Mage_Adminhtml_Block_Sales_Order_Shipment_View $block */
             $block = $this->getLayout()->getBlock('sales_shipment_view');
-            if ($block) {
-                $block->updateBackButtonUrl($this->getRequest()->getParam('come_from'));
-            }
+            $block->updateBackButtonUrl($this->getRequest()->getParam('come_from'));
 
             $this->_setActiveMenu('sales/shipment')
                 ->renderLayout();
--- a/app/code/core/Mage/Adminhtml/controllers/Sales/Order/ShipmentController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Sales/Order/ShipmentController.php
@@ -116,9 +116,11 @@ class Mage_Adminhtml_Sales_Order_ShipmentController extends Mage_Adminhtml_Contr
 
             $this->loadLayout();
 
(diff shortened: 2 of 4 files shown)


claudiobianco:
This is probably not the proper fix and rather use #[Maho\Config\Route] somewhere. Not sure where tho :grimacing:

fballiano:
Pushed a follow-up commit (8c3f2eb) that addresses the root cause at the routing layer rather than per-controller.

## What's actually wrong

The crash isn't really about Shipments. It's a routing-migration regression in `lib/Maho/Routing/ControllerDispatcher.php::dispatchForward()`:

```php
if (!$request->getRouteName()) {
    $request->setRouteName($moduleName); // moduleName = frontName, e.g. 'admin'
}
```

Compare with the Symfony-matched admin path (`setAdminRequestNames`, line 417) which correctly does:
```php
$request->setRouteName('adminhtml');
```

So Symfony-matched admin routes get `routeName='adminhtml'`, but **any admin URL without a `#[Route]` attribute** falls through to legacy dispatch and ends up with `routeName='admin'`. That breaks:
- `getFullActionName()` (yields `admin_*` instead of `adminhtml_*`), so layout handles with the `adminhtml_` prefix never load;
- `controller_action_predispatch_<routeName>_*` and `_postdispatch_*` events, which fire under the wrong key.

In M1, `Mage_Core_Controller_Varien_Router_Admin` always set `routeName='adminhtml'` regardless of frontName. The Maho fallback path lost that invariant during the Symfony migration.

## Same bug, two more grids

Verified the same pattern affects **Invoices and Credit Memos**:
- `Mage_Adminhtml_Controller_Sales_Invoice::viewAction()` → `_forward('view', 'sales_order_invoice', ...)` → `admin_sales_order_invoice_view` handle missing → fatal in `Sales/Order/InvoiceController.php:132`.
- `Mage_Adminhtml_Controller_Sales_Creditmemo::viewAction()` → same shape → fatal in `Sales/Order/CreditmemoController.php:180`.

Their grid blocks generate `*/sales_invoice/view` and `*/sales_creditmemo/view` respectively (`Block/Sales/{Invoice,Creditmemo}/Grid.php:146`/`144`), and their parent `viewAction()` methods lack `#[Route]` too. So the original PR would have left those two grids crashing identically.

## The fix

One line, at the chokepoint:

```php
$request->setRouteName(RouteCollectionBuilder::getRouteByFrontName($moduleName) ?? $moduleName);
```

`RouteCollectionBuilder::getRouteByFrontName()` already exists for this exact mapping (returns `'adminhtml'` for the admin frontName, `null` otherwise, so the `?? $moduleName` keeps frontend behavior unchanged).

This:
- Restores the M1 invariant that admin requests always carry `routeName='adminhtml'` regardless of dispatch path.
- Fixes Shipments, Invoices, Credit Memos, and any other admin URL that falls through to legacy dispatch, in one place.
- Makes the layout-handle alias and the controller null-check unnecessary, so I reverted both (the null-check was inconsistent with `InvoiceController.php:132`, `CreditmemoController.php:180`, and `CustomerController.php:521`/`585`, which all call `getBlock()->method()` directly without guarding).

The PR's original diagnosis was 100% correct, just at the wrong layer. Thanks for the clear write-up, it made the root cause easy to spot.

PR #941, opened 2026-05-24
Read `sales_flat_shipment.packages` as JSON

claudiobianco:
Mage_Sales_Model_Order_Shipment::_beforeSave() has stored the packages column as JSON since the serialize/unserialize sweep in #552, but the admin block still read it back with unserialize(). unserialize() on a JSON string returns false (with a notice), so every shipment with packed packages displayed an empty packages popup in the admin and produced an empty packing-slip PDF (the PDF block delegates to the admin block's getPackages()).

Switch the read path to json_decode() so the column is read in the format it was written. The maho-26.2.1 setup script already converts any pre-migration PHP-serialized packages rows to JSON, so dropping the unserialize() branch is safe for both fresh and upgraded installs.

Defensive guards retained: empty values still yield [], non-array inputs that already happen to be arrays pass through unchanged, and invalid JSON yields [] rather than null/false reaching the caller.


fballiano:
Thanks @claudiobianco! I went for a slightly different approach which seems cleaner, hope it works the same for you!

PR #945, opened 2026-05-24
Fixed one-step checkout for logged-in customers with an address book

claudiobianco:
The one-step checkout (`checkout/options/onestep_checkout_enabled = 1`, the default) was unusable for logged-in customers with a saved address: the right column sat on "Waiting for address..." / "Waiting for shipping method..." indefinitely, the radio defaulted to "Ship to different address" even when the customer's default billing and shipping pointed at the same entry, and changing the country in the form back to one with regions (e.g. DE) silently stopped the auto-load.

Five small gaps combine. Each is fixed at the smallest possible scope; together they restore the auto-load.

1. `OneStepCheckout.hasMinimumAddressData(prefix)` inspected only the explicit `${prefix}:country_id` / `:postcode` / `:city` inputs. For a customer with an address book those sit inside the hidden `#${prefix}-new-address-form` and are empty; the selection lives in `#${prefix}-address-select` instead. `checkPrefilledBilling()` short-circuited and `saveBilling()` was never auto-fired.

   Fix: a non-empty `${prefix}-address-select` value is sufficient. The matching server action loads the customer address from that id.

2. The same method then required a region whenever the region select had more than one option, regardless of whether the country actually demanded one. DE has 17 Bundesländer in `directory_country_region` but is not listed in `general/region/state_required`, so switching the form back from a region-less country (CN) to DE produced a populated region select with no value selected and silently blocked `saveBilling` until the user picked a Bundesland.

   Fix: trust the `required-entry` class that `RegionUpdater` toggles per `state_required`. Only require a region when the country actually demands one.

3. `Mage_Checkout_OnepageController::estimateBillingAction` read only the `billing[*]` POST payload and ignored `billing_address_id`. Even if the client had fired `saveBilling`, the server fed an empty array to `$billingAddress->addData()` and again to the shipping address (when `use_for_shipping = 1`). The classic onepage flow doesn't hit this because `saveBillingAction` already forwards `billing_address_id` into `Mage_Checkout_Model_Type_Onepage::saveBilling`, which has the customer-address branch. `estimateBillingAction` was added later for one-step shipping-rate estimation and forgot to do the same.

   Fix: when `billing_address_id` is posted, load the customer address with an ownership check against the quote customer, then `importCustomerAddress()` onto the billing quote address. Inside the `use_for_shipping` branch, do the same for the shipping quote address so rates are collected against the resolved address. ~~Posted `billing[*]` data still layers on top via `addData()`, so inline edits and `use_for_shipping` toggles keep winning.~~

   > **Edited (correction):** the description above was inaccurate — the implementation is an exclusive `if`/`else`, so when `billing_address_id` is present the saved address **fully replaces** the posted `billing[*]` fields (`addData()` is not called). This is intentional: the hidden new-address form holds only the customer's stale *default*, so layering it would clobber the chosen address. The `use_for_shipping` flag is still read from the posted payload, so the toggle keeps working; the guest / new-address flow (no `billing_address_id`) keeps the original `addData()` path unchanged.

4. `Mage_Checkout_Model_Type_Onepage::initCheckout` imports billing and shipping from the customer's defaults but never sets `same_as_billing` on the shipping address. The billing block's `isUseBillingAddressForShipping()` therefore returns false on a fresh quote and the template checks "Ship to different address" by default, even when the customer has a single saved address used for both purposes.

   Fix: when the customer's default billing and shipping point at the same `customer_address_entity` row, set `same_as_billing = 1` on the quote shipping address right after `assignCustomer()`. Different defaults are left untouched.

5. `OneStepCheckout.checkPrefilledBilling()` fired only `saveBilling`. For a customer who has genuinely separate billing and shipping addresses (different defaults), the shipping side never reached the server until they touched the shipping form, leaving the checkout on "Waiting for shipping method...".

   Fix: when `billing:use_for_shipping_no` is checked AND `hasMinimumAddressData('shipping')` passes, also call `saveShippingEstimate()` from `checkPrefilledBilling()`.

Backwards compatibility — guest flow, new-address flow, and customers with different defaults are all untouched on the paths they already exercised. An address belonging to a different customer fails the ownership check and falls through to the existing empty-data behaviour, so the server fix cannot leak addresses across accounts.


claudiobianco:
Maybe the check if the address id from the POST is a valid one that belongs to the user should be more robust.

app/code/core/Mage/Checkout/controllers/OnepageController.php
```diff
@@ -433,10 +433,27 @@ public function estimateBillingAction(): void
...
         $quote = $this->getOnepage()->getQuote();
 
+        // With an address-book selection the billing[*] payload is empty (the
+        // new-address form is hidden). Resolve the customer address by id;
+        // posted fields layer on top below via addData().
+        $customerAddress = null;
+        if ($customerAddressId) {
+            $candidate = Mage::getModel('customer/address')->load($customerAddressId);
+            if ($candidate->getId()
+                && (int) $candidate->getCustomerId() === (int) $quote->getCustomerId()
```
fballiano:
```suggestion
            if ($candidate->getId() && (int) $candidate->getCustomerId() === (int) $quote->getCustomerId()
```

app/code/core/Mage/Checkout/controllers/OnepageController.php
```diff
@@ -433,10 +433,26 @@ public function estimateBillingAction(): void
...
         $quote = $this->getOnepage()->getQuote();
 
+        // With an address-book selection the billing[*] payload is empty (the
+        // new-address form is hidden). Resolve the customer address by id;
+        // posted fields layer on top below via addData().
+        $customerAddress = null;
+        if ($customerAddressId) {
+            $candidate = Mage::getModel('customer/address')->load($customerAddressId);
+            if ($candidate->getId() && (int) $candidate->getCustomerId() === (int) $quote->getCustomerId()
+            ) {
```
fballiano:
```suggestion
            if ($candidate->getId() && (int) $candidate->getCustomerId() === (int) $quote->getCustomerId()) {
```

claudiobianco:
I was thinking about maybe loading all address ids from the customer and then check if the id from the POST is one of them. I don't like loading data from a potentially different user. I can take a look at it maybe later or tomorrow.

fballiano:
@claudiobianco great comment, I think last commit addresses it :-)

fballiano:
@claudiobianco can you give this last version a test?

PR #950, opened 2026-05-26
Backfill paypal_setup schema for M1.9 migrations via 6.0.0->6.0.1 upgrade

claudiobianco:
## Problem

Legacy Mage_Paypal shops migrated from M1.9 can end up with the `paypal_setup` resource version advanced to `6.0.0` while the schema was never created.

`upgrade-1.6.0.0-6.0.0.php` only runs when the recorded `paypal_setup` version is `1.6.0.0`. The resolver (`_getModifySqlFiles`, `TYPE_DB_UPGRADE`) requires the file's "from" version to be `>=` the recorded DB version, but Mage_Paypal shipped up to `1.6.0.6`. So for any patch level above `1.6.0.0` the bridge file is skipped, yet `_upgradeResourceDb` still advances `paypal_setup` to `6.0.0`.

The result: `paypal_webhook_event` / `paypal_vault_token` tables and the `paypal_order_id` columns on `sales/quote_payment` and `sales/order_payment` are never created, and `/checkout/onepage/` 500s.

## Fix

Bump `Maho_Paypal` to `6.0.1` and add two files, both of which just `require install-6.0.0.php` (which is fully idempotent: `createTable` is guarded by `isTableExists`, and `addColumn`/`addIndex` are idempotent in Maho's adapter):

- **`upgrade-6.0.0-6.0.1.php`** repairs existing shops. Its "from" of `6.0.0` is `>=` both the broken `6.0.0` state and any un-migrated `1.6.0.x` version, so it reaches every affected merchant in a single step. It is a no-op on healthy shops.
- **`install-6.0.1.php`** lets fresh installs resolve directly to `6.0.1`, so the schema body runs exactly once instead of running again via the `6.0.0 -> 6.0.1` upgrade.

> Note: this supersedes the original `maho_setup` approach (commit c6ed0f0); the `6.0.0 -> 6.0.1` upgrade path is simpler and keeps the repair inside the standard DB-upgrade pipeline.


fballiano:
@claudiobianco genuine question, since the old paypal modules were removed in 26.5, is this change still needed?

claudiobianco:
Genuine answer... I don't know ;). I was coming from a mage 1.9 / mageone base. After I ran the migrations the table paypal_vault_token was missing. My paypal resource was on 1.6.0.6 IIRC.

Not sure what the version code was before the Maho_Paypal module.

fballiano:
## Final implementation

The fix repairs `paypal_setup` through the standard DB-upgrade pipeline by bumping the module to `6.0.1`, using two files.

**Files (in `app/code/core/Maho/Paypal/sql/paypal_setup/`):**

- `install-6.0.1.php` holds the full, idempotent schema body: `createTable` guarded by `isTableExists`, and `addColumn`/`addIndex` already idempotent in Maho's adapter (`addColumn` no-ops when the column exists, `addIndex` does `DROP INDEX ... ADD`). Safe to re-run on populated tables, no data loss. Fresh installs resolve to the highest install file `<= 6.0.1`, so they run this directly, record `6.0.1`, and no post-install upgrade fires.
- `upgrade-6.0.0-6.0.1.php` is the repair path for existing shops; it just `require`s `install-6.0.1.php`. Its `from` of `6.0.0` is `>=` both the broken `6.0.0` state and any un-migrated `1.6.0.x` version, so it reaches every affected merchant in a single step and is a no-op on healthy shops.

**Removed:**
- `upgrade-1.6.0.0-6.0.0.php` (the original 26.5.0 bridge). It only matched a recorded version `<= 1.6.0.0`, while legacy Mage_Paypal shipped up to `1.6.0.6`, which is what caused the bug. It is now fully dominated by `6.0.0 -> 6.0.1` and added no coverage.
- The separate `install-6.0.0.php` + `install-6.0.1.php` stub split was collapsed into a single `install-6.0.1.php` that carries the body, so the upgrade requires it directly. One fewer file, identical behavior.

**Coverage verified against `_getModifySqlFiles` (`Mage_Core_Model_Resource_Setup`):**

| Starting state | What runs | Result |
|---|---|---|
| Fresh install | `install-6.0.1.php` only | schema created once, records `6.0.1` |
| Healthy shop at `6.0.0` | `upgrade-6.0.0-6.0.1.php` | idempotent no-op, records `6.0.1` |
| Broken M1.9 shop at `6.0.0` (no schema) | `upgrade-6.0.0-6.0.1.php` | schema created, `/checkout/onepage/` fixed |
| Un-migrated shop still at `1.6.0.x` | `upgrade-6.0.0-6.0.1.php` | schema created in one step |

Supersedes the original `maho_setup` approach (c6ed0f0): same outcome, but kept inside the normal DB-upgrade pipeline rather than the `maho_setup` channel.

fballiano:
hey @claudiobianco grazie! I think I finally have the final version 😅

PR #951, opened 2026-05-26
Fixed address-book delete: customFormSubmit POST instead of GET 

claudiobianco:
# PR: Address-book delete uses POST instead of GET

**Target repo:** `mahocommerce/maho`
**Base branch:** `main`

Files changed:
- `app/code/core/Mage/Customer/Block/Address/Book/Item/Remove.php` (new)
- `app/design/frontend/base/default/layout/customer.xml`
- `app/design/frontend/base/default/template/customer/address/book.phtml`
- `app/design/frontend/base/default/template/customer/address/book/item/remove.phtml` (new)

---

## Title

> Fix `405 Method Not Allowed` when deleting an address from the customer address book

## Summary

Clicking "Delete Address" in `/customer/address/` returns `405 Method Not Allowed`
on every stock Maho install.

## Root cause

`Mage_Customer_AddressController::deleteAction()` was restricted to POST in
PR #834 (symfony/routing migration):

```php
#[Maho\Config\Route('/customer/address/delete', methods: ['POST'])]
```

The base/default template still does a GET navigation:

```js
window.location = "<?= $this->getDeleteUrl() ?>id/" + addressId + "/";
```

## Fix

Mirror the wishlist-remove setup that already drives the POST-only
`wishlist/index/remove` route:

* New block `Mage_Customer_Block_Address_Book_Item_Remove` exposes
  `getJs()` returning a `confirmRemoveAddress()` JS helper, analogous to
  `Mage_Wishlist_Block_Customer_Wishlist_Item_Column_Remove::getJs()` /
  `confirmRemoveWishlistItem()`.
* New template `customer/address/book/item/remove.phtml` renders the
  delete link with `if (confirmRemoveAddress()) { customFormSubmit(url,
  {form_key}, 'post') }`, using the global `customFormSubmit()` helper
  from `public/js/varien/js.js` that wishlist already uses.
* `customer.xml` registers the remove block as a child of `address_book`.
* `book.phtml` calls `getChild(...)->setAddress($_address)->toHtml()` per
  row (same shape as the wishlist column's `setItem($item)->toHtml()`)
  and emits the confirm function once via `getJs()` at the bottom.

The legacy `Mage_Customer_Block_Address_Book::getDeleteUrl()` helper is
left untouched for back-compat with third-party themes still calling it.

## Test plan

* Log in as a customer with at least one additional address.
* Click "Delete Address" -> confirm -> address removed, success message shown.
* Before the fix: 405 response page.


fballiano:
Thank you @claudiobianco!! I reverted this change to its smaller format, the wishlist one had to be separated in an external block but this one is never used anywhere else so I preferred to keep it minimal. I tested it and it seems to work correctly.

🙏🙏🙏

PR #954, opened 2026-05-28
Fix UTF-8 mangling in MaliciousCode::linkFilter() via XML PI hint

claudiobianco:
Maybe there is a better way to fix this.

`Mage_Core_Model_Input_Filter_MaliciousCode::linkFilter()` parses the HTML through `DOMDocument::loadHTML()` without an explicit charset. libxml defaults to ISO-8859-1, so every UTF-8 multi-byte sequence in the input is treated as a run of single Latin-1 characters and then re-emitted by `saveHTML()` as their named-entity equivalents:

    "ö" (0xC3 0xB6) -> "&Atilde;&para;"
    "ü" (0xC3 0xBC) -> "&Atilde;&frac14;"
    "ß" (0xC3 0x9F) -> "&Atilde;&#159;"

This affects every caller that runs user-supplied HTML through the filter and the content contains both an `<a>` tag (which triggers the DOMDocument path past the early return) and any non-ASCII text. In practice the most visible victim is `Maho_Blog_Model_Resource_Post::_beforeSave()`, where saving a blog post that contains umlauts and at least one link produces permanently corrupted content in `blog_post_entity.content`. Content without an `<a>` tag is unaffected because of the `stristr($html, '<a ') === false` early return, which is why the bug is easy to miss in casual testing.

Fix: prepend `<?xml encoding="UTF-8">` before the input passed to `loadHTML()` so libxml parses the document as UTF-8 (a well-known PHP libxml workaround). The PI then surfaces in `saveHTML()`'s output, so extend the existing wrapper-strip regex to also drop `<?xml ...>` (matching both `?>`-terminated and `>`-terminated forms libxml emits across versions). The early-return path for inputs without `<a>` is unchanged, so the fast path stays byte-identical to before.

Verified with a standalone reproducer:

    $f = new Mage_Core_Model_Input_Filter_MaliciousCode();
    echo $f->linkFilter('<div>Grüße ö</div><a href="x">l</a>');
    // before: <div>Gr&Atilde;&frac14;&Atilde;&#159;e &Atilde;&para;</div><a href="x" rel="noopener noreferrer" target="_blank">l</a>
    // after:  <div>Gr&uuml;&szlig;e &ouml;</div><a href="x" rel="noopener noreferrer" target="_blank">l</a>

Bildschirmfoto vom 2026-05-29 00-04-45
Bildschirmfoto vom 2026-05-29 00-04-22
Bildschirmfoto vom 2026-05-29 00-04-12

### DIFF
--- a/app/code/core/Mage/Core/Model/Input/Filter/MaliciousCode.php
+++ b/app/code/core/Mage/Core/Model/Input/Filter/MaliciousCode.php
@@ -100,6 +100,10 @@ class Mage_Core_Model_Input_Filter_MaliciousCode
         // is present, which mangles UTF-8 multi-byte sequences during the
         // saveHTML() round-trip (e.g. "ö" becomes "&Atilde;&para;"). Prepend a
         // <?xml encoding> processing instruction so libxml parses as UTF-8.
+        // TODO: when the minimum PHP version reaches 8.4, replace this whole
+        // DOMDocument + XML-PI workaround with \DOM\HTMLDocument::createFromString(),
+        // which parses UTF-8 natively (and drop the <?xml ...> strip from the
+        // wrapper regex below).
         if (!$dom->loadHTML('<?xml encoding="UTF-8">' . $html)) {
             Mage::throwException(Mage::helper('core')->__('HTML filtration has failed.'));
         }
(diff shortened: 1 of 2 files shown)


fballiano:
Grazie mille @claudiobianco!

PR #957, opened 2026-05-29
Fixed resolve CMS noroute reverse-lookup collision in URL generation

hirale:
## Problem

Compiling PHP attributes (`composer dump-autoload`) emitted a warning:

```
Maho: Compiling PHP attributes...
  Reverse lookup collision on "cms/index/noroute": route "cms.index_index" overwrites "cms.index" (getUrl() will resolve to the later route)
```

`Mage_Cms_IndexController::norouteAction` declared **two** `#[Route]` attributes:

```php
#[Maho\Config\Route('/cms', name: 'cms.index')]
#[Maho\Config\Route('/cms/index', name: 'cms.index_index')]
```

The attribute compiler keys its reverse-lookup map by `frontName/controller/action`. Both routes resolve to the same key — `cms/index/noroute` — so the second overwrites the first.
The practical effect: `Mage::getUrl('cms/index/noroute')` no longer produced `/cms/index/noroute/`; it collapsed onto whichever alias (`/cms` or `/cms/index`) compiled last.

## Fix

Consolidate the two attributes into a single route with a constrained optional path variable:

```php
#[Maho\Config\Route('/cms/{_index}', name: 'cms.index', defaults: ['_index' => ''], requirements: ['_index' => 'index'])]
```

- **One route → one reverse-lookup entry**, so the collision is gone.
- **Inbound matching is preserved**: the route still matches `/cms` (default `_index=''`) and  `/cms/index` (`_index` constrained to `index`), and rejects anything else under `/cms/*`.
- **Outbound generation is correct again**: because  `Mage_Core_Model_Url::_generateSymfonyRoutePath()` falls back to legacy path building when a  path variable is missing/empty, `Mage::getUrl('cms/index/noroute')` now yields  `/cms/index/noroute/` instead of an alias.

## Verification

**Compile-time** — `composer dump-autoload` no longer prints the `Reverse lookup collision` warning.

**Outbound URL generation:**

```
Mage::getUrl('cms/index/noroute')  →  http://maho.test/cms/index/noroute/
Mage::getUrl('cms')                →  http://maho.test/
```

**Inbound routing** (verified against the compiled matcher and in a live browser):

| URL | Resolves to | HTTP status |
|-----|-------------|-------------|
| `/cms/` | `cms.index` route, `_index=''` | 404 (noroute page) |
| `/cms/index/` | `cms.index` route, `_index='index'` | 404 (noroute page) |
| `/cms/totally-made-up/` | default router → noroute | 404 (noroute page) |

### DIFF
--- a/app/code/core/Mage/Cms/controllers/IndexController.php
+++ b/app/code/core/Mage/Cms/controllers/IndexController.php
@@ -51,6 +51,7 @@ class Mage_Cms_IndexController extends Mage_Core_Controller_Front_Action
     #[\Override]
     #[Maho\Config\Route('/cms', name: 'cms.index')]
     #[Maho\Config\Route('/cms/index', name: 'cms.index_index')]
+    #[Maho\Config\Route('/cms/index/noroute', name: 'cms.noroute')]
     public function norouteAction($coreRoute = null): void
     {
         $this->getResponse()->setHeader('HTTP/1.1', '404 Not Found');
--- a/composer.lock
+++ b/composer.lock
@@ -1223,16 +1223,16 @@
         },
         {
             "name": "mahocommerce/maho-composer-plugin",
-            "version": "v4.0.29",
+            "version": "v4.0.31",
             "source": {
                 "type": "git",
                 "url": "https://github.com/MahoCommerce/maho-composer-plugin.git",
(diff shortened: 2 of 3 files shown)


fballiano:
Thank you @hirale, I changed the fix slightly, putting it mainly in the composer plugin, hope that's ok! 🙏

PR #959, opened 2026-05-30
Added `Mage_CatalogRule` enabled check before using `catalogrule/rule` model

tmewes:
## Description

When the `Mage_CatalogRule` module is disabled in `app/etc/modules`, the location specified in the PR indicates an error.
This PR adds a check if the module is enabled to fix this problem.

For months now, I have been applying this adjustment as a Composer patch across all the Maho instances I maintain where catalog rules are not required (including in production environments). Rather than remaining confined to my local Composer patches, this change should ideally be integrated directly into the core. :)


fballiano:
why not :-)

PR #960, opened 2026-05-30
Adminhtml Sales Order Create: Set attribute frontend labels instead of store labels

tmewes:
## Description

When having different anguages in adminhtml and some store views, the customer address attributes are displayed in the language of the store view if an order is created via sales oder create in adminhtml, which is very confusing and difficult for the customer service (if the employee does not speak the store view language).

For months now, I have been applying this adjustment as a Composer patch across all the Maho instances I maintain (including in production environments). Rather than remaining confined to my local Composer patches, this change should ideally be integrated directly into the core. :)

Here is an example of a German adminhtml (imagine only German speaking people are creating orders for customers) with a Polish store view. An order should be created in the Polish store view via adminhtml.

### Before

grafik

### After

grafik


fballiano:
great idea @tmewes, merging it now!

PR #962, opened 2026-05-30
Adminhtml Sales Order View: Add payment method existence check

tmewes:
## Description

Another local composer patch which should be added to the core. :)

Currently, no longer existing methods cause an exception in adminhtml sales order view.
A 2nd condition is added to check if a payment method exist before using it to try to create a method instance.

As always, this PR is merely a suggestion. I am also happy with any other approach that resolves the exception.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Sales/Order/View.php
+++ b/app/code/core/Mage/Adminhtml/Block/Sales/Order/View.php
@@ -159,7 +159,11 @@ class Mage_Adminhtml_Block_Sales_Order_View extends Mage_Adminhtml_Block_Widget_
 
         if ($this->_isAllowedAction('creditmemo') && $order->canCreditmemo()) {
             $onClick = Mage::helper('core/js')->getSetLocationJs($this->getCreditmemoUrl());
-            if ($order->getPayment() && $order->getPayment()->getMethodInstance()->isGateway()) {
+            if (
+                $order->getPayment()
+                && Mage::helper('payment')->paymentMethodExists($order->getPayment())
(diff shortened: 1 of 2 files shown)


fballiano:
hey @tmewes, Maho should already handle it, just at a different layer than your patch.

 Rather than guarding each call site, Mage_Payment_Model_Info::getMethodInstance() returns a null-object fallback when the original method is no longer installed:

```
  $unavailable = Mage::getModel('payment/method_unavailable');
  $unavailable->setOriginalCode($this->getMethod() ?: 'unknown');
```

  That fallback is Mage_Payment_Model_Method_Unavailable, which has $_isGateway = false and every capability disabled. So the exact line your PR targets, $order->getPayment()->getMethodInstance()->isGateway(), returns false instead of throwing, and the same protection covers every other getMethodInstance() call site (invoice, capture, void, recurring profiles, etc.) rather than just the credit-memo button.

  This landed in #746 ("Fixed admin order view crash when payment method is no longer available"). It also covers a case paymentMethodExists() wouldn't: when the method's config still exists but the model class is gone.

Is it possible that you didn't notice that it was already working? or am i missing something?

tmewes:
@fballiano Thank you for the detailed explanation! That helps a lot. Honestly I didn't notice that and I will immediately test without the patch. When no patch is needed because everything is already working fine, this is definitely the best solution. :D Sorry for being blind.

tmewes:
@fballiano You're absolutely right! Thank you! This PR is not needed and I can just remove my composer patch. :)

fballiano:
:-))))) thanks for reporting it anyway!

PR #963, opened 2026-05-30
API: Let xmlrpc adapter return response instead of printing

tmewes:
## Description

Another composer patch I have in all instances that are using XML-RPC (`composer require laminas/laminas-xmlrpc`) since Maho 25.11. :)
Laminas server needs another setting to return responses instead of directly printing them.


fballiano:
done! :-)

PR #967, opened 2026-06-01
Fixed `getAllowedTypes()` losing data when cache backend is disabled

claudiobianco:
With the cache disabled (e.g. cache:disable, or a non-persisting backend), getAllowedTypes() returned false instead of the allowlist: _generateCache() wrote to the cache and discarded the built data, the re-read still returned false, and jsonDecode(false) returned false. The result: every {{block type="..."}} directive in transactional emails (incl. cms/block) was silently dropped.

Fix: _generateCache() returns the freshly built array, getAllowedTypes() uses it as fallback when the cache read returns false, and the decoded value is type-checked. Cache-enabled behaviour is unchanged.


fballiano:
Grazie @claudiobianco!!!

PR #970, opened 2026-06-01
Increased `ATTRIBUTE_CODE_MAX_LENGTH` from 30 to 64

tmewes:
## Description

Another try to get rid of some local composer patches. :)

Unfortunately, for large shops with a vast number of attributes, 30 characters are insufficient.
Therefore, it would be great if we could increase the maximum length for attribute codes from 30 to at least 64 directly in the core.


fballiano:
Since flat catalog is deprecated and will be removed (that would be the real problem with long attribute codes) we can merge this no problem

PR #972, opened 2026-06-01
Layered navigation SEO: noindex filtered/paginated pages, canonical defaults, nofollow filter links

hirale:
Implements **Option A: Aggressive Crawl Control** from #482 to eliminate duplicate content and optimise crawl budget for layered navigation, while keeping the indexing and canonical signals consistent with Google's guidance.

The core principle: **a page advertises a canonical *or* a `noindex`, never both.** A filtered or paginated layered-navigation view is suppressed with `NOINDEX,FOLLOW`; an indexable view canonicalizes to the clean category URL. (See "Design note" below for why this differs from the issue's original signal table.)

## Behaviour

| Page type | Meta robots | Canonical | Filter links |
|-----------|-------------|-----------|--------------|
| Base category | category / store default | self (clean URL) | `rel="nofollow"` |
| Filtered | `NOINDEX,FOLLOW` | none | `rel="nofollow"` |
| Paginated (p>1) | `NOINDEX,FOLLOW` | none | `rel="nofollow"` |
| Filtered + paginated | `NOINDEX,FOLLOW` | none | `rel="nofollow"` |

All layered-navigation filters are treated uniformly, including the category (`cat`) filter: its links are `?cat=<id>` query duplicates of the parent category, not the subcategory's canonical URL, so they get the same `NOINDEX,FOLLOW` and `rel="nofollow"` treatment. Subcategories stay discoverable through their own clean URLs in the category children listing.

Precedence rules layered on top:
- An **explicit category `noindex`** is kept verbatim on every view (and the canonical is suppressed). An **explicit category `index`** is kept on the base view but overridden to `NOINDEX,FOLLOW` on filtered/paginated views, since those are duplicates regardless of the base category's indexability.
- A **facet landing page** (opt-in via a registry flag, for a future #971 feature) stays indexable, keeps its own canonical, and renders normal (followed) filter links.

## Changes mapped to the issue's components

1. **NOINDEX,FOLLOW for filtered pages** — `Mage_Catalog_Block_Category_View::getForcedRobots()` returns `NOINDEX,FOLLOW` when any layered-navigation filter is active. `hasActiveFilters()` detects this straight from the request (the `cat` filter plus every globally filterable attribute code), so it does not depend on the leftnav block having applied the layer first. Memoized for the render.
[...]

Issue #482, opened 2026-01-04
Improve SEO for Layered Navigation: Prevent Duplicate Content and Optimize Crawl Budget


fballiano:
## Summary

Implement comprehensive SEO improvements for layered navigation to eliminate duplicate content, optimize crawl budget, and prevent pagination-related 404 errors.

## Problem

Currently, Maho's layered navigation creates SEO issues:

1. **Duplicate Content**: Every filter combination creates an indexable URL
   - `/shoes.html?brand=nike`
   - `/shoes.html?color=red`
   - `/shoes.html?brand=nike&color=red`
   - All indexed separately by search engines

2. **Crawl Budget Waste**: With 5 attributes × 5 values × 4 pages = 12,500 possible URLs per category, search engines waste resources crawling filter combinations instead of actual content

3. **Pagination 404 Risk**: If paginated pages are indexed, they become 404s when product count changes

4. **No Link Control**: Filter links have no `rel="nofollow"`, explicitly inviting crawlers to follow every combination

## Proposed Solution

Implement **Option A: Aggressive Crawl Control** with the following components:

### 1. Automatic NOINDEX,FOLLOW for Filtered Pages

Add `<meta name="robots" content="noindex,follow" />` when any layered navigation filter is active.

**Why:**
- NOINDEX: Prevents filtered page from search results
- FOLLOW: Allows product discovery via links on filtered pages

### 2. Automatic NOINDEX,FOLLOW for Paginated Pages (p > 1)

Add `<meta name="robots" content="noindex,follow" />` to all paginated category pages except page 1.

**Why:**
- Prevents 404 errors when page count changes dynamically
- Products still discovered via crawling pagination links
- No SEO downside since products accessible from base category

### 3. Enable Canonical Tags by Default

Change default configuration from `0` to `1` for `catalog/seo/category_canonical_tag`.

**Canonical Logic:**
- Base category → Self or none
- Filtered pages → Base category URL
- Paginated pages → Base category URL (page 1)
- Filtered + Paginated → Base category URL

### 4. Add rel="nofollow" to Filter Links

Update filter template to add `rel="nofollow"` attribute to filter links.

**Why:**
- Explicit signal to crawlers not to prioritize filter combinations
- Combined with NOINDEX provides maximum control
- Pagination links remain normal (for product discovery)

### 5. Configuration Options

Add new admin configuration section: **Catalog > Catalog > Search Engine Optimization > Layered Navigation SEO**

Options (all enabled by default):
[...]

PR #981, opened 2026-06-03
test(catalog): layered-navigation SEO integration coverage

fballiano:
Restores automated regression coverage for the layered-navigation SEO crawl controls introduced in #972 (originally tracked in #482).

The original test file was removed in #972 because its mere collection tipped the Ubuntu `pdo_sqlite` CI runner over a memory threshold and segfaulted the suite. This replacement is leaner and **render-free** (no layout/template rendering, which the earlier crash iterations implicated), and it is order-independent so it survives the Install suite's mid-run DB resets.

## What it covers

`tests/Frontend/Integration/Catalog/LayeredNavigationSeoTest.php` — 32 tests / 45 assertions:

- **Filterable request vars** — the live `catalog/product_attribute_collection` query discovers the real `cat` + filterable attribute codes
- **Filter detection** — real request params count as filters; toolbar params (pagination/sort/mode/limit) and empty values do not
- **Pagination detection** — via the real toolbar page var (`p > 1`)
- **Forced robots** — `NOINDEX,FOLLOW` on filtered / paginated / `cat` views; an explicit category `noindex` preserved verbatim; an explicit `index` overridden to `NOINDEX,FOLLOW` on duplicate views; #971 facet-landing-page opt-out
- **Canonical decision** — mutual exclusivity with `noindex`, the `category_canonical_tag` flag, and the landing-page opt-out
- **Head fallback** — the real `Mage_Page_Block_Html_Head::getRobots()` chain plus explicit override
- **Config defaults and toggles** — real store config for all four flags

## Design notes

- The current category is a real `Mage_Catalog_Model_Category` populated **in memory**. `getForcedRobots()` / `shouldUseCanonicalTag()` only read `getMetaRobots()` off it, so an unsaved instance hits the same code paths without coupling the test to a persisted sample-data row (earlier suites in the run mutate category rows, which is what made a `load(4)`-based fixture flaky).
- The file is **read-only** (no DB writes; `setConfig` is in-memory, registry is cleared per test), so it cannot pollute other suites.

## Validation

Run through the `pest-with-test-db.php` sqlite harness (the same path CI uses):

- All 32 pass standalone and **within the full suite** — no segfault, suite completes (2147 passed, 212k assertions)
- Pre-existing unrelated failures in `Backend/SalesRule/Model/ObserverTest` on sqlite were confirmed to reproduce without this change and are out of scope here

### DIFF
--- a/tests/Frontend/Integration/Catalog/LayeredNavigationSeoTest.php
+++ b/tests/Frontend/Integration/Catalog/LayeredNavigationSeoTest.php
@@ -64,11 +64,6 @@ describe('Layered Navigation SEO', function () {
             expect($this->block->hasActiveFilters())->toBeTrue();
         });
 
-        test('the category (cat) filter counts like any other facet', function () {
-            $this->request->setParam('cat', '5');
-            expect($this->block->hasActiveFilters())->toBeTrue();
-        });
-
         test('toolbar params (pagination, sort, mode, limit) are not filters', function () {
             $this->request->setParam('p', '2');
             $this->request->setParam('order', 'price');
--- a/tests/Frontend/Integration/Catalog/LayeredNavigationSeoTest.php
+++ b/tests/Frontend/Integration/Catalog/LayeredNavigationSeoTest.php
@@ -1,0 +1,231 @@
+<?php
+
+/**
+ * Maho
+ *
+ * @copyright  Copyright (c) 2026 Maho (https://mahocommerce.com)
+ * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+declare(strict_types=1);
+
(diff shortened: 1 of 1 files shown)


tests/Frontend/Integration/Catalog/LayeredNavigationSeoTest.php
```diff
@@ -0,0 +1,231 @@
...
+        test('a real filterable attribute param is an active filter', function () {
+            $this->request->setParam('color', '123');
+            expect($this->block->hasActiveFilters())->toBeTrue();
+        });
+
+        test('the category (cat) filter counts like any other facet', function () {
+            $this->request->setParam('cat', '5');
+            expect($this->block->hasActiveFilters())->toBeTrue();
+        });
+
```
fballiano:
```suggestion
```

PR #987, opened 2026-06-04
Fix PayPal checkout currency (#985): transact in display currency, with full E2E coverage

### LINKED ISSUE #985: Bug: Checkout displays base currency instead of quote/store display currency
### Description

When a store view has a display currency different from the website's base currency (e.g., base currency is **USD** but the English store view displays **EUR/€**), the checkout shows USD in multiple places instead of EUR. Prices throughout the rest of the site display € correctly.

**Version:** Maho 26.5.0 and main branch from composer dev-main 097764a 26.7.0

**Steps to reproduce**

Image

Image

Image

### Root Cause

There are **two distinct issues**:

#### 1. PayPal OrderBuilder uses base currency methods

The [`Maho_Paypal_Model_Api_OrderBuilder`](vendor/mahocommerce/maho/app/code/core/Maho/Paypal/Model/Api/OrderBuilder.php) and [`Maho_Paypal_CheckoutController`](vendor/mahocommerce/maho/app/code/core/Maho/Paypal/controllers/CheckoutController.php) consistently use `getBase*()` methods which return values in the **website-level base currency**, ignoring the per-store display currency:

| File | Line(s) | Current code | Should be |
|---|---|---|---|
| [`OrderBuilder.php`](vendor/mahocommerce/maho/app/code/core/Maho/Paypal/Model/Api/OrderBuilder.php) | 28 | `getBaseCurrencyCode()` | `getQuoteCurrencyCode()` |
| [`OrderBuilder.php`](vendor/mahocommerce/maho/app/code/core/Maho/Paypal/Model/Api/OrderBuilder.php) | 29 | `getBaseGrandTotal()` | `getGrandTotal()` |
| [`OrderBuilder.php`](vendor/mahocommerce/maho/app/code/core/Maho/Paypal/Model/Api/OrderBuilder.php) | 152-196 | `getBaseCalculationPrice()`, `getBaseRowTotal()`, `getBaseTaxAmount()`, `getBaseShippingAmount()`, `getBaseDiscountAmount()`, `getBaseGrandTotal()` | non-base equivalents |
| [`OrderBuilder.php`](vendor/mahocommerce/maho/app/code/core/Maho/Paypal/Model/Api/OrderBuilder.php) | 201-233 | `getBaseCalculationPrice()`, `getBaseRowTotal()` | non-base equivalents |
| [`CheckoutController.php`](vendor/mahocommerce/maho/app/code/core/Maho/Paypal/controllers/CheckoutController.php) | 377 | `getBaseCurrencyCode()` | `getQuoteCurrencyCode()` |
| [`CheckoutController.php`](vendor/mahocommerce/maho/app/code/core/Maho/Paypal/controllers/CheckoutController.php) | 381 | `buildBreakdown($quote, $currency)` passes base currency | uses quote currency instead |
| [`CheckoutController.php`](vendor/mahocommerce/maho/app/code/core/Maho/Paypal/controllers/CheckoutController.php) | 386 | `getBaseGrandTotal()` | `getGrandTotal()` |

#### 2. Checkout review totals template shows "Grand Total to be Charged" in base currency

The template [`checkout/onepage/review/totals.phtml`](vendor/mahocommerce/maho/app/design/frontend/base/default/template/checkout/onepage/review/totals.phtml) renders an additional row showing the grand total in **base currency**:

```php
<?php if ($this->needDisplayBaseGrandtotal()):?>
<tr>
    <td class="a-right" colspan="<?= $_colspan ?>">
        <small><?= $this->helper('sales')->__('Grand Total to be Charged') ?></small>
    </td>
    <td class="a-right">
        <small><?= $this->displayBaseGrandtotal() ?></small>
    </td>
</tr>
<?php endif ?>
```

The method [`displayBaseGrandtotal()`](vendor/mahocommerce/maho/app/code/core/Mage/Checkout/Block/Cart/Totals.php:122-130) formats using `Mage::app()->getStore()->getBaseCurrency()->format(...)`, which always outputs in the website base currency (USD) regardless of the store view's display currency.

### Expected Behavior

All prices in the checkout should be displayed in the **quote currency** (the display currency of the current store view), matching what the customer sees throughout the rest of the site (product prices, cart totals, etc.).

PayPal API requests should also use the **quote currency** so that the PayPal popup matches the store's displayed prices.

### Fix required

1. **`OrderBuilder.php`:** Replace all `getBase*()` calls with their non-base counterparts
2. **`CheckoutController.php`:** Replace `getBaseCurrencyCode()` and `getBaseGrandTotal()` with quote equivalents
3. **`review/totals.phtml` (optional/design decision):** Remove or fix the "Grand Total to be Charged" row to either show in quote currency or be hidden when the store view already displays the correct currency

fballiano:
## Summary

Fixes #985: when a store view's display currency differs from the website base currency (base **USD**, display **EUR**), PayPal showed and charged the **base** currency instead of the display currency the customer sees. PayPal now transacts in the quote/display currency throughout, the base-currency "amount charged" disclosure is suppressed for PayPal (but kept and reworded for gateways that still charge in base currency), and the whole flow is covered by real-browser end-to-end tests that drive an actual sandbox card payment to the success page and reconcile every amount against PayPal's own records.

## For release notes (behavior change)

> **PayPal now charges in the customer's display currency, not the store base currency, in multi-currency stores.**
>
> Previously, a store with base **USD** and a display currency of **EUR** showed the customer **€315.80** but charged PayPal in **USD ($350.89)**, leaving the customer's bank to convert it at an unpredictable rate plus fees. PayPal now charges the **€315.80** the customer actually saw, honoring the quoted price in the quoted currency.
>
> **Merchant impact:** PayPal payments in multi-currency stores may now arrive in the customer's display currency rather than always in the base currency. Merchants who relied on base-currency settlement should review their PayPal account's currency-conversion settings (PayPal can auto-convert incoming foreign payments to a primary balance).
>
> The base-currency "Grand Total to be Charged" disclosure (shown in multi-currency stores on the checkout review, order view, account history, order emails and PDFs) is now hidden for PayPal, since PayPal charges exactly the displayed total. It is kept and reworded ("Amount charged in USD (your bank may convert at its own rate)") for payment methods that genuinely settle in the base currency, such as Authorize.Net.
>
> **Single-currency stores are unaffected.** Other payment methods continue to charge in the base currency for now.

## The core fix (#985)

PayPal's REST integration was written assuming `base == quote` currency and used `getBase*()` everywhere. Switching it to transact in quote currency required touching every point where Maho talks to PayPal or interprets PayPal amounts, not just the two files named in the issue:

- **`Api/OrderBuilder.php`** - build the whole PayPal order (amount, breakdown, line items) in quote currency so PayPal's "items sum to total" check still passes.
- **`controllers/CheckoutController.php`** - `shippingCallback` and `_assertPaypalOrderMatchesQuote` now compare against quote currency/total; otherwise our own replay-protection check would reject the new quote-currency orders.
- **`Model/Method/Abstract.php`** - admin partial capture sends the order-currency amount + code to match a quote-currency authorization.
- **`Webhook/Handler/CaptureCompleted.php`** - PayPal now reports the captured amount in quote currency, so it is converted back to base before `registerCaptureNotification()` (which expects base). Full synchronous captures stay exact (no rate math); only partials use the rate.
- **Base-currency disclosure row** - rather than charging base and disclosing it, PayPal now charges the displayed total, so the base-currency "Grand Total to be Charged" row is suppressed for PayPal. A new `Mage_Payment_Model_Method_Abstract::chargesInDisplayCurrency()` capability (default `false`; PayPal overrides to `true`) gates it: `Mage_Checkout_Block_Cart_Totals` (checkout review) and `Mage_Sales_Block_Order_Totals` (order view, account history, emails, PDFs) keep the row for base-charging methods and reword it to explain the charge currency and that the customer's bank may convert.

Single-currency stores are unaffected: every non-base getter returns the same value as its base counterpart, and the conversion branches short-circuit on equal currency codes.

## Additional bugs found and fixed

Building the end-to-end test surfaced three more real bugs:

- **clientToken 405** (`sdk-loader.js`): the standard PayPal button failed to initialize because the client-token request used GET against a POST-only route. Token generation is non-idempotent, so the route stays POST and the JS was corrected.
- **Quote not persisted in `createOrder`** (`CheckoutController.php`): `buildFromQuote` reserves the order id on the in-memory quote, but the action saved only the payment. `approveOrder` reloads the quote fresh, so the reserved id was lost and the invoice_id security check rejected every order ("PayPal order does not belong to this cart"). Now the quote is saved so the reserved id survives.
- **Sparse `fields` filter strips invoice_id** (`CheckoutController.php`): `approveOrder` fetched the order with `fields=purchase_units.payments`, which returns only the payments sub-object and drops `invoice_id`/`amount` that the security assertion needs. Now it fetches the full order representation (which still includes payments for a completed order).

Both checkout bugs affect the live Advanced and Standard flows, not just tests.

## Test coverage

**Server-side integration** (`tests/Frontend/Integration/Paypal/`): `OrderBuilderTest` asserts the built order uses quote currency; `SandboxOrderTest` creates a real PayPal sandbox order in EUR and checks status/currency/value.

**Real-browser E2E** (Pest v4 + `pest-plugin-browser` / Playwright, in `tests/Browser/`): a new harness boots `./maho serve` and drives Chromium. `PaypalCheckoutTest` runs the full Advanced-Checkout card flow (inline hosted fields, no popup) through to `checkout/onepage/success` for **both** single (USD/USD) and multi (USD/EUR) currency, asserting:

- every stored order amount (subtotal, tax, shipping, discount) converts from base at the order rate and sums to the grand total, in **both** currencies;
- the order's currency, rate and grand total are correct;
- a **live cross-check against PayPal's API**: the order's recorded status, invoice id, order amount, and the authorization amount/currency PayPal will actually settle all match the placed order.

The EUR order is authorized in **EUR 315.80**, not the USD base (350.89) - exactly what #985 was about; before the fix the cross-check would fail here.

### Test harness notes

- PayPal sandbox credentials come from `PAYPAL_SANDBOX_CLIENT_ID` / `PAYPAL_SANDBOX_CLIENT_SECRET` (CI secrets, or a gitignored `.env.testing` locally; `.env.testing.dist` is the template). They are injected **encrypted** at install time so all suites share one config, mirroring a real store. Sandbox/browser tests **skip cleanly** when the credentials are absent.
- Node artifacts (`node_modules`, `package.json`) are gitignored and regenerated in CI, never committed.
- The whole suite installs on a fixed host/port (`127.0.0.1:8901`) so browser and non-browser tests share one configuration (Chromium under Playwright ignores `/etc/hosts`).

## Verification

- `composer lint` clean (cs-fixer, rector, phpstan level 6).
- All PayPal frontend integration tests green.
- Browser E2E: 3 tests / 72 assertions, both currencies reaching the success page with all amounts reconciled against PayPal.
- Manually confirmed against PayPal's live sandbox API that the recorded authorization matches the stored order (EUR 315.80 / USD 350.89).


fballiano:
Superseded by #1008.

Closing this in favour of a simpler approach. Rather than making PayPal transact in the display currency (which makes PayPal the one currency-exceptional gateway and pulls in base↔display conversions plus per-currency edge handling), #1008 keeps PayPal charging in the base currency like every other Maho gateway and resolves the #985 confusion through a clear checkout disclosure, the same pattern Magento and Shopify use for third-party gateways.

#1008 keeps the genuinely useful parts of this branch: the browser/sandbox test infrastructure and the three checkout bug fixes (clientToken 405, quote persistence, full-order fetch).

PR #989, opened 2026-06-05
fix(adminhtml): Add check for menuNode in getMenuItemLabel

tmewes:
## Description

Another try to get rid of some local composer patches. :)

I use `System > Configuration` in adminhtml very often, so I moved the `Configuration` entry from `System` into an own menu entry to avoid hovering `System` and scroll down to the last entry every time. So far, so good - no problem at all.

But there is a problem when a user without permissions for `Configuration` tries to load the default values of a transactional email template.

This PR fixes this functionality by adding a check of `$menuNode !== false` in `getMenuItemLabel`.

### Steps to reproduce

* Move adminhtml `System > Configuration` to own `Configuration` menu entry:
```
<menu>
    <configuration>
        <title>Configuration</title>
        <sort_order>1100</sort_order>
        <action>adminhtml/system_config/edit/section/dev</action>
    </configuration>
    <system>
        <children>
            <config>
                <disabled>1</disabled>
            </config>
        </children>
    </system>
</menu>
```

* Add a new adminhtml user with permissions for `Transactional Emails`
* Login with that new adminhtml user
* Open `System > Transactional Emails`
* Click on `Add New Template`
* Select a template to load, e.g. `Changed Password or Email`
* Click on the `Load Template` button
* The template isn't loaded because of the following error:
```
ERROR: Warning: Attempt to read property "title" on false  in /var/www/html/app/code/core/Mage/Admin/Model/Config.php on line 144
```

### DIFF
--- a/app/code/core/Mage/Admin/Model/Config.php
+++ b/app/code/core/Mage/Admin/Model/Config.php
@@ -138,7 +138,7 @@ class Mage_Admin_Model_Config extends \Maho\Simplexml\Config
     {
         $moduleName = 'adminhtml';
         $menuNode = $this->getAdminhtmlConfig()->getNode('menu/' . str_replace('/', '/children/', trim($path, '/')));
-        if ($menuNode->getAttribute('module')) {
+        if ($menuNode !== false && $menuNode->getAttribute('module')) {
             $moduleName = (string) $menuNode->getAttribute('module');
         }
         return Mage::helper($moduleName)->__((string) $menuNode->title);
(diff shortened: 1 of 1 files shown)


fballiano:
Hey @tmewes, thank you for this :-)


  I pushed a small follow-up to your branch (3e8afbfe93) that completes the fix. The !== false guard you added covers the getAttribute('module') call, but $menuNode is also dereferenced one line below in (string) $menuNode->title so when the node is missing, that line still emits the exact Attempt to read property "title" on false warning. That warning is harmless in production (logged, returns ''), but mageCoreErrorHandler throws on warnings in developer mode (Mage/Core/functions.php:117), so the load would still fail there.

  I switched it to an early return so both dereferences are guarded.

can you give this version a check?

tmewes:
@fballiano I tested your update and it works fine. Thank you! :)

PR #991, opened 2026-06-05
Refresh minicart on bfcache restore via pageshow event

### LINKED ISSUE #990: [Bug] Minicart and cart page show stale data when restored from browser bfcache
#### Summary

When a user removes a product from the cart (via the minicart sidebar) and then navigates back using the browser's back button, the page is often restored from the browser's **Back-Forward Cache (bfcache)**. This causes the minicart to display stale data — the removed product still appears in the sidebar.

#### Steps to Reproduce

1. Add a product to the cart from home page
2. Navigate to a different page
3. Open the minicart dropdown
4. Remove the product via AJAX (`/checkout/cart/ajaxDelete`)
5. Press the browser's **Back** button

**Expected:** The minicart shows the updated cart (product removed).

**Actual:** The page is served from bfcache with the old DOM/JS state — the removed product is still visible in the minicart.

#### Root Cause

Previously, OpenMage globally sent a `Cache-Control: no-cache` header on all frontend pages, which prevented this kind of stale-data issue by ensuring the browser would revalidate content with the server on bfcache restore.

In Maho, this global no-cache header was removed. As a result, modern browsers aggressively cache pages in bfcache and serve them on back/forward navigation without any server revalidation, causing dynamic sections like the minicart to display stale data.

#### Proposed Solutions

We are opening this issue to discuss two approaches:

**1. Configuration option to restore global `Cache-Control: no-cache` header (main proposal)**

Add a system configuration option to restore the `Cache-Control: no-cache` header globally (like OpenMage).

This configuration option would allow store administrators to control whether the `Cache-Control: no-cache` header is sent on all frontend pages. The option should be placed under *System → Configuration → General → Web* (or similar).

- When **enabled** (recommended, restores OpenMage behavior): sends `Cache-Control: no-cache` on all frontend pages, preventing the browser from serving stale cached content when navigating back
- When **disabled**: allows more aggressive browser caching for better performance on static pages, but risks stale data on dynamic sections

**Default value:** `enabled` (to restore OpenMage compatibility).

**2. Implement `pageshow` event handling in the base template (suggestion for follow-up)**

As a complementary improvement, add a [`pageshow`](https://developer.mozilla.org/en-US/docs/Web/API/Window/pageshow_event) event listener in the base default template — specifically in the `skin/frontend/base/default/js/minicart.js` class — to detect bfcache restores and refresh dynamic sections (sidebar, minicart) without a full page reload.

Example approach:
```javascript
// In Minicart.init()
window.addEventListener('pageshow', (event) => {
    if (event.persisted) {
        this.refresh();
        // Fetch updated minicart content via XHR
    }
});
```

fballiano:
Closes #990

## Problem

When a page is restored from the browser's back-forward cache (bfcache), dynamic sections like the minicart are served from the cached DOM with no server revalidation. After removing a product via the minicart's ajax delete and pressing Back, the removed product still appears in the sidebar.

## Approach

Per the discussion on #990, this implements option 2 (the low-pain JS refresh) rather than restoring a global `Cache-Control: no-cache` header. On a bfcache restore we refresh only the minicart data via XHR, with no full page reload.

## Changes

- **New read-only endpoint** `GET /checkout/cart/ajaxContent` (`Mage_Checkout_CartController::ajaxContentAction`) returning `{ success, qty, content }` for the current minicart. Mirrors the existing `ajaxDelete`/`ajaxUpdate` response shape but mutates nothing.
- **`pageshow` handler + `refresh()`** in `minicart.js`. `init()` registers a `pageshow` listener (guarded so re-running `init()` doesn't stack listeners); when `event.persisted` is true it calls `refresh()`, which fetches the new endpoint and reuses the existing `updateCartQty()` / `updateContent()` plumbing. No-ops if `contentUrl` is unset or the request fails.
- **Templates** `minicart.phtml` and `minicart/items.phtml` pass `contentUrl` to the `Minicart` constructor.

## Notes

The `qty` count span on the page uses `getSummaryCount()` while all ajax actions (including this new one) return `getSummaryQty()`. I matched existing ajax behavior rather than introduce a third pattern; that pre-existing inconsistency is out of scope here.

### DIFF
--- a/app/code/core/Mage/Checkout/controllers/CartController.php
+++ b/app/code/core/Mage/Checkout/controllers/CartController.php
@@ -1059,6 +1059,20 @@ class Mage_Checkout_CartController extends Mage_Core_Controller_Front_Action
     }
 
     /**
+     * Return current minicart content, used to refresh stale data (e.g. after a bfcache restore)
+     */
+    #[Maho\Config\Route('/checkout/cart/ajaxContent', name: 'checkout.cart.ajaxContent', methods: ['GET'])]
+    public function ajaxContentAction(): void
+    {
+        $this->loadLayout();
+        $this->getResponse()->setBodyJson([
+            'success' => 1,
+            'qty' => $this->_getCart()->getSummaryQty(),
+            'content' => $this->getLayout()->getBlock('minicart_content')->toHtml(),
+        ]);
+    }
+
+    /**
      * Get customer session model
      *
      * @return Mage_Customer_Model_Session
--- a/app/code/core/Mage/Checkout/controllers/CartController.php
+++ b/app/code/core/Mage/Checkout/controllers/CartController.php
@@ -1065,11 +1065,13 @@ class Mage_Checkout_CartController extends Mage_Core_Controller_Front_Action
     public function ajaxContentAction(): void
     {
         $this->loadLayout();
-        $this->getResponse()->setBodyJson([
-            'success' => 1,
-            'qty' => $this->_getCart()->getSummaryQty(),
-            'content' => $this->getLayout()->getBlock('minicart_content')->toHtml(),
-        ]);
+        $this->getResponse()
+            ->setHeader('Cache-Control', 'private, no-store, no-cache, must-revalidate', true)
+            ->setBodyJson([
+                'success' => 1,
+                'qty' => $this->_getCart()->getSummaryQty(),
+                'content' => $this->getLayout()->getBlock('minicart_content')->toHtml(),
+            ]);
     }
 
     /**
--- a/app/design/frontend/base/default/template/checkout/cart/minicart.phtml
+++ b/app/design/frontend/base/default/template/checkout/cart/minicart.phtml
@@ -27,7 +27,8 @@
     document.addEventListener('DOMContentLoaded', function() {
         if (typeof window.minicart === 'undefined') {
             window.minicart = new Minicart({
-                formKey: "<?= $this->getFormKey() ?>"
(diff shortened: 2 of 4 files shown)


fballiano:
Thanks again @empiricompany, your analysis in #990 nailed the root cause and the `pageshow` snippet you sketched is essentially what this PR ships.

On the config option to restore a global `Cache-Control: no-cache` header (your option 1): we went back and forth on it and decided not to add it, and I want to explain the reasoning since it was your main proposal.

- The global no-cache header doesn't just stop stale minicarts, it also disables the browser's back-forward cache (bfcache) for the whole storefront. bfcache makes back/forward navigation instant (no network, no re-render), so turning it off site-wide is a real, measurable performance regression for every visitor, in order to fix one dynamic widget.
- The modern, browser-vendor-recommended fix for exactly this case is the `pageshow` + `event.persisted` refresh, which is your option 2. It keeps bfcache and its speed benefit while guaranteeing the cart is fresh on restore. So rather than treating it as a follow-up, we promoted it to the primary fix.
- A toggle would also add config surface that most stores should never flip (flipping it makes the site slower), which is ongoing maintenance for negative value. If a specific deployment ever genuinely needs `no-store` everywhere, that's a one-line web-server header without a Maho option.

So this PR refreshes the minicart sidebar via a small read-only XHR on bfcache restore, and does a full reload only on the cart page itself (where the grid can be stale too). Same end result you asked for, without giving up bfcache globally.

Really appreciate the detailed report and the concrete suggestions, both options were well thought out and made this easy to implement.

PR #995, opened 2026-06-07
Fix: Keep failed logins on the login form

claudiobianco:
# Keep failed logins on the login form

A failed login on `/customer/account/login/` can redirect to
`/checkout/onepage/` instead of redisplaying the login form, losing the
"invalid login or password" message.

`_loginPostRedirect()` runs after every login attempt and follows the session's
`beforeAuthUrl`. On a failed attempt it left that URL untouched and redirected
there anyway. `OnepageController::indexAction()` sets `beforeAuthUrl` to
`checkout/onepage` on every guest visit to the checkout, so a wrong password
bounced the visitor to the checkout.

## Reproduce

1. As a guest, add a product to the cart and open `/checkout/onepage/` (sets `beforeAuthUrl = checkout/onepage`).
2. Go to `/customer/account/login/`.
3. Submit a wrong password.
4. Bug: redirected to `/checkout/onepage/` instead of back to the login form with the error.

## Fix

Bail out of `_loginPostRedirect()` when the customer is not logged in and return
to the login form. `beforeAuthUrl` stays in the session, so a later successful
login still follows the intended destination.

Covered by `tests/Frontend/Integration/Customer/AccountLoginRedirectTest.php`.


fballiano:
@claudiobianco thanks for this one too!

after https://github.com/MahoCommerce/maho/pull/987 lands (hopefully today) we'll have a browser testing functionality which i think it's perfect for cases like this one ;-)

PR #996, opened 2026-06-07
Allow GET on paypal/checkout/clientToken route (fixes 405)

claudiobianco:


claudiobianco:
Will be fixed with #987 (other way around)

PR #1012, opened 2026-06-10
Adminhtml: Fix invoice view without payment data

tmewes:
## Description

This PR is kind of a followup to #15. On local environments I always work with anonymized database dumps (generated by https://github.com/Smile-SA/gdpr-dump).

I tested the current main branch and saw that invoices can't be viewed when the related payment data is removed.
These are the related errors:
```
Fatal error: Uncaught Error: Call to a member function canRefundPartialPerInvoice() on false in /var/www/html/vendor/mahocommerce/maho/app/code/core/Mage/Adminhtml/Block/Sales/Order/Invoice/View.php on line 59
```
When this is solved, the next error occurs:
```
Fatal error: Uncaught Error: Call to a member function canVoid() on false in /var/www/html/vendor/mahocommerce/maho/app/code/core/Mage/Sales/Model/Order/Invoice.php on line 342
```

This PR solves the rendering of the invoice view page in adminhtml without the related payment data.
As always, this PR is just a proposal. I'm grateful for any working version. :)

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Sales/Order/Invoice/View.php
+++ b/app/code/core/Mage/Adminhtml/Block/Sales/Order/Invoice/View.php
@@ -55,7 +55,10 @@ class Mage_Adminhtml_Block_Sales_Order_Invoice_View extends Mage_Adminhtml_Block
 
         $orderPayment = $this->getInvoice()->getOrder()->getPayment();
 
-        if ($this->_isAllowedAction('creditmemo') && $this->getInvoice()->getOrder()->canCreditmemo()) {
+        if ($orderPayment !== false
+            && $this->_isAllowedAction('creditmemo')
+            && $this->getInvoice()->getOrder()->canCreditmemo()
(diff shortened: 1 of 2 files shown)


fballiano:
Thanks for the fix, @tmewes, and for the clear reproduction with anonymized dumps. This is a real gap and the `!== false` guards are exactly right since `Order::getPayment()` is documented as `Mage_Sales_Model_Order_Payment|false`.

I pushed one small addition on top: `Mage_Sales_Model_Order_Invoice::canCapture()` had the same unguarded `getPayment()->canCapture()` call. It sits on the same admin invoice-view page (via the Capture button check), so it would fatal the same way for any invoice not in PAID/CANCELED state. Your test invoice was PAID, so `canCapture()` short-circuited before reaching `getPayment()`, which is why you only saw the creditmemo and void errors. The guard mirrors your `canVoid()` change so the page renders without payment data in every invoice state.

Merging now.

PR #1023, opened 2026-06-11
Fixed frontend `Add to Compare`

### LINKED ISSUE #1022: Frontend: Add to Compare does not work
## Description

Currently, the `Add to Compare` button on frontend product detail pages doesn't work. When the button is clicked, the page just reloads without any message and no product is added to the compare list.

### Steps to reproduce

* Set `catalog/recently_products/enabled_product_compare` to `Yes` in `System > Configuration > Catalog > Catalog` adminhtml settings
* Open a product detail page in frontend
* Click on `Add to Compare` button
Image

* A `POST` request is initiated to something like `/catalog/product_compare/add/?product=1&uenc=aHR0cHM6Ly9tYWhvLWdpdGh1Yi5kZGV2LnNpdGUvdGVzdC8,` with only `form_key` as `POST` data
* The product detail page will reload
* Nothing else happens

fballiano:
Fixes #1022.

## Problem

Clicking **Add to Compare** on a frontend product page reloaded the page and added nothing.

## Root causes (two, both required for the failure)

1. **Visitor singleton never populated (primary).** The `log/visitor` observers (`initByRequest`, `saveByRequest`, `bindCustomer*`, `bindQuote*`) used the `#[Observer]` default `type: 'model'`, so they ran on throwaway `Mage::getModel('log/visitor')` instances rather than the shared `Mage::getSingleton('log/visitor')`. The compare controller authorises a guest's compare item with `Mage::getSingleton('log/visitor')->getId()`, which was always `null`, so the add was dropped via `_redirectReferer()`. The same inversion created a new `log_visitor` row on nearly every request.

   Fix: mark those observers `type: 'singleton'`, restoring the M1 shared-instance behaviour for this stateful model. (The dispatcher's `getModel` default is intentional and unchanged; only the observers that depend on a shared instance opt in.)

2. **Form POST aborted by the anchor (secondary).** The compare/wishlist `<a href="#">` `onclick="customFormSubmit(...)"` had no `return false;`, so the default `#` navigation raced and aborted the JS-built form POST before it reached the controller. Added `return false;` to the compare and wishlist handlers in `view/addto.phtml` and `list.phtml` (the cart `<button type="button">` elements need no change).

This also repairs the guest -> customer comparison merge on login, which reads the same visitor singleton.

## Tests

New `tests/Frontend/Integration/Catalog/CompareAddTest.php`:
- the predispatch observer populates the shared `log/visitor` singleton;
- a guest add-to-compare creates a `catalog_compare_item` row end to end.

Both fail before the fix, pass after.

## Verification

- Backend suite: 2117 passed. Frontend suite: 113 passed (incl. new tests).
- `php-cs-fixer` clean, `phpstan` (level 6) OK.
- Manual browser check: clicking Add to Compare adds the product and shows "The product ... has been added to comparison list."; one `log_visitor` row per session instead of one per request.

### DIFF
--- a/tests/Frontend/Integration/Catalog/CompareAddTest.php
+++ b/tests/Frontend/Integration/Catalog/CompareAddTest.php
@@ -1,0 +1,101 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: 2026 Maho <https://mahocommerce.com>
+ * SPDX-License-Identifier: OSL-3.0
+ */
+
+declare(strict_types=1);
+
+use Symfony\Component\HttpFoundation\Request as SymfonyRequest;
+use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\Storage\MockArraySessionStorage;
+
+uses(Tests\MahoFrontendTestCase::class);
+
+/**
+ * Regression coverage for issue #1022 ("Add to Compare does not work").
+ *
+ * The frontend visitor is initialised by Mage_Log_Model_Visitor::initByRequest on
+ * controller_action_predispatch. The compare controller then reads the *same*
+ * Mage::getSingleton('log/visitor') to decide whether a guest may own a compare item.
+ *
+ * The #[Observer] attribute defaults to type 'model' (a fresh instance per dispatch),
+ * whereas the original M1 XML observer had no <type> and therefore defaulted to
+ * 'singleton'. Under 'model' the observer initialises a throwaway visitor, leaving the
+ * shared singleton empty, so a guest's getId() is null and the add is silently dropped.
+ */
+function startGuestSession(): void
+{
+    $session = new Session(new MockArraySessionStorage());
+    $session->start();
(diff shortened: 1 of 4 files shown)


tests/Frontend/Integration/Catalog/CompareAddTest.php
```diff
@@ -0,0 +1,101 @@
...
+        ->setPageSize(1)
+        ->getFirstItem();
+}
+
+it('populates the shared log/visitor singleton on controller_action_predispatch', function () {
+    startGuestSession();
+
+    $request = Mage::app()->getRequest();
+    $request->setRouteName('catalog');
+    $controllerAction = new class($request, new Mage_Core_Controller_Response_Http()) extends Mage_Core_Controller_Front_Action {};
```
fballiano:
```suggestion
    $controllerAction = new class ($request, new Mage_Core_Controller_Response_Http()) extends Mage_Core_Controller_Front_Action {};
```

PR #1047, opened 2026-06-20
Overhauled the Feed Manager Google Shopping preset and fixed Mapper / Rule / Transformer

mageaustralia:
Bundles a Google Shopping preset expansion alongside several independently-useful FeedManager fixes that surfaced while pushing real catalog data through it. Happy to split into smaller PRs on request.

### Google Shopping preset (Google.php + Xml.php)
- Expand `$_requiredAttributes` + `$_optionalAttributes` to cover the full practical Google Merchant Center product spec (9 required + 55 optional total). Tightens descriptions on conditionally-required fields per the spec.
- Add per-field length-cap and `strip_tags` chains so descriptions ship as plain text and individual fields stay under their documented limits (e.g. title 150, description 5000, brand 70, gtin 50, custom_label_0..4 100, additional/lifestyle/mobile/ads_redirect 2000).
- Default `g:availability` source to `is_in_stock` with a conditional transformer that emits `in_stock` / `out_of_stock` literals (Google rejects "In Stock").
- Default `g:identifier_exists` to a conditional on gtin so it emits `yes` / `no` rather than empty / unknown.
- Default `g:sale_price_effective_date` to a rolling now / +90 days range using a new `relative_date_range` transformer (see below).
- Platform-aware `_getDefaultXmlStructure()` so a brand-new feed with a platform set picks up that adapter's full attribute list immediately rather than the hardcoded short fallback.
- `XmlBuilder.loadPreset(platform, {force:true})` opt-in flag so programmatic callers (e.g. the General-tab Feed Template dropdown) can bypass the "are you sure?" confirm. The interactive Structure-panel preset dropdown keeps the safety prompt.

### Mapper.php fixes
- **`use_parent` ignored on `taxonomy` source.** Only the ATTRIBUTE branch of `_getSourceValue`'s match received `$useParentMode`. Variants with no category assignment (typical for shoe/clothing children) emitted empty `<g:google_product_category>` even when the parent was mapped. Now passes the mode through and walks to parent category IDs.
- **`image_1`..`image_10` source values never resolved.** The AbstractBuilder offers "Gallery Image 1..10" as picker options but the data array had no matching keys, so every `additional_image_link` mapping using them emitted empty. Populated from `_getAdditionalImages()`.
- **Media gallery not loaded.** `getMediaGalleryImages()` returned null because the generator's collection doesn't trigger the media_gallery backend; force-load it before reading.
- **Parent gallery missing from parent_data extraction.** Use_parent fallback on additional_image_link couldn't reach the parent's gallery because `_extractParentProductData` only mirrored a hardcoded subset. Added the same `_getAdditionalImages` block.
- **Main image_link URL duplicated in additional_image_link.** `_getAdditionalImages()` returned every media gallery entry including the one tagged as the main image, so feeds emitted the same URL in both `<g:image_link>` and `<g:additional_image_link>`. Google's spec wants additionals to be extras only. Drop the URL that matches `getImageUrl('image')`.

### Rule conditions (Product.php)
- Add `is_empty` / `is_not_empty` operators to `Maho_FeedManager_Model_Rule_Condition_Product`. The closest existing operator is `==` with a blank value, which behaves inconsistently on select/int EAV attributes (a missing row doesn't match `value=''`). validateAttribute treats null, empty string, and empty array as "empty"; numeric 0 is intentionally NOT empty because for many attributes 0 is legitimate. Also overrides `loadOperatorOptions` to extend the input-type allow-list so the new operators appear in the per-row dropdown for select/multiselect/etc.

### Transformer.php + feedmanager-mapping.js
- **Whitespace stripped from option values.** Both parsers trimmed values when parsing the pipe-separated chain string, so `prepend_append:append= kg` (with a leading space, used for unit formatting like `1.3 kg` per Google's requirement) became `append=kg` and emitted `1.3kg`. Keep trimming the key, pass the value through verbatim.
- Register the new `relative_date_range` transformer in the dates_urls category.

### Mapper change is the only path-sensitive piece
FeedManager evaluates conditions per-product in PHP (`ProductWriterTrait::_validateProduct` → `$conditions->validate($product)`), so only the in-memory `validateAttribute` hook is needed for the is_empty operators — no SQL builder work.

PHPStan / cs-fixer / rector all clean across all touched files.


fballiano:
Thank you @mageaustralia!

I've to leave in 2 min but I'm seeing this:


  1. JSON/CSV preset confirm still fires — General.php passes {force:true} to all three builders, but only Xml.php's loadPreset was updated. Json.php:368 and Csv.php:252 are still
  function(platform) and still hit confirm(...). Apply the same options.force guard to both (or scope the PR title to XML only).
  3. identifier_exists default is too aggressive — emits no whenever gtin is empty, even when brand+MPN exist (Google still treats that as identifier-present). Either gate on
  gtin+mpn+brand or document loudly that the default assumes GTIN-only catalogs.
  4. Verbatim option value can leak whitespace — $options[trim($kv[0])] = $kv[1] now preserves trailing space for all keys, so a stray compare_value=1  would break conditional's (string)
  equality. The  kg use case is real; consider trimming only for known-numeric keys, or just confirm no existing saved chains rely on the old trimming.

  Lower priority / not blocking:
  - media_gallery force-load per product, confirmed it issues one gallery load per product object (cached on the object, so repeat _extractProductData calls are cheap). Necessary for the feature; just worth a timing check on a large catalog. The PR's in-memory-only claim for the rule operators also checks out — ProductProcessorTrait filters via
  $conditions->validate($product), not SQL.
  
what do you think of these?

mageaustralia:
1. JSON/CSV - Yes, my bad, I didn't check it - I was mostly working through our own feeds and fixing these as I went and bundled it all up to push it at the end.

2. identifier_exists default is too aggressive  - Maybe you're right, the defaults simply weren't there before so I saw this as an improvement. I can't quite think how a rule would work that checks through everything correctly. Possibly a 'Dynamic rule' would be needed to handle this more complex scenario. 

3. I've made a change so that on prepend we do an ltrim and on append, we do an rtrim. The provides trimming where it's most important and still allows the spaces where you would need them.

PR #1058, opened 2026-06-23
Added RFC 8058 one-click unsubscribe to newsletter

claudiobianco:
Newsletter mails carried no List-Unsubscribe header, and the unsubscribe endpoint was GET-only, so the one-click POST that Gmail/Yahoo require from bulk senders could not work.

- Mage_Newsletter_Model_Observer emits List-Unsubscribe (the existing unsubscribe URL) plus List-Unsubscribe-Post on email_template_send_before, gated on the 'subscriber' variable so transactional mail is untouched.
- unsubscribeAction now also accepts POST and answers the one-click request with a bare status code instead of a redirect; the human GET link is unchanged. POST returns 200 on success and on an invalid/expired code alike (no enumeration oracle, no client-side error on stale links); only an unexpected server-side failure returns 503 so the client retries.
- NewsletterUnsubscribeMethodsTest asserts the compiled matcher resolves /newsletter/subscriber/unsubscribe under both GET and POST (405 regression guard); UnsubscribeOneClickTest exercises the action end to end: a valid POST flips the subscriber to UNSUBSCRIBED and returns 200, a wrong-code POST leaves it subscribed and still returns 200.

### DIFF
--- a/app/code/core/Mage/Newsletter/Model/Observer.php
+++ b/app/code/core/Mage/Newsletter/Model/Observer.php
@@ -13,12 +13,13 @@ use Symfony\Component\Mime\Email;
 class Mage_Newsletter_Model_Observer
 {
     /**
-     * Add RFC 8058 List-Unsubscribe headers to outgoing newsletter mails.
+     * Add RFC 8058 List-Unsubscribe headers to outgoing newsletter broadcasts.
      *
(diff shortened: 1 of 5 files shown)


fballiano:
Thank you @claudiobianco, it's a great idea!

PR #1060, opened 2026-06-24
fix(core): poll config cache save lock instead of blocking on it

### LINKED ISSUE #1050: Overwhelmed By GetLock Queries (core_config_cache_save_lock) After Cache Flush
# Description
We (as in @tmewes and I) moved with our shops from 26.3 to 26.5 (hurrah 🎉) successfully but encountered a roadblock during initial deployment. Besides that the first metrics look really good, better than before 👍. 

In previous versions of Maho we already have noticed a significant spike in response times (caused by DB load) for a short while immediately after the cache is flushed. Of course when there are no cache entries it is to be expected to have higher response times, we are aware of that. However, with with 26.5 our shop was simply overloaded. This could be only resolved by restarting the DB altogether. 
We found the root cause to be a lot of those GET_LOCK queries.
Image

The relevant code requesting those locks lives in `Mage_Core_Model_Config::getCacheSaveLock()`: 
Image
One can see the wait times from either CLI (60) or default (3). We don't use `MAGE_CONFIG_CACHE_LOCK_WAIT`. 
The method is called four times, whereof 1 and 4 are not important, those are from the action in Adminhtml and CLI command respectively. 
Image
The two calls in question first check Image. 

In conclusion immediately after the cache is cleared containing the module information, a lot of requests try to request the lock which leads to a pileup of requests blocking PHP and DB resources. We reached a state where every call that was processed took longer than three seconds and a lot of requests resulted in status codes in the 500 range. 

What we are looking for is input how to possible tackle this issue. We considered decreasing the wait time where zero would mean non-blocking and would immediately return the lock if possible, but this could lead to errors if the 2nd parameter of the method is not true. Deleting the relevant cache less often would mitigate the problem as well. We expect to have even more requests in the future which would amplify the issue further.

# Additional Context
We currently do the cache flush directly in our cache redis instance as part of our deployment, but using the Maho command in CLI or in Adminhtml shows the same characteristics.

fballiano:
## Problem

Closes #1050.

After a config cache flush, a stampede of concurrent requests each call:

```php
$connection->getLock('core_config_cache_save_lock', $waitTime);
```

A blocking `GET_LOCK` parks the request inside MySQL in a **"User lock"** wait state, holding one busy DB thread per waiting request for the entire timeout. Under load this saturates DB threads/connections — the reporter saw the processlist fill with `GET_LOCK` queries and had to restart the database to recover.

## Change

Acquire the lock by **polling** with non-blocking `GET_LOCK(..., 0)` attempts and sleeping 1s in PHP between tries, instead of one long blocking acquire:

```php
$deadline = microtime(true) + (float) $waitTime;
while (!$connection->getLock('core_config_cache_save_lock', 0)) {
    if (microtime(true) >= $deadline) {
        // ...existing failure handling (ignoreFailure / CLI throw / 503)...
    }
    sleep(1);
}
```

External behavior is unchanged — it still blocks until the lock is held or the deadline passes, with the same `ignoreFailure` / CLI-throw / 503 paths. The difference is that a waiting request no longer pins a busy DB thread in lock-wait; between tries it sleeps in PHP, so the DB only sees one trivial non-blocking lock check per waiter per second instead of a continuously parked thread.

## Scope

- Single method touched: `Mage_Core_Model_Config::getCacheSaveLock()`.
- No call-site changes; the double-checked locking, `applyAllUpdates()` under the lock, and the lock staying on the DB (required for multi-frontend coordination) are all untouched.

## Note

This narrows the failure mode but doesn't make cache-miss bootstraps free — the rebuild wait is inherent. The most effective operational mitigation remains warming the config cache on deploy (one bootstrap with no concurrency) rather than flushing into live traffic.

### DIFF
--- a/app/code/core/Mage/Core/Model/Config.php
+++ b/app/code/core/Mage/Core/Model/Config.php
@@ -452,12 +452,6 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
     /**
      * Call before building and saving cache to ensure only one process can save the cache
      *
-     * Acquires the lock by polling with non-blocking GET_LOCK attempts and
-     * sleeping in PHP between tries, rather than one blocking GET_LOCK that
-     * parks a DB connection in "User lock" wait for the whole timeout. Under a
-     * post-flush stampede the blocking form keeps one busy DB thread per waiting
-     * request, which can overwhelm the server (see issue #1050).
-     *
      * If failed to get cache lock:
      *   - CLI: throws exception
      *   - Other: 503 error
@@ -475,9 +469,11 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
         if (!$connection) {
             return;
         }
-        $deadline = microtime(true) + (float) $waitTime;
+        // Poll with non-blocking GET_LOCK + sleep instead of one blocking call, which
+        // would park a busy DB thread per waiting request and pile up under load (#1050).
+        $deadline = time() + $waitTime;
         while (!$connection->getLock('core_config_cache_save_lock', 0)) {
-            if (microtime(true) >= $deadline) {
+            if (time() >= $deadline) {
                 if ($ignoreFailure) {
                     return;
                 }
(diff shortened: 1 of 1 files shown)


fballiano:
with 1-second polling granularity, a waiter can take up to ~1s longer to acquire than the old blocking call would in the
  common case where the lock is released mid-second.
  
  At the moment I don't see a real improvement with this PR so I'm closing it

PR #1062, opened 2026-06-25
Added multi-website gift card associations, customer "Check Balance" page, and rate-limited lookups

mageaustralia:
A gift card was tied to exactly one website by a `giftcard.website_id` scalar. That does not match how multi-website merchants use gift cards, where head office issues a card and the customer spends it across any of the brand's storefronts. This replaces the scalar with a junction table, then builds the admin and customer surfaces that the multi-website model needs.

## Multi-website associations

- New `giftcard_website` junction: PK `(giftcard_id, website_id)`, indexed by `website_id`, foreign keys to `giftcard` and `core_website`, both `ON DELETE CASCADE`.
- `Maho_Giftcard_Model_Giftcard::getWebsiteIds()` / `setWebsiteIds()` are backed by the junction. `isValidForWebsite()` is now a membership check and fails closed when a card has no associations.
- The resource model syncs the junction inside the save transaction, and skips the delete/insert churn when the posted selection matches what is stored. An explicitly empty set throws `Mage_Core_Exception` rather than orphaning the card silently.
- All associated websites must share one base currency, since the balance is denominated in it. Re-scoping an existing card cannot re-denominate its balance.
- Grid and collection filter by membership through `Collection::addWebsiteIdsFilter()`, which uses a subquery rather than `FIND_IN_SET`, so it works on MySQL, PostgreSQL and SQLite alike.

Deleting a website cascades the junction rows away and leaves the card behind with no associations. Every read surface tolerates that orphan and reports a null currency instead of throwing, so the admin can still open the card and re-associate it.

## Schema migration, 1.0.0 to 1.1.0

`upgrade-1.0.0-1.1.0.php` moves each card's existing association into the junction, one row per card, which preserves the pre-1.1.0 scoping exactly. It then drops the `website_id` column that the declarative pass leaves behind. The column-existence guard and the `LEFT JOIN` keep fresh installs, re-runs and interrupted runs idempotent.

## Admin

- The single-website select becomes a multiselect, editable on existing cards, with a note showing the base currency of the current selection.
- The edit page gains a card-scoped, read-only Transaction History tab next to the General form, with AJAX paging through `historyGridAction`. It is hidden on the create form. The grid uses its own id so saved grid state does not bleed into the global history page.
- Balance fields now render up to 4 decimals instead of rounding to 2, reusing `formatPriceForInput()` from #1263. The columns are `decimal(12,4)`, so rendering them at 2dp made an untouched field post back a rounded value, which overwrote the balance and logged a false adjustment. `adjustBalance()` is now the only writer of an existing card's balance.

## Customer "Check Gift Card Balance" page

A new My Account page where a logged-in customer looks up the remaining balance and expiry of a card by code. The template lives in `base/default` and uses stock theme classes only.

- The lookup result travels through `giftcard/session`, not query params, so a checked code never reaches browser history or referer logs. The read is one-shot, so back/forward navigation cannot re-display a previous customer's result on a shared device.
- Errors are deliberately opaque. Not found, expired, disabled and no website membership all collapse to one message, so the endpoint cannot be used to enumerate codes.
- Rate limited per customer through `rateLimiterBy()`, 10 failed lookups per hour, checking up front and recording only on failure so a customer holding several cards is not penalised for genuine lookups.
- The POST carries an explicit form-key check, so a cross-site POST cannot burn the victim's rate-limit slots.

## API

`StoreRestrictionTrait` gains `assertAnyWebsiteAllowed()` and `assertAllWebsitesAllowed()`. Reads need an overlap with the token's scope; writes need every associated website to be in scope, so a restricted token cannot edit a card that also reaches a website it cannot see. Rewriting `websiteIds` is checked against both the incoming and the stored set, so a restricted token cannot claim a foreign card.

`PUT` accepts a read-modify-write round trip: echoing the stored `websiteIds` back is fine, and only an actual re-scope is rejected.

### Breaking changes

The API Platform layer is still beta, so these land without a deprecation cycle:

- REST and GraphQL responses expose `websiteIds` (a list) and no longer return the scalar `websiteId`.
- The GraphQL `createGiftcard` mutation replaces the `websiteId` argument with `websiteIds`. A client still sending `websiteId` fails input validation.
- REST create still accepts a scalar `websiteId` as a write-only legacy alias, honored only when `websiteIds` is omitted, so an old client is not silently scoped to the current website.

## Tests

New coverage for the junction lifecycle, the orphaned-card path after a website is deleted, the currency constraint across websites, the store-restriction gates, and the admin balance precision. Existing gift card tests were updated for the junction.

> [!NOTE]


fballiano:
Reviewed in depth (logic / security / conventions) and pushed fixes in e36a83d2c4. Two correctness bugs plus a CSRF gap were the notable ones:

**Correctness**
- `saveAction` was recording a **zero-amount** balance-adjustment history entry — `adjustBalance()` ran after the new balance was already set, so the delta computed as 0. Now keeps the model on the old balance through the first save so both the history delta and the final balance are correct.
- `Model/Resource/Giftcard::_afterLoad()` populated `website_ids` on every load, which defeated the `_afterSave` skip-guard: junction rows were deleted+reinserted on every save (checkout/refund hot paths), and an orphaned card with no associations threw "must be associated with at least one website" when `isValid()` auto-expired it — surfacing as a 500 in `BalanceController::checkAction`. Removed the population; `getWebsiteIds()` already lazy-loads on demand, so the guard now behaves as documented (null = untouched = skip, array = explicitly set = sync).

**Security**
- `BalanceController::checkAction` is a state-changing POST but didn't validate the form key (frontend doesn't do this automatically). A cross-site POST could burn a logged-in customer's rate-limit slots and lock them out for an hour. Added an explicit `_validateFormKey()`, matching `CartController`.
- Added `save` to `_setForcedFormKeyActions()` on the admin controller.

**Conventions / polish**
- `checkBalanceAction`: native `json_encode()` → `Mage::helper('core')->jsonEncode()`.
- Admin currency-hint JS targeted the stale `website_id` id; updated to the `website_ids` multiselect with `selectedOptions[0]`, so the currency hint updates again.
- `balance.phtml` SPDX corrected to `AFL-3.0` (templates are AFL, not OSL).

Verification: `composer lint` (phpstan/cs-fixer/rector) all clean, and the giftcard test suite passes (193/193). The multi-website model, junction sync (transactional, fail-closed on empty set), opaque-error enumeration defense, and per-customer rate limiting all hold up.

fballiano:
Thank you @mageaustralia!

Question: what if a giftcard is associated to 2 websites with 2 different currencies? since there's no currency directly tied to the giftcard it kinda worries me

fballiano:
I think we have 2 ways:
  1. Constrain the multiselect to websites that share the card's base currency. Keeps the whole
  stored-value model intact; "multi-website" just means "multiple same-currency storefronts" (the common head-office case the PR description actually cites). Minimal change, fully correct.
  2. Make currency explicit on the card (currency_code column, frozen at issue) and then convert at every redeem/use
  
I'm might be missing something here but with multiwebsite I don't think it's safe to leave it as it is.

what do you think?

mageaustralia:
@fballiano - You're always one step ahead.

I'd probably go with restrict associations to currency-matched websites . Validate in _beforeSave() that all website_ids share the same getBaseCurrencyCode() in the immediate term:

Longer term options:

 2. Add currency_code to the giftcard + convert on application. Card is denominated in one currency, redeemed elsewhere via Maho's currency rates. This would be more flexible, but rates are merchant-controlled and balance arithmetic across rate changes gets ugly (partial redemptions, refunds, history reconciliation).
 
 3. Per-website balance in the junction. Schema becomes (giftcard_id, website_id, balance). Each store tracks its own pot, conversion locked at first redemption. Most flexible, this could get complex - refund/reversal semantics are non-trivial
 
 What do you think?
 
**My particular use case** is that we have the online store and we have a separate store for POS - POS handles things a bit differently to live, mostly that it allows purchasing disabled / out of stock products. It also allows us to track In Store sales vs Online Sales easily but it also requires allowing the use of a GC purchased online to be used in-store.

fballiano:
@mageaustralia as a first implementation it's ok but I wonder how could we enforce it, like what if you edit a giftcard and change the websites completely to another currency. i mean it's an edge case but..

Also, would it be possible to keep the website_ids as CSV in the giftcard column? It's only OCD but the new table and the migration script with the leftover column (a real limitation of declarative schema) trigger me a bit. It's true that the separated table is a standard in our codebase..

mageaustralia:
@fballiano — sorry, I've been on holidays the last week, so slow to reply to this.

Two solid points, replying to both:

**On the currency edit-mid-life concern**

Rather than a `_beforeSave` guard, I think the cleaner fix is at the UI layer: only display websites in the multiselect whose base currency matches the card's currency. Ones that don't match are shown but greyed out with the currency printed next to the label — so the admin can see "Store B (EUR)" is unselectable and understand *why* immediately, without needing to know the internal rule.

That makes the mismatch impossible to save in the first place without an error message being needed, and covers the edit-mid-life case naturally (once the card has a currency, the picker only offers same-currency websites).

**On the schema — CSV column vs junction**

Happy to do whatever you like here. I was trying to stick with Maho conventions — every multi-website relationship in core (`catalog_product_website`, `salesrule_website`, `catalogrule_website`, `catalogrule_group_website`) uses a junction, so that's what I reached for. But if you'd prefer CSV on the `giftcard` row, I'll switch it — say the word and I'll rework this in the same commit as the currency picker fix.

fballiano:
Pushed a8a7abafa1: the `website_id` scalar is now fully migrated to the `giftcard_website` junction and **dropped** — no leftover column.

**How the migration works** (it turns out the declarative schema handles this fine): `sql/schema.php` simply stops declaring `website_id` + its index/FK. The declarative pass never drops columns (additive merge in the Canonicalizer — only undeclared indexes/FKs are removed), so on existing installs the data survives the schema pass, and the new `sql/giftcard_setup/upgrade-1.0.0-1.1.0.php` then backfills the junction and drops the column itself. Fresh installs skip the guarded block entirely; re-runs are no-ops; fresh and upgraded installs converge to identical schemas.

**Backfill is now conservative, not permissive.** The data-phase script associated every existing card with every active website — but pre-1.1.0 validation was a *strict* website match (`Total/Quote` → `isValidForWebsite()` → `website_id === $websiteId`), not store-agnostic. The permissive backfill would have made every existing card spendable on every website at face value, including cross-currency ones. The new backfill is one row per card (its original website), preserving behaviour exactly; operators broaden per card from the admin page — which also fits your POS use case (associate the online cards with the POS website once, deliberately).

**Latent bug fixed along the way**: `CartController` (×3), the quote observer, and core `Checkout/CartController` were still validating against the scalar, so a card associated to websites [2, 3] would be rejected at cart-apply on website 3 even though the junction said it's valid. All entry paths now do junction membership checks.

**Also in this commit**: same-currency guard on save (all associated websites must share one base currency — the server half of the picker rule we discussed; the greyed-out multiselect UI is still all yours if you want to add it on top), default-to-current-website for programmatic saves (restores the pre-1.1.0 default), `getWebsiteIds()` no longer marks lazy reads as pending changes (junction was being rewritten on every checkout balance save), `Collection::addWebsiteFilter()`, `setWebsiteId()` removed outright (new module, no BC baggage), and a new `WebsiteAssociationTest` that rebuilds the pre-1.1.0 schema and exercises the upgrade script end to end.

**Note for dev machines that already ran this branch**: your `core_resource` row for `giftcard_setup` is already at 1.1.0, so the new upgrade script won't fire. Either `./maho db:query "UPDATE core_resource SET version='1.0.0' WHERE code='giftcard_setup'"` and re-run `./maho migrate`, or drop the column manually.

Verification: `composer lint` clean (the one PHPStan error on the branch is an unrelated `Mage_Api2` ignore-entry leftover from the main merge), full Backend suite green (2242 passed), giftcard suite 200/200 including the new migration test.

fballiano:
Merged current main into the branch (4bfe2e8e3b) and adapted the giftcard code the merge brought in:

- **Model creation defaults**: main consolidated code/expiration/initial-balance defaults into `Giftcard::_beforeSave()` — merged the branch's junction default into it (`website_ids` → current website when nothing was set) and dropped the now-redundant blocks from the admin controller's saveAction.
- **New API Platform resource** (`Maho/Giftcard/Api/*` from #578) still targeted the dropped scalar: the GraphQL `create` mutation's `websiteId Int` arg wrote `website_id`, which would have been silently discarded at save. It's now `websiteIds [Int!]` feeding `setWebsiteIds()` (defaults to the current website when omitted), and the REST/GraphQL DTO exposes `websiteIds` — writable on create, populated from the junction on reads. The public `checkBalance` projection intentionally does not include website associations.

Verification after the merge: `composer lint` fully clean (including PHPStan — the Api2 ignore-entry issue is gone with current main), Backend suite 2332 passed, Api suite 565 passed (includes the GiftCard REST v2 read/write tests).

PR #1065, opened 2026-06-25
Fixed schema convergence re-emitting no-op charset CHANGEs (utf8/utf8mb3)

postadelmaga:
## Problem

On a legacy install, declarative schema reconciliation never converges: every
`maho migrate` re-applies a batch of `ALTER TABLE ... CHANGE` statements that
change nothing. On the database where this was found, ~23 such statements (≈70
columns) were re-emitted on every run, each altering only a column's
charset/collation to the value it already had (`utf8mb3` → `utf8mb3`). The deploy
is not blocked, but `migrate` is not idempotent and each run touches dozens of
columns for no reason.

The affected columns are all undeclared columns added to managed **core** tables
by third-party modules (legacy installs accumulate these). No indexes/FKs are
involved — it is purely column charset churn.

## Root cause

An asymmetry in how MySQL's DBAL `Comparator` strips column-level charset
(`Doctrine\DBAL\Platforms\MySQL\Comparator::normalizeTable` drops a column's
charset when it equals the *table's*, by a strict string compare).

- A legacy table is introspected as `utf8mb3`.
- The declarative target carries the legacy adapter's historical alias `utf8`
  (`Collector::applyTableDefaults`).

`utf8` and `utf8mb3` are the **same physical charset**, but the literal strings
differ. For a column merged onto the table by `preserveUndeclaredColumns`
(introspected as `utf8mb3`), the Comparator strips the charset on the **live**
side (`column utf8mb3 == table utf8mb3`) but keeps it on the **target** side
(`column utf8mb3 != table 'utf8'`), so it emits a `CHANGE` that can never
converge.

## Fix

`Canonicalizer::reconcile()` now aligns the target table's `charset`/`collation`
to the live value when the two are synonyms (`utf8` ≡ `utf8mb3`), making the
Comparator's strip symmetric on both sides. A genuine charset migration (e.g.
`utf8mb3` → `utf8mb4`) keeps differing strings and is still emitted.

## Validation

- New unit coverage in `CanonicalizerTest` for both the synonym case (no `CHANGE`,
  converges) and a real migration (`CHANGE` preserved). Full `CanonicalizerTest`
  green (11 passed).
- `composer lint:cs-fixer`, `composer lint:rector`, `composer lint:phpstan`
  (level 6) pass on the changed files.
- On a real legacy install (clone), `maho migrate` went from re-applying ~23
  statements on every run to "already up to date" on the first and second run —
  idempotent, converges to zero.

### DIFF
--- a/lib/Maho/Db/Schema/Canonicalizer.php
+++ b/lib/Maho/Db/Schema/Canonicalizer.php
@@ -57,6 +57,7 @@ final class Canonicalizer
     {
         self::stripPhantomIndexes($live, $physicalLiveIndexNames);
         self::alignIndexNames($live, $target);
+        self::alignTableCharset($live, $target);
         self::stripColumnComments($live);
         self::stripColumnComments($target);
         self::reconcileColumns($live, $target);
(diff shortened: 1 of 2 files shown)


fballiano:
Grazie millissime!!!!!!!!!!!!!

PR #1069, opened 2026-06-29
Fix: Caught Throwable in import validate/start so the admin loader always clears

claudiobianco:
The import form posts to a hidden iframe whose result frame calls
`postToFrameComplete()` to hide the "Please wait" loader. `validateAction()` and
`startAction()` only caught `Exception`, so a fatal `Error` (e.g. a `TypeError` from
`uploadSource()`) escaped, PHP rendered the 500 page into the iframe, the callback never
fired, and the loader spun forever with no feedback.

Widened both catches to `Throwable` so any error is shown in the result frame and the
loader clears.


fballiano:
Thanks @claudiobianco!

PR #1071, opened 2026-07-01
Fixed premature PayPal authorization release on partial invoices

claudiobianco:
# Fixed premature PayPal authorization release on partial invoices

## Problem

`Maho_Paypal_Model_Method_Abstract::capture()` set `final_capture: true` on
every capture where the invoiced amount differed from the order grand total:

```php
if ($amount != $payment->getOrder()->getBaseGrandTotal()) {
    $body = [ 'amount' => [...], 'final_capture' => true ];
}
```

`final_capture: true` tells PayPal to capture the amount **and release the
remaining authorized funds, closing the authorization**. That is wrong for the
first of several partial invoices: invoicing 40 of a 100 order releases the
remaining 60, so the later invoice for 60 fails against a closed authorization.

The heuristic conflated two different cases that both satisfy
`amount != grandTotal`:
- first of several partial invoices → must keep the authorization open
- reduced/last invoice → should release the remainder

## Fix

Derive finality from whether anything is left to invoice instead of from the
amount. `Invoice::register()` registers the invoice items (bumping
`qty_invoiced`) **before** calling `capture()`, so at capture time
`$order->canInvoice()` already reflects the post-invoice state:

```php
'final_capture' => !$order->canInvoice(),
```

- First partial invoice → items remain → `canInvoice()` true → `final_capture` false → authorization stays open.
- Last partial invoice → nothing left → `final_capture` true → authorization closes.
- Reduced final invoice (e.g. discount) → full items invoiced at a lower amount → `final_capture` true → remainder released as intended.

## Tests

`tests/Backend/Integration/Paypal/CaptureFinalizationTest.php` covers five
cases, asserting the exact body sent to `captureAuthorization()`:

- full single invoice → empty body (whole authorization captured)
- first of several partial invoices → `final_capture` false
- mid-sequence partial invoice → `final_capture` false
- last partial invoice → `final_capture` true
- reduced final invoice (discount) → `final_capture` true

The two "keep open" cases are the regression guards: they capture the same
amount as their finalizing counterparts and differ only in whether items are
left to invoice, so they fail against the old amount-based heuristic (which set
`final_capture` true for both) and pass with the fix.

`tests/Frontend/Integration/Paypal/SandboxCaptureRoundTripTest.php` (group
`sandbox`) proves the behaviour end-to-end against the real PayPal sandbox:
create a card order, authorize it, capture 40 of 100 with `final_capture` false,
assert the authorization is not closed, then capture the remaining 60 against
the still-open authorization. It skips when sandbox credentials are absent, or
when the sandbox account cannot produce a card authorization (ACDC disabled,
3DS required, card declined), so it never fails spuriously.

## Notes

The existing `amount != getBaseGrandTotal()` guard is kept, so the change is
limited to the `final_capture` decision. Making the capture amount diverge from
the order total on purpose (e.g. capturing a discounted amount against a
full-value authorization) is a separate feature and out of scope here.

> [!NOTE]


claudiobianco:
Disclaimer: haven't tested it against a proper paypal sandbox.
I'll try to add a proper sandbox test.

fballiano:
Thanks, solid fix. One related case was still broken though: invoices created with "Not Capture". Items register at invoice creation, not capture, so with two pending invoices `canInvoice()` is already false and capturing the first sent `final_capture: true`, closing the authorization the second still needed. (This was broken before this PR too, since the old code always sent true.)

Pushed a commit to this branch that covers it: `Payment::capture()` now exposes the current invoice via typed `setInvoice()`/`getInvoice()`, and the method only finalizes when nothing is left to invoice and no other open invoice awaits capture. It also sets `shouldCloseParentTransaction(true)` when finalizing so the local authorization transaction closes in step with the gateway, and adds `setApiClient()` so tests inject the fake client without reflection.

New tests cover the pending-invoice cases, including an end-to-end one that creates a real order with two "Not Capture" invoices and captures both through the full `Invoice::capture()` chain. Your original tests are unchanged and still pass.

fballiano:
@claudiobianco what do you think about this last version?

PR #1073, opened 2026-07-01
Preserved template directives when saving blog post content

claudiobianco:
`Maho_Blog_Model_Resource_Post::_beforeSave()` runs the malicious-code filter
  (HTMLPurifier) and `linkFilter()` over the post content. Both HTML-parse the
  markup, which breaks `{{media url="..."}}` / `{{widget ...}}` directives: their
  nested quotes inside an HTML attribute (e.g. an ``) aren't valid HTML,
  so the value is truncated at the inner quote and url-encoded to `%7B%7B...`. The
  image/widget then never renders, even though `getFilteredContent()` is meant to
  resolve the directive on output.

  Fix: mask `{{...}}` directives with placeholder tokens before filtering and
  restore them afterwards. Sanitization still runs; the directives survive.

  **Repro:** create a blog post whose content contains
  ``, save, view it — the image is
  missing (broken `%7B%7B` src) before this change, resolved after.
  
  Added a regression test in `PostBasicTest`.

> [!NOTE]

### DIFF
--- a/app/code/core/Maho/Blog/Model/Resource/Post.php
+++ b/app/code/core/Maho/Blog/Model/Resource/Post.php
@@ -172,12 +172,28 @@ class Maho_Blog_Model_Resource_Post extends Mage_Eav_Model_Entity_Abstract
             $object->setData('publish_date', Mage::app()->getLocale()->utcToStore()->format(Mage_Core_Model_Locale::DATE_FORMAT));
         }
 
-        // Filter HTML content
+        // Filter HTML content. Mask {{...}} template directives first: the malicious-code filter
+        // (HTMLPurifier) and linkFilter() HTML-parse the content, which would break directives
+        // whose nested quotes (e.g. {{media url="..."}} inside an img src) are not valid HTML
+        // attribute syntax. The directives are restored after filtering and resolved by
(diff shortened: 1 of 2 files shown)


fballiano:
Thanks @claudiobianco — good catch, and the diagnosis is right: `_beforeSave()` sanitizes the raw content before the directives are resolved, and `{{media url="..."}}` inside an attribute isn't valid HTML, so HTMLPurifier mangles it.

I pushed a commit that keeps your approach but tightens the masking. The original masked every `{{...}}` span, including `{{<script>…</script>}}`, which let arbitrary markup skip the filter. It now masks only real directives (`{{keyword …}}`) before filtering and restores them after, so `{{media}}`/`{{widget}}` survive while anything else in braces still goes through the malicious-code filter and gets stripped. Sanitization stays on save, so the stored value is never dangerous (the existing `BlogPostSecurityTest` cases keep passing); `getFilteredContent()` just resolves the preserved directives on output.

Extended the test to cover directive resolution on output and sanitization of markup wrapped in a directive. Opened #1076 to unify this resolve/sanitize handling across the rest of the backend (CMS page/block/widget, product descriptions) and to track directive-allowlist hardening separately.

PR #1074, opened 2026-07-01
Preserved template directives in newsletter and email content previews

claudiobianco:
`Mage_Adminhtml_Block_Template::maliciousCodeFilter()` runs the malicious-code
  filter (HTMLPurifier) over the raw template text *before* it is processed. That
  HTML-parses the content and breaks `{{media url="..."}}` / `{{widget ...}}`
  directives whose nested quotes inside an HTML attribute aren't valid HTML, so the
  newsletter template and queue previews show a broken image instead of the
  resolved one. The editor itself is unaffected (it renders images through a
  different endpoint), so the preview and the editor disagree.

  Fix: mask `{{...}}` directives with placeholder tokens before filtering and
  restore them afterwards; the preview then resolves them via
  `getProcessedTemplate()`. Being a shared block method, this fixes both the
  newsletter template preview and the queue preview.
 
  **Repro:** insert an image via the media browser into a newsletter template, open
  the preview — the image is missing before this change, shown after.

> [!NOTE]

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Newsletter/Queue/Preview.php
+++ b/app/code/core/Mage/Adminhtml/Block/Newsletter/Queue/Preview.php
@@ -27,13 +27,6 @@ class Mage_Adminhtml_Block_Newsletter_Queue_Preview extends Mage_Adminhtml_Block
             $template->setTemplateText($this->getRequest()->getParam('text'));
             $template->setTemplateStyles($this->getRequest()->getParam('styles'));
         }
-        $template->setTemplateStyles(
-            $this->maliciousCodeFilter($template->getTemplateStyles()),
-        );
-        $template->setTemplateText(
-            $this->maliciousCodeFilter($template->getTemplateText()),
-        );
-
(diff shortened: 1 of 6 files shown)


fballiano:
Thanks @claudiobianco — spot-on diagnosis: the previews sanitize the raw template before `getProcessedTemplate()` resolves the directives, so a `{{media url="..."}}` inside an attribute gets mangled by HTMLPurifier.

I pushed a commit that changes the approach: instead of masking the directives out of the filter (which restored them verbatim and let markup inside braces bypass sanitization), the previews now run the malicious-code filter on the **resolved** output of `getProcessedTemplate()`, where the directives are already real URLs/values. `maliciousCodeFilter()` reverts to the plain filter. Since a preview has no persistence step, sanitizing the rendered output is the natural single place to do it. That fixes the broken image and keeps the sanitizer intact.

Added a unit test for the filter plus an integration test for the newsletter preview covering the resolved media directive and the sanitized-output case. Opened #1076 to unify this across the backend.

PR #1088, opened 2026-07-07
Fixed FeedManager 'Use Parent' fallback for rule, CSV and category sources and added sale_price formatting

mageaustralia:
Found while building a Google Shopping feed for a store with configurable products. Four related bugs, all in `Maho_FeedManager_Model_Mapper`:

### 1. "Use Parent" is ignored when the source is a Dynamic Rule

If a feed field uses a Dynamic Rule as its source (e.g. the built-in `sale_price` rule) and "Use Parent" is set, the option does nothing - only attribute and taxonomy sources honour it. Typical result: on configurable products where the special price is set on the parent, every variant child outputs an empty `<g:sale_price>`.

Fix: when the rule returns nothing for the child (or the mode is "always"), the rule is re-evaluated against the **parent product model**. It has to be the loaded parent model rather than the parent's raw data row, because the row's `price` key already holds the discounted final price - a rule condition like `special_price < price` can never match against it.

### 2. "Use Parent" from the CSV builder never matches

The CSV column builder stores use_parent as a boolean. By the time it reaches the source resolver it has become the string `"1"`, which is neither `"if_empty"` nor `"always"`, so it is silently ignored. `"1"` is now accepted as `"if_empty"`.

### 3. `sale_price` missing from `PRICE_FIELDS`

Fields sourced from anything named `sale_price` skip the feed's price formatting (decimal places, currency suffix) and output raw DB values like `249.9500` instead of `249.95 AUD`. Added to the constant.

### 4. Parent fallback has no category data

`_extractParentProductData()` copies the parent's attributes but never computes the derived category keys (`category_ids`, `category_names`, `category_path`). Variant children are usually not assigned to categories themselves, so "Use Parent" on category-based fields (e.g. `g:product_type` sourced from `category_path`) outputs nothing. On our catalog more than half the items ended up with no `product_type`. The parent extraction now computes the same three keys child rows get.

### Caching
Parent models are cached per parent id (same pattern as the existing `_parentDataCache`), so sibling variants do not trigger repeated product loads.

### DIFF
--- a/app/code/core/Maho/FeedManager/Model/Mapper.php
+++ b/app/code/core/Maho/FeedManager/Model/Mapper.php
@@ -39,6 +39,7 @@ class Maho_FeedManager_Model_Mapper
         'max_price',
         'tier_price',
         'group_price',
+        'sale_price',
         'msrp',
         'cost',
     ];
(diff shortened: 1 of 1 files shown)


fballiano:
Thank you @mageaustralia!

PR #1093, opened 2026-07-08
Made FeedManager's "Use Parent: always" strict for Dynamic Rule sources and added a currency_convert transformer

mageaustralia:
Two follow-ups to #1088, found while continuing the same feed migration:

### 1. "Use Parent: always" should be strict for Dynamic Rule sources

The version merged in #1088 fell back to the child value when the parent produced nothing. That is wrong for configurable products: the storefront prices configurables from the **parent**, so a special price set only on a child variant is never what the customer actually pays. With the fallback, a child-only special produced a `<g:sale_price>` for a sale that does not exist on the site - Google flags these as price mismatches. With strict semantics the parent result is the answer, even when empty (products with no parent still use their own values).

### 2. New `currency_convert` transformer

FeedManager had no way to produce a converted-currency feed (e.g. an NZD Google feed from an AUD-base store) - `format_price` formats but never converts. The new transformer converts numeric values via the directory currency rates. Options: `to` (required, e.g. `NZD`), `from` (optional, defaults to the store base currency). Chain it before `format_price`:

```
currency_convert:to=NZD|format_price:currency=NZD,decimals=2
```

Verified on the same production-scale catalog as #1088: child-only specials no longer leak into sale_price (matches storefront pricing exactly), and the NZD feed produces converted values identical to the previous platform (rate-applied, 2dp).

### DIFF
--- a/app/code/core/Maho/FeedManager/Model/Mapper.php
+++ b/app/code/core/Maho/FeedManager/Model/Mapper.php
@@ -476,10 +476,11 @@ class Maho_FeedManager_Model_Mapper
         if ($useParentMode === 'always') {
             $parent = $this->_getParentProductModel($product);
             if ($parent !== null) {
-                $value = $this->_evaluateRule($ruleCode, $rawData, $parent);
-                if ($value !== null && $value !== '') {
-                    return $value;
-                }
+                // Strict parent semantics: when a parent exists its result IS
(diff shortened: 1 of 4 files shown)


fballiano:
amazing! thank you @mageaustralia!

PR #1104, opened 2026-07-16
Fixed TypeError in `EavConfig::getAttributes()` when called before store initialization

hirale:
### Problem

Any request whose *first* EAV access goes through `Mage_Eav_Model_Config::getAttributes()` fails with

```
TypeError: array_keys(): Argument #1 ($array) must be of type array, null given
```

and the whole request 500s.

This surfaces easily with the new API Platform layer: `rest.php` boots Maho in library mode, so nothing has touched EAV before the resource provider runs. For example, loading a wishlist item collection pulls the product listing attributes through `getAttributes()` as the very first EAV access of the request, and the endpoint returns a 500.

### Environment

- Maho: `mahocommerce/maho` `dev-main` (`5e9c0512a`)
- PHP 8.5.7 
- MySQL 8.4.9

### Steps to reproduce (vanilla `main`)

1. Enable the REST API (all protocols are off by default): System > Configuration > Services > API → enable REST v2.
2. Use any customer account that has **at least one wishlist item** (add one in the storefront if needed — the collection must hydrate products to reach the attribute index).
3. Request the wishlist over the API so `getAttributes()` is the first EAV access of the request:

```bash
BASE=https://your-maho.test

TOKEN=$(curl -s -X POST "$BASE/api/rest/v2/auth/token" \
  -H "Content-Type: application/json" \
  -d '{"grant_type":"customer","email":"customer@example.com","password":"..."}' \
  | jq -r .token)

curl -s -i "$BASE/api/rest/v2/customers/me/wishlist" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Accept: application/json"
```

**Actual:** `HTTP/1.1 500` with `{"error":"internal_server_error","message":"An internal error occurred","code":500}`; on a developer-mode install the debug block shows the underlying `TypeError` from `Mage_Eav_Model_Config::getAttributes()`.

**Expected (with this fix):** `HTTP/1.1 200` and the wishlist items.

### Root cause

`getAttributes()` reads `$this->_entityTypeAttributes[$storeId][$entityTypeId]` and calls `array_keys()` on it directly, but never ensures the per-store attribute index has been built for `$storeId`. Its sibling `getAttribute()` calls `$this->_initializeStore($storeId)` before touching the same index; `getAttributes()` is missing that guard, so the index is still `null` for any store not yet initialized in the current request.

On regular frontend requests this is masked because earlier code has almost always initialized the store by the time `getAttributes()` runs. In a fresh library-mode context (an API Platform request) nothing has.

### DIFF
--- a/app/code/core/Mage/Eav/Model/Config.php
+++ b/app/code/core/Mage/Eav/Model/Config.php
@@ -521,8 +521,12 @@ class Mage_Eav_Model_Config
         $entityType = $this->getEntityType($entityType);
         $attributes = [];
         $storeId = $this->_storeId();
+        // Ensure the per-store attribute index is loaded (getAttribute() does the same). Without
+        // this, a caller in a context that has not yet initialized the store — e.g. an API request
+        // whose first EAV access is getAttributes() — hits array_keys(null) below.
+        $this->_initializeStore($storeId);
         // need to access attributes to ensure they are hydrated and initialized
-        foreach (array_keys($this->_entityTypeAttributes[$storeId][$entityType->getId()]) as $attributeId) {
+        foreach (array_keys($this->_entityTypeAttributes[$storeId][$entityType->getId()] ?? []) as $attributeId) {
             $attributes[] = $this->getAttribute($entityType, $attributeId, $storeId);
         }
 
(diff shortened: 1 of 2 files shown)


fballiano:
Thank you @hirale! 🙏

PR #1105, opened 2026-07-16
Fixed multiselect attribute filters in the products API returning no results

hirale:
### Problem

Filtering the products collection by a multiselect attribute silently returns an empty result, even when matching products exist:

```
GET /api/rest/v2/products?attr_material=47   →  0 items
```

No error is raised — the filter just never matches. The same happens through GraphQL, since the `attributeFilters` JSON goes through the same code path.

### Environment

- Maho: `mahocommerce/maho` `dev-main` (`5e9c0512a`) 
- PHP 8.5.7
- MySQL 8.4.9

### Steps to reproduce (vanilla `main`)

1. Enable the REST API: System > Configuration > Services > API → enable REST v2.
2. Create a **multiselect** product attribute (e.g. `material`) with a few options, add it to the attribute set, and select one or more options on an enabled, visible product. Reindex (System > Index Management).
3. Filter the (public) products endpoint by one of the selected option ids:

```bash
BASE=https://your-maho.test

curl -s "$BASE/api/rest/v2/products?attr_material=<optionId>" \
  -H "Accept: application/json"
```

**Actual:** an empty collection — the product with that option selected is not returned.

**Expected (with this fix):** the matching products.

### Root cause

A multiselect attribute persists its selected option ids as one comma-joined string (e.g. `"47,51,46"`). The attribute-filter loop in `ProductProvider` applies a plain equality match for every attribute:

```php
$collection->addAttributeToFilter($code, $value);
```

`= '47'` can never match `"47,51,46"`, so any multiselect filter yields zero rows. Set membership needs `FIND_IN_SET`, which the collection API already exposes as the `finset` condition type — it is what layered navigation uses to count facet options.

### API contract

- `?attr_<code>=47,51` — comma-separated option ids mean **any-of** (OR), consistent with how layered-navigation facets count products (a facet count answers "how many products have this option", not "only this option").
- An array value is accepted through the same path — GraphQL's `attributeFilters` JSON naturally delivers arrays.
- Non-multiselect attributes are untouched: exact same equality behavior as before.

> [!NOTE]

### DIFF
--- a/app/code/core/Mage/Catalog/Api/ProductProvider.php
+++ b/app/code/core/Mage/Catalog/Api/ProductProvider.php
@@ -350,9 +350,28 @@ final class ProductProvider extends \Maho\ApiPlatform\Provider
             }
         }
         foreach ($attributeFilters as $code => $value) {
-            if (preg_match('/^[a-z][a-z0-9_]*$/', $code)) {
-                $collection->addAttributeToFilter($code, $value);
+            if (!preg_match('/^[a-z][a-z0-9_]*$/', $code)) {
+                continue;
             }
+            // Multiselect attributes persist their selected option ids as a comma-joined set,
+            // so an equality match never hits (the bug that made e.g. ?attr_material= return
+            // zero). Filter each id with FIND_IN_SET (finset) instead, OR-ing multiple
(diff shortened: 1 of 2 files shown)


fballiano:
Thank you so much @hirale!!

PR #1107, opened 2026-07-16
Align wishlist item pricing with the products API (price + finalPrice)

hirale:
### Problem

A wishlist is one of the highest purchase-intent surfaces in a storefront, and the standard merchandising pattern there is "was $49.99 — now $39.99" with a sale badge: showing a customer that something they already want has dropped in price is one of the cheapest conversion wins available.

The wishlist API currently can't support it — and its one price field is also inconsistent with the products API:

| | products API | wishlist item |
|---|---|---|
| base price | `price` | — |
| effective price | `finalPrice` | `productPrice` |

`WishlistItem.productPrice` carries the *final* price while the products API uses `price` for the *base* price, and there is no second field, so a client can neither detect a discount nor render the regular-vs-sale pair.

### Proposed change

Adopt the products API naming on the wishlist item (with its usual `product` prefix):

- `productPrice` — now the **base** price, mirroring `price`; `null` when the product type carries no own price (e.g. configurable parents priced via their children).
- `productFinalPrice` — **new**, computed, read-only: the effective price after catalog rules and specials, mirroring `finalPrice`.

Clients display `productFinalPrice` and derive "on sale" from `productPrice > productFinalPrice`; no extra boolean is needed. Both fields appear on REST and GraphQL alike. The two wishlist API tests are extended to cover the new field.

```json
// GET /api/rest/v2/customers/me/wishlist  (item, after)
{
  "productName": "16GB Memory Card",
  "productPrice": 49.99,
  "productFinalPrice": 39.99,
  ...
}
```

> [!NOTE]

### DIFF
--- a/app/code/core/Mage/Wishlist/Api/WishlistItem.php
+++ b/app/code/core/Mage/Wishlist/Api/WishlistItem.php
@@ -154,11 +154,15 @@ class WishlistItem extends CrudResource
 
         $dto->productName = $product->getName();
         $dto->productSku = $product->getSku();
-        // Mirrors the products API: productPrice = base price (null when the type has no own
-        // price, e.g. configurable parents), productFinalPrice = effective price after rules.
-        $basePrice = (float) $product->getPrice();
-        $dto->productPrice = $basePrice > 0.0 ? $basePrice : null;
-        $dto->productFinalPrice = (float) $product->getFinalPrice();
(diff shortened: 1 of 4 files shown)


fballiano:
@hirale I make a couple of changes, are you ok with those?

hirale:
> @hirale I make a couple of changes, are you ok with those?

LGTM. Thanks!

PR #1110, opened 2026-07-18
Fixed PayPal webhook signature verification for payloads containing empty objects

hirale:
`verifyWebhookSignature()` decodes the delivered webhook body and lets the HTTP client re-encode it into the `verify-webhook-signature` request. PHP's associative decode collapses empty JSON objects to empty arrays, so `{}` is re-encoded as `[]`. PayPal's transmission signature covers the exact bytes it delivered — one folded `{}` and verification returns `FAILURE` deterministically.

Minimal repro of the mutation:

```php
json_encode(json_decode('{"breakdown":{}}', true)); // '{"breakdown":[]}'
```

In practice this means every `CHECKOUT.ORDER.APPROVED` event fails verification (its `amount.breakdown` is an empty object), so the `Webhook_Handler_OrderApproved` browser-died failsafe can never run — the webhook endpoint 401s the delivery and all retries. `PAYMENT.CAPTURE.COMPLETED` happens to contain no empty object and passes, which disguises the bug as event-type-specific flakiness. Verified against the live sandbox by replaying a captured delivery with identical transmission headers: raw body spliced verbatim → `SUCCESS`; decode→re-encode → `FAILURE`.

Reproducible with curl against any real delivery you capture at your webhook endpoint — fill the five values from the delivery's `PAYPAL-*` headers and paste the raw request body as `webhook_event`:

```bash
# 1. webhook_event = the delivered bytes, verbatim (note "breakdown":{})  →  SUCCESS
curl -s https://api-m.sandbox.paypal.com/v1/notifications/verify-webhook-signature \
  -u "$CLIENT_ID:$CLIENT_SECRET" \
  -H 'Content-Type: application/json' \
  -d '{
    "auth_algo":        "<PAYPAL-AUTH-ALGO>",
    "cert_url":         "<PAYPAL-CERT-URL>",
    "transmission_id":  "<PAYPAL-TRANSMISSION-ID>",
    "transmission_sig": "<PAYPAL-TRANSMISSION-SIG>",
    "transmission_time":"<PAYPAL-TRANSMISSION-TIME>",
    "webhook_id":       "<registered webhook id>",
    "webhook_event":    {"id":"WH-…","event_type":"CHECKOUT.ORDER.APPROVED", …, "amount":{"value":"13.64","breakdown":{}}, …}
  }'
# → {"verification_status":"SUCCESS"}

# 2. identical call, but with the one-byte-pair PHP round-trip artifact:
#    "breakdown":{}  →  "breakdown":[]   (what the client currently sends)
# → {"verification_status":"FAILURE"}
```

The two requests differ only in those two bytes; PayPal's verify API accepts the first and rejects the second, which pins the failure to the body mutation rather than the headers, credentials, or webhook configuration.

### DIFF
--- a/app/code/core/Maho/Paypal/Model/Api/Client.php
+++ b/app/code/core/Maho/Paypal/Model/Api/Client.php
@@ -269,22 +269,18 @@ class Maho_Paypal_Model_Api_Client
     {
         $certUrl = $headers['PAYPAL-CERT-URL'] ?? '';
         if (!preg_match('#^https://api(-m)?\.(?:sandbox\.)?paypal\.com/#', $certUrl)) {
+            $this->_log('Webhook verification rejected', ['reason' => 'cert_url', 'cert_url' => $certUrl]);
             return false;
         }
 
-        $verifyBody = [
-            'auth_algo' => $headers['PAYPAL-AUTH-ALGO'] ?? '',
-            'cert_url' => $certUrl,
-            'transmission_id' => $headers['PAYPAL-TRANSMISSION-ID'] ?? '',
-            'transmission_sig' => $headers['PAYPAL-TRANSMISSION-SIG'] ?? '',
-            'transmission_time' => $headers['PAYPAL-TRANSMISSION-TIME'] ?? '',
-            'webhook_id' => $webhookId,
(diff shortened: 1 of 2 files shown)


fballiano:
Thank you @hirale 🙏🙏🙏

PR #1134, opened 2026-07-23
Added expression support to `{{if}}`, `{{depend}}` and `{{var}}` template directives and deprecated the legacy tokenizer

fballiano:
## Summary

Template directives are now evaluated through Symfony ExpressionLanguage (already a direct dependency, previously only exercised inside the API Platform kernel). `{{if}}` and `{{depend}}` gain comparisons, boolean logic and `in` checks, and `{{var}}` resolves through the *same* engine — so the hand-written `Variable` tokenizer is deleted and both directives share one evaluator and one security model.

```
{{if order.grand_total > 100 && customer.group_id == 2}}
  Thanks for the big wholesale order!
{{else}}
  Thanks for your order!
{{/if}}

{{if product.getAttributeText('color') == 'Red'}}...{{/if}}
{{if order.status in ['complete', 'closed']}}...{{/if}}
```

The legacy single-variable syntax (`order.increment_id`, `customer.getName()`) is a strict subset of the expression syntax, so every directive shipped in core templates parses unchanged — verified against all `{{if}}`/`{{depend}}`/`{{var}}` usages in `app/locale` and `app/design`.

## What changed

- **Conditions** (`{{if}}` / `{{depend}}`) are evaluated as expressions instead of a single-variable emptiness check.
- **`{{var}}`** (and the `$value` parameters of `{{include}}`/`{{template}}`) resolve through the same evaluator. The tokenizer-based `_getVariable()` internals are gone; the `Variable` tokenizer itself is no longer used by Maho but is **kept and marked `@deprecated`** (along with its `Varien_Filter_Template_Tokenizer_Variable` alias) in case third-party code instantiates it directly. The `Parameter` tokenizer that parses `key="value"` attribute lists is unrelated and stays.
- **Object access is sandboxed.** Symfony documents ExpressionLanguage as explicitly *not* a sandbox, so template variables are exposed through a wrapper that gates every property read and method call.

## Security model

Editing templates is a privileged admin capability, but conditions evaluate against live models and the comparison operators would otherwise turn any readable value into a character-by-character extraction oracle. The wrapper therefore enforces:

- **Read-only access.** Only `get`/`has`/`is`/`can`/`to`/`count`/`format` accessors are callable; a state-changing method (`save()`, `delete()`, `cancel()`, a setter, or an `is*` predicate handed an argument like `isDeleted($flag)`) is refused. This closes a long-standing hole: `{{var order.delete()}}` **executed on render**, inherited unchanged from the Varien/Magento filter where `method_exists()` was a free pass.
- **No secrets.** Fields named like credentials (`password`, `secret`, `token`, `api_key`, `private_key`, `cc_number`, `cc_cid`, or the `_enc` encrypted-column suffix) are refused as a getter, a property, or a `getData()` key. Boolean flags such as `customer.is_change_password` stay readable.
- **No infrastructure traversal.** Only `DataObject` values are walked into, so `order.getResource().getReadConnection()...` cannot reach database credentials.
- **No whole-object dumps or path walking** — `toArray()`/`toJson()`/`toXml()`, an unkeyed `getData()`, and `getDataByPath()` are refused.
- **No arbitrary constants or DoS operators** — `constant()`/`enum()` are disabled, and the `matches` (regex) and `..` (range) operators are rejected at the AST level.
- **Encrypted `getConfig()` paths are neutralized**, preserving the existing `Mage_Adminhtml_Model_Email_PathValidator` guard.

## Backward compatibility

- **Property resolution matches the legacy resolver**: `order.status_label` prefers a real `getStatusLabel()` over raw `getData()`.
- **Absent variables evaluate silently as empty** — `{{depend comment}}` / `{{var comment}}` with no such variable renders nothing.
- **A broken directive never breaks the template**: a parse/evaluate failure logs a warning and renders the else branch (or empty), instead of aborting.

Deliberate behaviour changes:

- `0` / `'0'` now evaluate as **false** in conditions (standard PHP truthiness). The old `== ''` comparison treated them as true — a quirk rather than a feature.
- `{{var}}` can no longer invoke state-changing methods (see above).
- Method access is narrowed to read-only accessors. Every method call in shipped templates is a `get*`/`format` accessor, so nothing in core regresses; custom templates calling other methods would need adjusting.
- Read getters that take an argument now work in **both** directives — `getAttributeText('color')`, `format('html')`, `getData('sku')`.
- `Maho\Filter\Template\Tokenizer\Variable` (and its `Varien_Filter_Template_Tokenizer_Variable` alias) is **deprecated, not removed** — Maho no longer uses it, but it still works for third-party callers and will be dropped in a future release.

## Testing

- 47 Pest tests across `TemplateConditionTest` and `TemplateVarMethodGateTest`, covering legacy syntax, expressions, nested objects, error degradation, and a regression test for every escape found during review (DB-credential chain, card data, `getPassword()`, `getDataUsingMethod()`, `getDataByPath()`, whole-object dumps, `constant()`/`enum()`, `matches`/`..`, and state-changing calls)
- Email and newsletter suites pass (85 tests), Frontend suite passes (160 tests)
- php-cs-fixer, rector and PHPStan (level 6) clean

> [!NOTE]

### DIFF
--- a/.phpstan.dist.baseline.neon
+++ b/.phpstan.dist.baseline.neon
@@ -24379,24 +24379,6 @@ parameters:
 			path: lib/Maho/Filter/Template/Tokenizer/Variable.php
 
 		-
-			rawMessage: '''
-				Access to constant on deprecated class Maho\Filter\Template\Tokenizer\Variable:
-				since 26.9 Use Symfony\Component\ExpressionLanguage\ExpressionLanguage instead, as Maho\Filter\Template now does to resolve {{var}} and {{if}}/{{depend}}.
-			'''
-			identifier: classConstant.deprecatedClass
-			count: 1
-			path: app/bootstrap.php
-
-		-
-			rawMessage: '''
-				Access to constant on deprecated class Maho\Filter\Template\Tokenizer\Variable:
-				since 26.9 Use Symfony\Component\ExpressionLanguage\ExpressionLanguage instead, as Maho\Filter\Template now does to resolve {{var}} and {{if}}/{{depend}}.
-			'''
-			identifier: classConstant.deprecatedClass
-			count: 1
-			path: lib/Maho/Rector/VarienToMahoClassMap.php
-
-		-
 			rawMessage: 'Method Maho\Io\File::dirname() has no return type specified.'
 			identifier: missingType.return
 			count: 1
--- a/.phpstan.dist.baseline.neon
+++ b/.phpstan.dist.baseline.neon
@@ -24379,6 +24379,24 @@ parameters:
 			path: lib/Maho/Filter/Template/Tokenizer/Variable.php
 
 		-
+			rawMessage: '''
+				Access to constant on deprecated class Maho\Filter\Template\Tokenizer\Variable:
+				since 26.9 Use Symfony\Component\ExpressionLanguage\ExpressionLanguage instead, as Maho\Filter\Template now does to resolve {{var}} and {{if}}/{{depend}}.
+			'''
+			identifier: classConstant.deprecatedClass
+			count: 1
+			path: app/bootstrap.php
+
+		-
+			rawMessage: '''
+				Access to constant on deprecated class Maho\Filter\Template\Tokenizer\Variable:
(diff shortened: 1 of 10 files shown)


fballiano:
This is hyper complicated and to me it doesn't feel right. closing.

PR #1142, opened 2026-07-25
Added a guard test for admin form values cleared by setValues()

fballiano:
Stacked on #1141 — that PR's commit is included here until it merges.

`Maho\Data\Form::setValues()` clears every element the data array doesn't mention, so a field declared as

```php
$fieldset->addField('website_ids', 'hidden', ['value' => $websiteId]);
```

loses that value as soon as the block runs its usual `$form->setValues($model->getData())` and the model has no such key — the case for every *new* record. For a hidden field or a checkbox nothing on screen hints at it. Three admin forms shipped this, each silent in its own way:

- **Customer segments** (#1141): new segments in single-store mode could not be saved at all — "Please select at least one website", with no website field on the form.
- **API Platform user**: the "Regenerate Credentials" checkbox rendered `value=""`, so ticking it posted an empty string and `!empty($data['regenerate_client_credentials'])` never fired. Regeneration silently did nothing.
- **Blog post / category**: the single-store branch seeded `setStoreId()`, a key no element uses, so the hidden `stores` field posted empty and the record was saved against store 0 instead of the store view. A loaded record renders empty too, since an array value yields an empty hidden input.

### The guard

`tests/Backend/Unit/Adminhtml/FormDeclaredValueTest.php` tokenizes the admin blocks and flags `addField()` calls that declare a value in a form that later calls `setValues()`, with the fix spelled out in the failure message (seed the same key on the model, as `Mage_Adminhtml_Block_Cms_Page_Edit_Tab_Main` does).

It is deliberately narrow, so it stays actionable rather than chatty:

- only `hidden` and `checkbox` fields, where a lost value is invisible and the admin cannot re-enter it;
- only values that can't come back on their own — an already-empty value is a no-op, and a value read off the object whose `getData()` feeds `setValues()` returns with it;
- anything the block seeds back, re-applies after `setValues()`, or that is allowlisted.

Three allowlist entries remain, each with the reason it is safe at runtime. A second test exercises the analyzer itself on the shape it exists to catch.

`composer lint` clean; Backend suite passes (985 tests).

### DIFF
--- a/app/code/core/Maho/ApiPlatform/Block/Adminhtml/Apiplatform/User/Edit/Tab/Main.php
+++ b/app/code/core/Maho/ApiPlatform/Block/Adminhtml/Apiplatform/User/Edit/Tab/Main.php
@@ -181,6 +181,9 @@ class Maho_ApiPlatform_Block_Adminhtml_Apiplatform_User_Edit_Tab_Main extends Ma
         // shows a "stored hashed" note instead of the value.
         $data = $model->getData();
         unset($data['api_key']);
+        // The checkbox is not a column of the user, so setValues() would clear
+        // its declared value and the box would post an empty string when ticked.
+        $data['regenerate_client_credentials'] = 1;
         $form->setValues($data);
         $this->setForm($form);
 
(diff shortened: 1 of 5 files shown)


fballiano:
Folded into #1141 — everything lives in one PR there.

PR #1158, opened 2026-07-26
Review follow-ups for the stuck-cron reaper: clearer admin comment and cron:unlock tests

fballiano:
Follow-up refinements from reviewing #1156, targeted at that PR's branch so they can fold into it.

## Changes

- **Clarified the `max_running_time` admin field comment** (`etc/system.xml` and its `Mage_Cron.csv` translation). The reaper flips *any* `running` row past the threshold to `error`, including a job that is legitimately still executing in a concurrent worker. The comment now tells the admin to keep the value above the runtime of the slowest legitimate job.
- **Hardened `StuckJobReaperTest`.** The "reaper is disabled" case overrides `max_running_time` to `0`; the restore now lives in `afterEach` so a failed assertion can't leak the override into sibling tests.
- **Added `CronUnlockTest`.** The `cron:unlock` command previously had no coverage. New tests exercise the default-threshold path (only rows older than `max_running_time`), `--all` (every running row regardless of age), `job_code` scoping, and the empty-result message.

## Notes

- No production logic changed — this is a documentation-string clarification plus test coverage.
- I also examined whether a stale reaper message could survive on a row that later succeeds. It cannot: `cron_schedule.messages` is nullable and the executing schedule is loaded with `messages = null`, so the success `save()` writes `NULL` via `_prepareDataForTable`, overwriting any concurrent reaper message. No code change was needed there.
- Lint and the Pest suite could not be run in this sandbox (composer cannot authenticate to github.com through the egress proxy, so `vendor/` is unavailable); CI will validate both.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Ee4XnSH26VkdPCYkTNX2JG

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ee4XnSH26VkdPCYkTNX2JG)_

### DIFF
--- a/app/code/core/Mage/Cron/etc/system.xml
+++ b/app/code/core/Mage/Cron/etc/system.xml
@@ -47,7 +47,7 @@ SPDX-License-Identifier: AFL-3.0
                             <show_in_website>1</show_in_website>
                             <show_in_store>1</show_in_store>
                             <validate>required-entry validate-digits validate-zero-or-greater</validate>
-                            <comment>Stale jobs whose worker died (OOM, timeout, SIGKILL) are flipped from "running" to "error" after this many minutes. Set to 0 to disable.</comment>
+                            <comment>Stale jobs whose worker died (OOM, timeout, SIGKILL) are flipped from "running" to "error" after this many minutes. Keep this above the runtime of your slowest job, as any job still running past this threshold is marked as an error. Set to 0 to disable.</comment>
                         </max_running_time>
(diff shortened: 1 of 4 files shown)


fballiano:
Closing — these commits were pushed directly onto #1156's branch (`claude/issue-1154-fix-m63g6n`) instead, so this separate PR is no longer needed.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #1167, opened 2026-07-27
Deleting an API role triggers 405 Method Not Allowed


empiricompany:
When trying to delete an API role from the admin edit form (System → API → Roles), clicking the delete link results in a 405 Method Not Allowed error.

Status: 405
Method: GET
URL pattern: /gestione/apiplatform_role/delete/role_id/{id}/form_key/{key}/key/{aclKey}/
Maho Version: dev-main dfe006c

PR #1168, opened 2026-07-27
Fixed empty API role permission tree and 405 when deleting an API role

### LINKED ISSUE #1166: API Roles permission tree not loading child nodes – tree renders empty with only root node
When creating a new API role under System → API → Roles, the permission tree renders only the root node (__root__) with no child permissions. The tree appears completely empty, making it impossible to grant granular ("Custom") resource access to API users.

Image

Latest dev-main dfe006c
No JavaScript console errors.

```html
<!-- Section 3: Service Account Permissions (configurable) -->
<div class="api-access-section">
    <div class="section-head section-head-service">
        Service Account Permissions    </div>
    <div class="section-body">
        <p style="margin:0 0 8px; font-size:12px; color:#555;">
            These permissions control what API users (service accounts) can do. Grant only the operations each integration needs.        </p>
        <input type="hidden" name="permissions_tree" id="role_permissions" value="">
        <div class="api-service-controls">
            <label for="permissions_all">Resource Access</label>
            <select id="permissions_all" name="permissions_access" class="select">
                <option value="0">Custom</option>
                <option value="1">All</option>
            </select>
        </div>
        <div id="permissions_container">
            <div id="permissions-tree" class="hor-scroll"><ul class="maho-tree hide-root-node"><li data-id="__root__" data-text="Root" data-loaded=""><details open=""><summary><label class="label"><span class="icon folder"></span><input type="checkbox" name="undefined"><span>Root</span></label></summary><ul></ul></details></li></ul></div>
        </div>
    </div>
</div>
```

fballiano:
Fixes #1166 and #1167.

### Empty permission tree (#1166)

The role editor builds all three of its sections (public storefront, customer session, service account tree) from `ApiPermissionRegistry`, which is backed by `vendor/composer/maho_api_permissions.php`. When that file is missing, or was compiled to an empty registry, the registry silently served an empty map: the reporter's screenshot shows all three sections blank, not just the tree.

The composer plugin only runs the API permission compile when it can autoload the API Platform classes during `post-autoload-dump`, so an install can legitimately end up without the file and with no visible reason.

- `ApiPermissionRegistry::load()` now rebuilds the map in-process (once per process, via the plugin's `compileRuntime()` entry point, the same one the admin cache page uses) when the file is missing or holds an empty registry, then re-reads it.
- If the rebuild is impossible too, for example a read-only `vendor/`, the empty-registry fallback and the log warning stay, and the permissions tab renders an explicit notice instead of a bare root node.

### 405 on role delete (#1167)

`deleteAction` was declared `methods: ['POST']`, but the delete button is generated by `Mage_Adminhtml_Block_Widget_Form_Container` and navigates with `setLocation()`, so the route answered 405 and a role could never be deleted. It now matches GET like every other admin delete action in core; CSRF stays covered by `_setForcedFormKeyActions(['delete', 'save'])` and the secret key URL.

### Tests

Both were written first and fail against the unfixed code:

- `tests/Backend/Integration/ApiPlatform/AdminRoleRouteTest.php` matches the delete path under GET and POST, and keeps a control assertion that `save` stays POST-only.
- `tests/Backend/Integration/ApiPlatform/ApiPermissionRegistryCompileTest.php` removes / empties the compiled map and asserts the registry recovers, that it rebuilds at most once per process, and that the permissions tab produces a tree with checkable leaf nodes.

> [!NOTE]


fballiano:
@empiricompany posso chiederti se puoi provare se questa PR ti risolve?

empiricompany:
```
Loading composer repositories with package information
Updating dependencies
Lock file operations: 0 installs, 1 update, 0 removals
  - Upgrading mahocommerce/maho (dev-main dfe006c => dev-fix-api-role-admin 1a3faa0)
Writing lock file
Installing dependencies from lock file (including require-dev)
Package operations: 0 installs, 1 update, 0 removals
  - Downloading mahocommerce/maho (dev-fix-api-role-admin 1a3faa0)
  - Upgrading mahocommerce/maho (dev-main dfe006c => dev-fix-api-role-admin 1a3faa0): Extracting archive
Generating autoload files
Maho: Compiling PHP attributes...
Maho: PHP attributes compiled successfully.
62 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
No security vulnerability advisories found.

```

@fballiano adesso vedo i permessi riesco ad eliminare il ruolo ma l'eliminazione del utente api ritorna sempre 405 Method Not Allowed
immagine

fballiano:
@empiricompany grazie per il check! ora dovrebbe essere a posto anche l'altra cosa!

empiricompany:
@fballiano testato confermo! grazie

PR #1193, opened 2026-07-30
Migrated the FedEx carrier from SOAP to the REST API

### LINKED ISSUE #918: Migrate FedEx SOAP integration to REST (`whatarmy/fedex-rest` optional dependency)
# Migrate FedEx SOAP → REST (whatarmy/fedex-rest, optional dependency)

## Context

The current FedEx carrier integration at `app/code/core/Mage/Usa/Model/Shipping/Carrier/Fedex.php` (1445 lines) uses three SOAP services (`RateService_v10`, `ShipService_v10`, `TrackService_v5`) via PHP's built-in `SoapClient` and WSDL files shipped at `app/code/core/Mage/Usa/etc/wsdl/FedEx/`. FedEx has been deprecating its SOAP web services and pushing all integrations to the REST API (OAuth2 Client Credentials, JSON over HTTPS); the SOAP endpoints will eventually stop working.

We will migrate to FedEx REST using `whatarmy/fedex-rest` (MIT, PHP ^8.0, OAuth2, actively maintained — covers Rates, Tracking, Shipments, AddressValidation, LocationSearch). The library will be an **optional Composer dependency** (declared in `suggest`, not `require`) so the default Maho install stays lean for users who don't ship via FedEx. When the library is absent, the carrier degrades silently in the storefront and surfaces a clear actionable error in admin.

The implementation mirrors the existing USPS REST pattern (`Mage_Usa_Model_Shipping_Carrier_Usps_RestClient` + `Mage_Usa_Model_Shipping_Carrier_Usps_OAuthClient`), which is the most modern carrier integration in the codebase and the right local convention to reuse.

### 1. Composer — declare optional dependency
`composer.json`:
- Add to the `suggest` block: `"whatarmy/fedex-rest": "Required for FedEx shipping rates, labels, and tracking via the REST API (^0.6)"`.
- Do **not** add to `require`. Do not remove `ext-soap` (other carriers/modules may still depend on it via the `suggest` for `laminas/laminas-soap`).
- `composer-runtime-api: ^2` is already required (composer.json:41), so `\Composer\InstalledVersions::isInstalled()` is available with no extra dependency.

### 1b. Admin-side missing-package signal — follow the SMTP convention
Maho already has a convention for "optional Composer package required for this admin option," used by the email transport dropdown at `app/code/core/Mage/Adminhtml/Model/System/Config/Source/Email/Transport.php:86-89`. It calls `\Composer\InstalledVersions::isInstalled('vendor/package')` inside the source model and suffixes the affected option's label with ` ⚠️ Install vendor/package` when the package is missing. We apply the same pattern, not a new system.xml tag.

Create `app/code/core/Mage/Adminhtml/Model/System/Config/Source/Carrier/Fedex/Active.php` (or co-locate inside Mage_Usa — preferred: `app/code/core/Mage/Usa/Model/Shipping/Carrier/Fedex/Source/Active.php` to keep FedEx code together):
- Standard yes/no source model returning `[['value' => 0, 'label' => 'No'], ['value' => 1, 'label' => 'Yes']]`.
- If `!\Composer\InstalledVersions::isInstalled('whatarmy/fedex-rest')`, append ` ⚠️ Install whatarmy/fedex-rest` to the "Yes" option label (mirroring the SMTP code).

In `system.xml`, point the existing FedEx `<active>` field's `<source_model>` to this new class instead of `adminhtml/system_config_source_yesno`.

This gives the admin a clear, contextual install hint right next to the "Enable" toggle — no new framework mechanism needed, fully consistent with the existing local convention.

### 2. Config & credentials — replace SOAP fields with REST fields
`app/code/core/Mage/Usa/etc/config.xml`:
- Bump `<Mage_Usa><version>` from `1.6.0.3` → `1.6.0.4`.
- Under `default/carriers/fedex`, remove `<meter_number>`, `<key>`, `<password>` nodes (SOAP-only).
- Add `<client_id backend_model="adminhtml/system_config_backend_encrypted"/>` and `<client_secret backend_model="adminhtml/system_config_backend_encrypted"/>`.
- Keep `<account>`, `<sandbox_mode>`, `<allowed_methods>`, `<dropoff>`, `<packaging>`, `<smartpost_hubid>`, `<unit_of_measure>` etc. — these are still valid for REST.

`app/code/core/Mage/Usa/etc/system.xml`:
- Replace the `meter_number`, `key`, `password` `<fields>` definitions (lines 24-89 area) with `client_id` and `client_secret`. Both encrypted, same sort order range, labels "Client ID" and "Client Secret".
- Leave the rest of the FedEx section unchanged.

### 3. Upgrade script — clean up obsolete config
Create `app/code/core/Mage/Usa/sql/usa_setup/upgrade-1.6.0.3-1.6.0.4.php`:
- Delete rows from `core_config_data` where `path IN ('carriers/fedex/meter_number','carriers/fedex/key','carriers/fedex/password')`. These are now meaningless and may contain stale encrypted SOAP credentials.
- Existing `account` values are preserved (still used by REST as the shipping account number).

### 4. OAuth token caching — mirror USPS pattern
Create `app/code/core/Mage/Usa/Model/Shipping/Carrier/Fedex/OAuthClient.php`:
- Almost a structural copy of `Mage_Usa_Model_Shipping_Carrier_Usps_OAuthClient` at `app/code/core/Mage/Usa/Model/Shipping/Carrier/Usps/OAuthClient.php`.
- Constructor: `(string $clientId, string $clientSecret, bool $sandbox)`.
- Resolves base URL from `$sandbox` (`https://apis-sandbox.fedex.com` vs `https://apis.fedex.com`).
- `getAccessToken(): string` returns cached or fetches a new token via POST to `/oauth/token` with `grant_type=client_credentials` (form-encoded body, not JSON — FedEx uses `application/x-www-form-urlencoded` for its OAuth endpoint).
- Caches the token in `Mage::app()->getCache()` keyed by `md5($clientId.$baseUrl)`, with TTL `max(60, expires_in - 300)`.
- Tag the cache entry `['fedex_oauth']` so it can be flushed alongside other shipping caches.

### 5. Rewrite the carrier — `Fedex.php`
`app/code/core/Mage/Usa/Model/Shipping/Carrier/Fedex.php` — full rewrite of the SOAP-touching code paths. Keep all public method signatures (`collectRates`, `getTracking`, `getAllowedMethods`, `getContainerTypes`, `getDeliveryConfirmationTypes`, `rollBack`, `_doShipmentRequest`) so callers (`Mage_Shipping_Model_Shipping`, admin shipment block, tracking display) keep working.

Per-method changes:

- **Constructor** — drop the WSDL path setup (lines 92-99). Drop `$_rateServiceWsdl`, `$_shipServiceWsdl`, `$_trackServiceWsdl` properties.
- **`_createSoapClient`/`_createRateSoapClient`/`_createShipSoapClient`/`_createTrackSoapClient`** — delete entirely (lines 108-148).
- **Library guard** — add a single `protected function _isLibraryAvailable(): bool` that returns `\Composer\InstalledVersions::isInstalled('whatarmy/fedex-rest')`. Matches the SMTP convention exactly; avoids depending on a specific class name from the library (less fragile if the library reorganizes namespaces between releases).
- **`collectRates`** — at the top, if `!$this->_isLibraryAvailable()`, log once via `Mage::log('FedEx REST library not installed; run: composer require whatarmy/fedex-rest', Mage::LOG_WARN)` and return `false`. The storefront then renders no FedEx rates, same as if the carrier were inactive.
- **`_formRateRequest`** — adapt the payload shape from SOAP (`WebAuthenticationDetail`/`ClientDetail`/`RequestedShipment`) to REST (`accountNumber`/`requestedShipment` per the library's `Rate` request DTO). Keep `SMART_POST` branch logic but switch to REST equivalents (FedEx REST still accepts `SMART_POST` service type; verify against sandbox).
- **`_doRatesRequest`** — replace SoapClient.getRates() (line 384) with the library's `Rate->request()` (or equivalent). Keep the `_getCachedQuotes`/`_setCachedQuotes` wrapping — but serialize the **array** response now, not a stdClass.
- **`_prepareRateResponse`** — rewrite to read array keys instead of stdClass properties (`$response['output']['rateReplyDetails']` per FedEx REST shape, vs `$response->RateReplyDetails`). Keep the rate-result construction and the `$allowedMethods` filter intact.
- **`_getXMLTracking`** — replace the SOAP `track()` call (line 931) with the library's tracking service (`TrackByTrackingNumber`). Adapt the payload and response parsing. Rename the method to `_doTrackingRequest` since "XML" is no longer accurate (small clarity win, low blast radius — `getTracking` is the public entry point).
- **`_parseTrackingResponse`** — rewrite array-key access for the REST response shape (`$response['output']['completeTrackResults'][0]['trackResults'][0]` and nested fields like `latestStatusDetail`, `dateAndTimes`, `scanEvents`).
- **`_doShipmentRequest`** — replace `processShipment()` SOAP call with the library's Shipments service. Map the existing `_formShipmentRequest` array to the REST `ship` payload (`requestedShipment` object with `shipper`, `recipients[]`, `shippingChargesPayment`, `labelSpecification`, `requestedPackageLineItems[]`). Decode the label PDF/PNG from REST response (`output.transactionShipments[0].pieceResponses[0].packageDocuments[0].encodedLabel` — base64) and put it into `$result->setShippingLabelContent(base64_decode(...))`. Set tracking number from `masterTrackingNumber` or `pieceResponses[0].trackingNumber`.
- **`rollBack`** — replace SoapClient.deleteShipment() (line 1371) with REST cancel call (`PUT /ship/v1/shipments/cancel`) via the library.
- **`_getAuthDetails`** — delete (lines 1101-1124). REST uses a Bearer header set on the HTTP client by the library, not embedded in the body.
- **`getVersionInfo`** — delete (no longer relevant; REST has no version block in the payload).

The library's authorization helper is instantiated once per carrier method call (or memoized on `$this`) using credentials from `$this->getConfigData('client_id'/'client_secret')` and `$this->getConfigFlag('sandbox_mode')`, with the bearer token resolved via `Mage_Usa_Model_Shipping_Carrier_Fedex_OAuthClient::getAccessToken()`. We use our own cache wrapper rather than the library's built-in fetch on every call because the library only recommends caching, doesn't implement it.

### 6. Remove dead WSDLs
Delete:
- `app/code/core/Mage/Usa/etc/wsdl/FedEx/RateService_v10.wsdl`
- `app/code/core/Mage/Usa/etc/wsdl/FedEx/RateService_v9.wsdl`
- `app/code/core/Mage/Usa/etc/wsdl/FedEx/ShipService_v10.wsdl`
- `app/code/core/Mage/Usa/etc/wsdl/FedEx/ShipService_v9.wsdl`
- `app/code/core/Mage/Usa/etc/wsdl/FedEx/TrackService_v5.wsdl`

Then `rmdir app/code/core/Mage/Usa/etc/wsdl/FedEx/` (and `wsdl/` if it's empty). Per AGENTS.md, removed files don't need backwards-compat shims.

### 7. Translations
`app/locale/en_US/Mage_Usa.csv` — add:
- `"Client ID","Client ID"`
- `"Client Secret","Client Secret"`
- `"FedEx REST library not installed; run: composer require whatarmy/fedex-rest","FedEx REST library not installed; run: composer require whatarmy/fedex-rest"`

Remove the now-unused strings for "Meter Number", "Key", "Password" if they were FedEx-specific (verify with grep — they may be shared with other contexts).

## Files

**Modified**
- `composer.json` — add to `suggest`
- `app/code/core/Mage/Usa/etc/config.xml` — version bump + credential nodes
- `app/code/core/Mage/Usa/etc/system.xml` — credential admin fields
- `app/code/core/Mage/Usa/Model/Shipping/Carrier/Fedex.php` — full SOAP → REST rewrite
- `app/locale/en_US/Mage_Usa.csv` — new labels

**Created**
- `app/code/core/Mage/Usa/Model/Shipping/Carrier/Fedex/OAuthClient.php`
- `app/code/core/Mage/Usa/Model/Shipping/Carrier/Fedex/Source/Active.php` — yes/no source with `⚠️ Install …` suffix, mirroring the SMTP `Transport.php` convention
- `app/code/core/Mage/Usa/sql/usa_setup/upgrade-1.6.0.3-1.6.0.4.php`

**Deleted**
- `app/code/core/Mage/Usa/etc/wsdl/FedEx/*.wsdl` (5 files)

## Patterns reused (no need to recreate)

- `Mage_Adminhtml_Model_System_Config_Source_Email_Transport` (lines 13-95) — exact template for the "warn when package missing" UX in our `Fedex/Source/Active.php`. Uses `\Composer\InstalledVersions::isInstalled()`.
- `Mage_Usa_Model_Shipping_Carrier_Usps_OAuthClient` — exact template for our `Fedex/OAuthClient.php`
- `Mage_Usa_Model_Shipping_Carrier_Usps_RestClient::makeRequest()` — error-handling shape (catch, extract JSON `error.errors[]`, debug-log via `Mage::log`) — apply the same defensive structure inside our `_doRatesRequest`/`_doTrackingRequest`/`_doShipmentRequest`
- `_getCachedQuotes` / `_setCachedQuotes` on the abstract carrier — keep using these for rate/tracking response caching
- `\Symfony\Component\HttpClient\HttpClient` — already used by USPS for token fetching; reuse in `Fedex/OAuthClient.php`
- `Mage::helper('core')->jsonDecode/jsonEncode` — for any JSON in/out we do directly

## Verification

End-to-end smoke test, in order:

1. **Library absent path**
   - With `whatarmy/fedex-rest` **not** installed, set `carriers/fedex/active=1` in admin and add an item to cart with a shippable address. Confirm FedEx rates do not appear in checkout and no fatal error fires. Tail `var/log/system.log` and confirm the "FedEx REST library not installed" notice was logged.

2. **Install library**
   - `composer require whatarmy/fedex-rest`
   - Clear caches: `./maho cache:flush`

3. **OAuth caching**
   - Trigger a rate quote (cart shipping estimator). Confirm a token request hits `/oauth/token` exactly once; the second rate quote within the cache TTL must reuse the cached token (verify by tailing the FedEx debug log or temporarily logging from `OAuthClient::fetchNewToken`).

4. **Rate quoting**
   - Set FedEx sandbox credentials in admin (`account`, `client_id`, `client_secret`, `sandbox_mode=1`). With a typical US-domestic shippable cart, confirm at least one FedEx rate renders and matches FedEx Developer Portal's expected sandbox response.
   - Try an international destination (US → Canada) and confirm `INTERNATIONAL_PRIORITY` / `INTERNATIONAL_ECONOMY` rates appear.

5. **Shipment label generation**
   - From admin, open an order with FedEx as the shipping method, click "Ship", confirm a label PNG/PDF downloads and the tracking number is stored on the shipment record (`sales/shipment.tracks`).

6. **Tracking lookup**
   - With the tracking number from step 5, click the public "Track Order" link or call `Mage::getModel('shipping/info')->loadByHash(...)` for the order. Confirm status events and delivery details render.

7. **Cancellation**
   - In a multi-package shipment flow, force the second package request to fail and confirm `rollBack()` invokes the REST cancel endpoint for the first package's tracking ID (manual code trace + log inspection — multi-package failure is hard to reproduce naturally).

8. **Regression**
   - `composer test -- --testsuite=Backend` — no new failures.
   - `vendor/bin/phpstan analyze app/code/core/Mage/Usa` — clean at level 6.
   - `vendor/bin/php-cs-fixer fix app/code/core/Mage/Usa` — no style violations.

9. **Service-code compatibility check**
   - The library/FedEx REST API may not accept every legacy service code in `allowed_methods` (`SMART_POST`, `FEDEX_FREIGHT`, `FEDEX_NATIONAL_FREIGHT`, `GROUND_HOME_DELIVERY`, etc.). During rate testing, watch for "invalid service type" responses and either map the affected codes to their REST equivalents inside `_doRatesRequest` or drop them from the default `allowed_methods` in `config.xml`. Document any code aliases in a brief comment in `_doRatesRequest`.

fballiano:
Closes #918.

FedEx has been retiring its SOAP web services, so the carrier now talks REST. Mirrors the USPS REST carrier: a thin `Fedex/OAuthClient` plus `Fedex/RestClient` on Symfony HttpClient, no new Composer dependency. The three WSDLs are gone, along with `_createSoapClient`, `getVersionInfo`, `_getAuthDetails` and the already-dead `_parseXmlResponse`.

Credentials move from meter number / key / password to an OAuth2 client id and secret, both encrypted. Tokens are cached under a `fedex_oauth` tag for `expires_in - 300`.

### Decisions worth reviewing

**Renamed config values are migrated, not aliased.** The five SOAP dropoff types map onto REST pickup types, and `INTERNATIONAL_PRIORITY` becomes `FEDEX_INTERNATIONAL_PRIORITY` in `allowed_methods` and `free_method`. Aliasing at read time would have collapsed `FEDEX_INTERNATIONAL_PRIORITY` and `FEDEX_INTERNATIONAL_PRIORITY_EXPRESS` onto one code, letting one differently-priced service silently overwrite the other. The one place that does alias at read time is shipment creation: pre-migration orders keep the SOAP code in `shipping_method`, order rows are never rewritten, and without the rename FedEx would refuse every label for those orders.

**SmartPost rate requests deliberately carry no declared value.** SOAP sent `TotalInsuredValue`; the REST Rate API answers HTTP 400 to any `declaredValue` on a `SMART_POST` request, at any amount (verified against the sandbox). Nothing is lost: Ground Economy includes the first $100 of liability and sells no more.

**New `rate_endpoint` setting**, because FedEx exposes two rate products: `/rate/v1/rates/quotes`, and `/rate/v1/comprehensiverates/quotes` which registered Integrator Providers are required to use and which 403s on the other path. Both take the same payload and return the same shape, so one builder and one parser serve both. Defaults to standard, which is what a normal shipping account wants.

**Freight codes dropped from the default `allowed_methods`** — the Rate API doesn't quote freight, and FedEx Freight became a separate company in June 2026. Labels stay in `getCode('method')` so existing configs still render.

### Bugs found while testing

**`rollBack()` claimed success on a refused cancel.** FedEx refuses a cancellation with HTTP 200, no `errors[]`, and `output.cancelledShipment: false`. The old code ignored the response and returned `true` regardless, so a label left live during a multi-package failure was reported as rolled back — a label the merchant still gets billed for. It now checks the flag per package, logs the reason, and still attempts the remaining packages before returning false.

**`carriers/fedex/unit_of_measure` had no default in `config.xml`**, so it sent `weight.units: null`, which FedEx rejects outright. An unsaved FedEx config was unquotable. Pre-existing, not introduced here; defaults to `LB` now with a regression test.

### Review-round fixes

A code review pass hardened the first cut:

- An OAuth response without `access_token`, and any `Throwable` in a request, now become a carrier error instead of a fatal on the checkout shipping step.
- A 401 evicts the cached token and retries once, so credential rotation no longer wedges the carrier for the token TTL; the cache key includes the client secret.
- A ship response with no label and no tracking number is an error, not a silent success.
- Multi-package shipments store each package's own tracking number; the master number was being repeated onto every package.
- A $0 ACCOUNT rate no longer beats a real LIST rate (accounts without negotiated pricing quoted free shipping).
- `shipDatestamp`/`shipDateStamp` send the store-local date; the UTC date is the wrong day for stores far from UTC.
- FedEx-reported error payloads are logged at WARNING even with debug off; malformed 200 bodies become error results and are never cached as quotes.

### Testing

45 unit tests over payload building, response parsing and rollback, with the rate and tracking parsers checked against real captured sandbox responses committed as fixtures.

6 integration tests hit the live sandbox, gated on `FEDEX_SANDBOX_*` org secrets so they skip where credentials are absent:

- OAuth token exchange and cache reuse
- tracking, raw and through the carrier
- rates, raw and through `collectRates()` as checkout calls it
- label creation, asserted to be a real PDF, then cancelled again
- a refused cancellation, asserting the HTTP-200-with-false shape that `rollBack()` has to notice

The sandbox returns `SERVICE.UNAVAILABLE.ERROR` and `SYSTEM.UNEXPECTED.ERROR` at random on byte-identical payloads, roughly one call in three. The integration tests retry those two codes with backoff; every other error fails on the first attempt. Production code does not retry, to avoid doubling checkout latency.

Migration verified against a seeded database: SOAP credential rows deleted, `dropoff` and service codes rewritten, `INTERNATIONAL_PRIORITY_FREIGHT` and `EUROPE_FIRST_INTERNATIONAL_PRIORITY` correctly untouched.


fballiano:
Shipping providers make it really hard to test, requiring real addresses in the contries they operate just to have a developer account, giving you different APIs cause you're an integrator instead of a shop... it's so much more difficult than payment providers.

So I tested what I could test, if you have a real developer account we can try with yours.

PR #1201, opened 2026-07-31
Fixed stored sessions expiring on session.gc_maxlifetime, removed session IDs from URLs, and isolated session keyspaces

hirale:
Three related changes: stored sessions now honour the configured lifetime, session IDs are gone from URLs, and each session name gets its own storage keyspace so an id cannot be presented under another name.

## 1. Stored sessions expired on `session.gc_maxlifetime`

`Mage_Core_Model_Session_Abstract::createSymfonySession()` computed a session lifetime and then discarded it, so nothing configured storage expiry: the Redis handler fell back to `session.gc_maxlifetime` (PHP's default `1440`) and stored sessions vanished after 24 minutes of inactivity regardless of configuration.

`app/code/core/Mage/Core/Model/Session/Abstract.php:103-116` on `main`:

```php
$adminLifetime = (int) Mage::getStoreConfig('admin/security/session_cookie_lifetime');
$frontendLifetime = (int) Mage::getStoreConfig('web/cookie/cookie_lifetime');
$sessionLifetime = max($adminLifetime, $frontendLifetime, 86400);   // never read again
```

1. `$sessionLifetime` is assigned and never used. The `86400` floor shows the intended minimum was one day.
2. It is a single `max()` of both settings applied to whichever session is being created, so `admin/security/session_cookie_lifetime` can never produce a *shorter* admin session than the storefront's. Raising the storefront lifetime silently raised the admin idle window.
3. Both areas already resolve the correct lifetime for their session cookie, in `Mage_Customer_Model_Observer::setCookieLifetime()` and `Mage_Adminhtml_Model_Observer::setCookieLifetime()`. Neither value ever reached the stored session.

### Impact

| | Configured | Effective |
| --- | --- | --- |
| Session cookie | `604800` (7 days) | 7 days |
| with Remember Me | `2592000` (30 days) | 30 days |
| Server-side session data | *(intended: `$sessionLifetime`)* | **1440 (24 min)** |

A customer holding a valid 30-day cookie was silently logged out after 24 idle minutes. For admin, `Mage_Admin_Model_Session::isLoggedIn()` is a plain session-payload check with no independent idle timeout, so this expiry *is* the admin idle timeout, and `admin/security/session_cookie_lifetime` had no effect on it.

### Reproduction

```php
Mage::getSingleton('customer/session')->setRememberMe(true);
Mage::getSingleton('core/session', ['name' => 'maho_session'])->start('maho_session');
$id = session_id();
session_write_close();
```

```
$ redis-cli -n 9 TTL maho_session:<id>
1440
```

The cookie is stamped for 30 days; the record it points at for 24 minutes.

The docs assume this TTL is meaningful: <https://mahocommerce.com/hosting/redis/> tells operators the session database needs no special `maxmemory-policy` handling because *"session keys always carry a TTL"*. The TTL is real, but it was a hardcoded 1440s that no documented setting reached.

### How the lifetime is resolved

Reaching the stored lifetime **destroys** the record, so it must not be computable from anything the request controls. Two inputs look natural and are both request-driven:

- `web/cookie/*` is `show_in_store`, and the store view comes from the request (`___store`, or the first path segment when `web/url/use_store` is on).
- Remember Me lives in the customer session, whose namespace is `customer_<websiteCode>` by default, so a cross-website request simply loses the flag.

Either one let an unauthenticated third party grade someone else's session under a shorter policy and destroy it. So the two values are resolved separately:

- **The cookie** keeps its exact per-request lifetime, per store view and honouring Remember Me. Identical to `main`.
- **The stored record** (Redis `ttl` and the read-time check) uses `resolveStoredSessionLifetime()`, which takes no per-request input: the default-scope admin policy for the admin namespace, and for everything else the maximum over all store views of both the plain and the Remember Me policy.

Because the record's value is the longest its area can grant, it always covers the cookie, so a valid cookie can never point at a record that was already destroyed.

### Resulting behavior

Stock configuration, measured on a clean install:

| | Cookie (unchanged) | Record before (Redis) | Record before (files) | Record after |
| --- | --- | --- | --- | --- |
| Storefront | 7 days | 24 min | never | 30 days |
| Storefront + Remember Me | 30 days | 24 min | never | 30 days |
| Admin | 3 hours | 24 min | never | 3 hours |

No new configuration fields, no changed defaults, no cookie handling changes.

### The read-time check

`gc_probability => '0'` means PHP never calls the handler's `gc()`, and `NativeFileSessionHandler` adds no read-time check, so a file session had no server-side expiry whatsoever. The only reaper was the `core_session_clean` cron.

Expiry is therefore decided where the session is read, in `start()`, which is handler-agnostic. It runs immediately after `$symfonySession->start()`, before the secure-cookie block and before the `session_before_renew_cookie` dispatch, so no observer sees a session that is about to be destroyed. The last-used stamp is read first, because constructing a session model calls `validate()` then `setValidatorSessionRenewTimestamp()` then `MetadataBag::stampNew()`, which resets the stamp to now and would make the check a silent no-op.

The data is emptied through the `$_SESSION` references rather than by calling `Session::invalidate()`. `invalidate()` reassigns `$_SESSION`, and PHP does not write through an existing element reference, so a session model already bound to a namespace would keep serving the expired data for the rest of the request. There is a regression test for exactly that.

`gc_probability` stays `'0'` deliberately. `session.gc_maxlifetime` is one process-wide value while the lifetime is per namespace and every namespace shares a save path, so an admin request that won the probability roll would delete storefront sessions on the admin's shorter threshold.

### Cleanup cron

`_cleanFileSystemSessions()` computed its threshold as `max(admin, cookie, 86400)` in a single scope, so it deleted 30-day Remember Me sessions on day 7 and ignored store views configuring a longer lifetime. It now takes the longest lifetime any store view can hand out, honouring `remember_enabled` and clamped to `SESSION_MAX_LIFETIME`. Read-time expiry enforces the policy, so the cron only reclaims disk. Stock retention goes from 7 to 30 days.

### Also

The lifetime clamps were duplicated across the two observers and the session model and had already drifted; they are now one `resolveConfiguredSessionLifetime()` both observers call, with `XML_PATH_*` constants for the config paths. `Mage_Customer_AccountController` clears Remember Me when credentials are rejected, since the flag is taken from the POST before the password is checked and a failed attempt left an unauthenticated visitor holding a 30-day cookie.

## 2. Removed session IDs from URLs

The change above raises how long a stored record lives, which is only safe if the cookie is the only way to reach it. It was not: `web/session/use_frontend_sid` defaulted to `1`, and `Mage_Core_Model_Url::isOwnOriginUrl()` returns true when there is no `Referer`, so `?SID=<id>` adopted a session with no cookie at all. A session id in a query string also leaks through `Referer` headers, access logs and browser history, and lets an unauthenticated request hand an id to a namespace that does not own it.

SID existed so multi-domain store views could share one session, which a cookie cannot do across registrable domains. SameSite defaults and ITP have made that unworkable regardless, so it is removed on both sides: the request is no longer read for a session id, and no URL is built with one.

Removed:

- `useSid()`, and the `$_GET['SID']` branch in `setSessionId()`
- `getSessionIdQueryParam()`, `getEncryptedSessionId()`, `getSessionIdForHost()`, `isValidForHost()`, `isValidForPath()`, `setSkipSessionIdFlag()`/`getSkipSessionIdFlag()`
- `Mage_Core_Model_Url::sessionUrlVar()`, `sessionVarCallback()`, `useSessionIdForUrl()`, `checkCookieDomains()`, `addSessionParam()`, `getRedirectUrl()`, `_prepareSessionUrl()`/`_prepareSessionUrlWithParams()`, `setUseSession()`/`getUseSession()`, and the `___SID` placeholder
- `Mage_Core_Model_App::setUseSessionInUrl()`/`getUseSessionInUrl()`/`setUseSessionVar()`/`getUseSessionVar()`
- `setUseSessionInUrl()` on the Email, Catalog and Cms template filters
- The block cache SID placeholder (`_getSidPlaceholder()` and the `_beforeCacheUrl()`/`_afterCacheUrl()` pair)
- The `_nosid` route param and its call sites
- The `web/session/use_frontend_sid` setting, from `config.xml` and `system.xml`

**Behaviour change:** multi-domain store views no longer share a session, so a customer moving between domains starts a new session there.

## 3. Gave each session name its own storage keyspace

Removing SID closed the *URL* path to presenting a foreign session id, but not the cookie one, which is more direct. A cookie *name* selects which header the server reads; it does not constrain the value, and an attacker sending the request themselves is under no browser constraint. `start()` takes the id straight from the cookie value and only charset-validates it.

Records were keyed on the session id alone while the lifetime came from the session name, so:

```
curl -b 'maho_admin_session=<storefront session id>' https://shop/admin/
```

read a storefront record, graded it on the admin's three hours, and destroyed it if idle longer than that. The admin session starts in `preDispatch()` before any auth check, so no login was needed. Destroying it was not even required for harm: the write-back at shutdown reset the record's ttl from 30 days to 10800 regardless.

So the keyspaces are now separate:

- The storefront keeps the bare keyspace, prefix and save path byte-identical, so existing customer sessions survive the upgrade.
- Every other session name gets `<key_prefix><name>:` on Redis and `var/session/<name>/` on the filesystem.

A foreign id now misses rather than reads, so neither the record nor its ttl is touched. This is a keyspace property rather than a guard, so there is no check that a later cleanup can delete on a plausible-looking reading. It also covers records written before the change, which a guard could not: they are simply not in the admin keyspace to begin with.

The subdirectory is created mirroring the parent directory's full mode rather than Symfony's `0777` minus the umask. Unlinking needs write and execute on the *directory*, so `0777` against the default umask 022 yields `755` and can leave a cron reaper running as a different user in the shared group unable to reap; and the mode is carried as `& 07777` so setgid survives, which is what keeps a shared web and cron group working. `mkdir()` is race-tolerant and its failure throws rather than silently falling back to the shared path. The cron reaper scans the subdirectories.

**Deploy note:** admin and any third-party session names move keyspace, so those sessions are dropped once and admins sign in again. No customer is affected.

## Notes for reviewers

- Storefront records now live 30 days on stock config rather than the reaper's previous 7. That is the price of removing the request-controlled destroy axis. Redis is strictly better than before either way, since `main` passed no `ttl` and dropped every session after 24 minutes.
- For a namespace with no observer of its own, nothing sets a cookie lifetime, so the cookie keeps `Mage_Core_Model_Cookie::DEFAULT_COOKIE_LIFETIME` (one year) while the record follows the configured policy. Such a session is replaced rather than resumed after the record expires. Giving the record a matching one-year floor was tried and reverted, because the cron reaper would have undercut it anyway and every extension session file would have been retained for a year.
- `Mage_Api_Model_Session` overrides `start()` and `setSessionId()` and is untouched; the API still takes its session id from the SOAP/XML-RPC payload.


hirale:
Hi @fballiano, If you have some time, I'd really appreciate a review. I'd like to make sure the direction looks reasonable before moving forward.

fballiano:
Thanks for this, and for the writeup. The diagnosis is exactly right: the lifetime was computed and thrown away, and the single `max()` of both settings was the wrong shape regardless. Reviewed it in depth and pushed four fixes on top rather than sending you round again.

**The admin namespace never got the admin lifetime.** Only the storefront namespace had its own branch, so `maho_admin_session` fell into `max(admin, frontend, 86400)` and was created with `gc_maxlifetime` = 604800 with stock config, i.e. seven days for a policy that says three hours. The Redis `ttl` callback corrected it once the observer ran, so it was invisible on Redis, but the value the session is created with was still 56x the policy, and `Mage_Admin_Model_Session::isLoggedIn()` has no independent idle check. It now resolves through its own branch clamped to the admin min/max, mirroring `Mage_Adminhtml_Model_Observer`. Your point 2 was right, it just needed the symmetric branch as well as the removal of the `max()`.

**Remember Me on unauthenticated sessions.** `AccountController::loginPostAction()` sets `remember_me` from the POST *before* `login()` runs and never cleared it on failure. Previously that only stretched the cookie; once the stored session follows the cookie, a bot POSTing `remember_me=1` with bogus credentials mints a 30-day session record per request instead of a 7-day one. Cleared now when the attempt does not end logged in. The 2FA and checkout-registration paths re-post the flag from the same form, so nothing legitimate loses it.

**Two smaller ones.** `getSessionLifetime()` is floored at 1, because a third-party observer passing `0` would make Redis `setEx()` fail and silently drop the session at shutdown; and it is public rather than private, since a non-public getter on a `Maho\DataObject` subclass falls through to `__call` and returns `null` instead of erroring.

Also trimmed the comments and docblocks to the house style and switched the new config reads to `Mage::getStoreConfigAsInt()`. Tests extended for the admin branch and the ttl floor, plus a frontend integration test for the Remember Me cleanup. `composer lint` and the full suite are green.

One thing worth recording: I tried changing `session_regenerate_id(false)` to `true` in the secure-cookie-check branch, because that path now orphans a record carrying the full configured lifetime instead of one expiring in ~24 minutes. That was wrong and I reverted it. The branch fires on a request that may not own the id it presented, so destroying the record hands anyone holding a session id a deletion primitive against its owner, and it also breaks cross-domain `SID` sharing. The longer-lived orphan stays as the lesser problem; the line now carries that reasoning.

Your *Known limitation* on the file handler stands and I left it alone: `gc_probability` is `0`, so nothing consumes `gc_maxlifetime`, and `_cleanFileSystemSessions()` uses one global threshold it cannot split per namespace. Read-time expiry is the right fix and is its own change. If you want to take that on as a follow-up, please do.

fballiano:
Followed up on the file handler after all, so the PR now covers both save handlers. Your *Known limitation* section called the shape of the fix correctly, including the `MetadataBag` as the place to read idle time from, so this is mostly your design.

**Where the check lives.** Expiry is decided in `start()`, which is handler-agnostic. Its placement is pinned from both sides. It has to run *after* the `session_before_renew_cookie` dispatch, because only then is the lifetime settled: on the storefront it depends on Remember Me, which lives inside the session an observer has to read to resolve it. That is precisely why a handler's `doRead()` can't make the decision, and why this couldn't be done as a custom file handler. But it has to be judged against the last-used stamp captured *before* those observers run, because constructing a session model calls `validate()` → `setValidatorSessionRenewTimestamp()` → `MetadataBag::stampNew()`, which resets the stamp to now. Read it after the dispatch and the whole feature is a silent no-op on the storefront.

**The part that surprised me.** The obvious `$symfonySession->invalidate()` is wrong here, and not subtly. It calls `NativeSessionStorage::clear()`, whose body is `$_SESSION = []`, and PHP does not write through an existing element reference when you reassign the container. `Mage_Customer_Model_Observer::setCookieLifetime()` constructs `customer/session` *inside* the dispatch, and `init()` binds `$_data = &$_SESSION['customer']`. So `invalidate()` leaves that singleton serving the pre-expiry data: `isLoggedIn()` keeps returning true against a record that was just emptied and re-keyed, for the rest of the request. I confirmed it rather than reasoning about it: the regression test fails with `invalidate()` on "Failed asserting that 42 is null". The fix is to empty through the references first, then `migrate(true)`.

Destroying the record is safe here, unlike the secure cookie branch we discussed: nothing the requester supplies moves the decision, only the stored stamp and store config.

**Cleanup cron.** Fixed alongside, and it had a second bug beyond the missing `remember_cookie_lifetime`: it reads one scope, but the cookie lifetimes are per store view, so a website configuring a longer lifetime had its files deleted on the default store's shorter threshold. It now takes the longest lifetime any store view can hand out, sharing one formula with the session model. Since read-time expiry does the actual policing, the cron only reclaims disk and is free to err long. Stock retention goes 7 → 30 days, which is worth knowing if you run a busy store.

**Removed `gc_maxlifetime`** from the storage options. With only file and Redis supported it reached nothing: files never collect, and the Redis `ttl` option is always set so the ini fallback never applies. `gc_probability` stays `'0'` on purpose, since one process-wide value can't express a per-namespace lifetime across a shared save path.

Two caveats worth recording. `start()` itself isn't unit-testable (it needs a real `session_start()`, which no test in this suite does), so the decision lives in a private method the tests drive by reflection and three wiring lines in `start()` are uncovered. And the idle window is measured from the *start* of the last request, since that's when `_sf2_meta['u']` is stamped, so a long-running request inflates the next measurement by its own duration; the cookie expires on the same schedule anyway, so it isn't the binding constraint.

14 new tests, full suite green at 3579.

fballiano:
Reviewed over several rounds and pushed fixes. The core bug and its fix stand; what changed is how the stored record's lifetime is resolved.

**The problem found:** the read-time expiry *destroys* the record, so grading it on a lifetime the request can influence let any request destroy a session it does not own. `web/cookie/*` is per store view and `___store` comes from the request; the Remember Me flag lives in a customer namespace that is per website. A cross-store or cross-website GET (an `` tag is enough, the session cookie defaults to `SameSite=None`) graded a victim's session under a shorter policy and wiped it. Two intermediate designs were tried and rejected: a monotonic max never decayed, so policy tightenings never reached live sessions and the value could be ratcheted up permanently; pinning to an establishing store just moved the attacker's choice to the first request.

**What it does now:** the cookie keeps its exact per-request lifetime, unchanged from `main`. The stored record (Redis TTL + idle expiry) uses `resolveStoredSessionLifetime()`, which takes no per-request input: the default-scope admin policy for admin, and for the storefront the max over all store views of both the plain and Remember Me policies. Being the longest, it always covers the cookie, so a valid cookie can never point at a record already destroyed.

Also in this push:
- A record is bound to the namespace that wrote it, so an id presented to another namespace via `?SID=` is not graded and re-stamped on the wrong policy.
- The lifetime clamps were duplicated across four sites and had already drifted; they are now one resolver both observers call, with `XML_PATH_*` constants for the config paths.
- The expiry check moved before the `session_before_renew_cookie` dispatch, so observers no longer see a session that is about to be destroyed.
- The filesystem reaper honours `remember_enabled` and is clamped to `SESSION_MAX_LIFETIME`.
- Test fixes: `IdleExpiryTest` fatalled on every case, the per-store-view reaper case was vacuous (`getStores()` hands back the same object `getStore()` does, so the loop overwrote its own premise), and cleanup left a dotfile behind so the temp dir was never removed.

42 tests pass, `composer lint` clean.

**Two things worth a decision:**

1. Storefront records now live as long as the longest configured policy (30 days on stock config) rather than per-request, up from the reaper's previous 7 days. That is the price of removing the attacker-controlled destroy axis. Redis is strictly better than before either way: `main` passed no `ttl`, so it fell back to `session.gc_maxlifetime` and dropped every session after 24 minutes.

2. `web/session/use_frontend_sid` defaults to `1` and `isOwnOriginUrl()` returns true when there is no `Referer`, so `?SID=<id>` adopts a session without the cookie. That is pre-existing and not made worse here, but it means the cookie is not the only access path, which weakens the argument that longer record retention is harmless. Worth addressing separately.

The PR description is now out of date in a few places: the `gc_maxlifetime` note describes a storage-options change that is not in the diff, the "after the dispatch because only then is the lifetime settled" rationale no longer applies, the behaviour tables predate this design, and the `AccountController` Remember Me fix is not mentioned.

fballiano:
Actually the SID option must go, it's a 2005 era thing, will update soon

fballiano:
Pushed the SID removal (`410b283`). The PR title and description now cover both changes.

SID in URLs is gone on both sides: the request is no longer read for a session id, and no URL is built with one. This is what makes the longer stored-record retention from the first commit defensible, since the cookie is now genuinely the only way to reach a session. Previously `web/session/use_frontend_sid` defaulted to `1` and `isOwnOriginUrl()` returns true when there is no `Referer`, so `?SID=<id>` adopted a session with no cookie at all.

Removed: `useSid()` and the `$_GET['SID']` branch in `setSessionId()`; `getSessionIdQueryParam()`, `getEncryptedSessionId()`, `getSessionIdForHost()`, `isValidForHost()`, `isValidForPath()`, the skip-session-id flag; `Mage_Core_Model_Url::sessionUrlVar()`/`sessionVarCallback()`/`useSessionIdForUrl()`/`checkCookieDomains()`/`addSessionParam()`/`getRedirectUrl()`/`_prepareSessionUrl*()`/`setUseSession()`/`getUseSession()` and the `___SID` placeholder; `App::setUseSessionInUrl()`/`getUseSessionInUrl()`/`setUseSessionVar()`/`getUseSessionVar()`; `setUseSessionInUrl()` on the three template filters; the block-cache SID placeholder and the `_beforeCacheUrl()`/`_afterCacheUrl()` pair; the `_nosid` route param and its call sites; and `web/session/use_frontend_sid` from both `config.xml` and `system.xml`.

**Behaviour change worth flagging:** SID existed so multi-domain store views could share one session, which is what the setting's admin comment described ("Allows customers to stay logged in when switching between different stores"). Customers moving between domains now start a new session on each. SameSite defaults and ITP had already made cross-domain session sharing unworkable, so this is mostly formalising what browsers already enforce, but it is a real change for multi-domain merchants.

Two knock-on simplifications: the namespace stamp I added in the previous commit is now redundant (a session id can only come from the cookie, and cookies are named per namespace) so it was removed, and the `setUseSessionInUrl(false)` workaround in `HttpsEnforcementTest` is gone, since the branch it was avoiding no longer exists.

`Mage_Api_Model_Session` overrides `start()` and `setSessionId()` and is untouched, so the API still takes its session id from the SOAP/XML-RPC payload.

`composer lint` is clean. I ran Backend and Frontend locally: Frontend is fully green, and Backend shows the same 21 failures with and without these changes (verified by stashing), all pre-existing DB-state pollution from running `pest` directly without the harness rebuild. Leaving the real verification to CI.

Unrelated bug spotted while mapping the store-in-URL code, filed separately as #1202: the admin store switcher concatenates `store/<id>/` onto a URL that may end in a query string, corrupting `form_key`.

hirale:
Removing SID is the right call, and it settles the retention question — with the
cookie as the only way to reach a session, a record outliving its cookie is a
storage cost rather than an exposure. No objection left to the 30-day lifetime.

One thing in `410b283` is a regression, though. I reproduced it and then fixed it,
both against a real checkout of that commit.

**The problem.** The namespace stamp was not redundant. Removing SID closed the
*URL* path to presenting a foreign session id, but the cookie path is still open
and is the more direct one: the cookie *name* selects which header the server
reads, it does not constrain what the client puts in the value. No browser and no
CSRF — the attacker sends the request themselves.

```
BEFORE   EXISTS=1  TTL=2592000     storefront maho_session, idle 4h
ATTACK   curl -b 'maho_admin_session=<storefront id>' http://…/admin/   → 200
AFTER    EXISTS=0  TTL=-2          destroyed
```

`start()` takes the id straight from the client cookie (`Abstract.php:275-276`)
and `setSessionId()` only charset-validates it; there is no ownership or namespace
binding left, because that commit removed the only one. The admin session starts
before any auth check (`Adminhtml/Controller/Action.php:167` precedes `:171`),
`sessionLifetime` is resolved as the admin policy at `:89` (10800 stock), and
`expireIdleSession()` at `:287` destroys anything idle longer than that. So a
storefront record on a 30-day policy is graded on three hours.

Severity is medium: it needs a known victim session id and a 3h–30d idle window,
and the impact is integrity/availability only — the record is destroyed, not
disclosed, and there is no privilege escalation. But it is exactly the axis the
redesign exists to close, and it is now unguarded where `isForeignNamespace()`
used to catch any stamped record.

**The fix** is to restore the binding you wrote in `baf5b7a76`, unchanged. Nothing
new, and your existing tests pass as they are — 36 passed in
`tests/Backend/Unit/Core/Model/Session/`. Applies cleanly on `410b283`:

```diff
diff --git a/app/code/core/Mage/Core/Model/Session/Abstract.php b/app/code/core/Mage/Core/Model/Session/Abstract.php
--- a/app/code/core/Mage/Core/Model/Session/Abstract.php
+++ b/app/code/core/Mage/Core/Model/Session/Abstract.php
@@ -37,6 +37,7 @@ class Mage_Core_Model_Session_Abstract extends \Maho\DataObject
     public const VALIDATOR_REMOTE_ADDR_KEY             = 'remote_addr';
     public const VALIDATOR_PASSWORD_CREATE_TIMESTAMP   = 'password_create_timestamp';
     public const SECURE_COOKIE_CHECK_KEY               = '_secure_cookie_check';
+    public const NAMESPACE_KEY                         = '_session_namespace';
 
     public const XML_PATH_COOKIE_DOMAIN        = 'web/cookie/cookie_domain';
     public const XML_PATH_COOKIE_PATH          = 'web/cookie/cookie_path';
@@ -284,7 +285,13 @@ class Mage_Core_Model_Session_Abstract extends \Maho\DataObject
         // Read before anything constructs a session model, whose validate() re-stamps this
         $lastUsed = $symfonySession->getMetadataBag()->getLastUsed();
 
-        $this->expireIdleSession($symfonySession, $lastUsed);
+        if ($this->isForeignNamespace()) {
+            // Kept rather than destroyed: the requester may not own the id it presented
+            session_regenerate_id(false);
+            session_unset();
+        } else {
+            $this->expireIdleSession($symfonySession, $lastUsed);
+        }
 
         // Secure cookie check to prevent MITM attack
         if (Mage::app()->getFrontController()->getRequest()->isSecure() && !$cookie->isSecure()) {
@@ -327,6 +334,9 @@ class Mage_Core_Model_Session_Abstract extends \Maho\DataObject
         // Observers can change settings of the cookie such as lifetime, regenerate the session id, etc
         Mage::dispatchEvent('session_before_renew_cookie', ['cookie' => $cookie, 'session_name' => $sessionName]);
 
+        // Last, so the branches above that clear $_SESSION cannot leave the record unstamped
+        $_SESSION[self::NAMESPACE_KEY] = $sessionName;
+
         // Set or renew regular session cookie
         $this->setSessionCookie();
 
@@ -340,6 +350,19 @@ class Mage_Core_Model_Session_Abstract extends \Maho\DataObject
         return $this;
     }
 
+    /**
+     * A record is keyed on the session id alone while its lifetime comes from the namespace, and a
+     * cookie name does not constrain the value a requester sends, so an id presented to another
+     * namespace would be graded and re-stamped on the wrong policy. Records written before this key
+     * existed carry no namespace and are adopted rather than rejected.
+     */
+    private function isForeignNamespace(): bool
+    {
+        $recorded = $_SESSION[self::NAMESPACE_KEY] ?? null;
+
+        return is_string($recorded) && $recorded !== $this->getSessionName();
+    }
+
     private function expireIdleSession(Session $symfonySession, int $lastUsed): bool
     {
         if ($lastUsed <= 0 || time() - $lastUsed <= $this->getSessionLifetime()) {
```

Verified over HTTP against `410b283` plus that patch:

| case | expected | result |
| --- | --- | --- |
| stamped storefront record + forged admin request | survives | ✅ (destroyed without the patch) |
| storefront record idle 40d, own namespace | destroyed | ✅ |
| storefront record idle 4h, own namespace | survives | ✅ |
| admin's own record idle 4h (>3h policy) | destroyed | ✅ idle timeout intact |


**One suggestion, separate from the fix** 

The guard closes the hole, but it is still a guard: something a later cleanup can
remove on a reading that looks correct, which is how we got here. The structural
version is to stop namespaces sharing a keyspace at all — keep the storefront
prefix and path byte-identical, and give every other namespace its own
(`<key_prefix><name>:`, `var/session/<name>/`). An id presented to the admin
namespace then simply misses, and the victim's record is never read.

It only needs to be the non-storefront namespaces, because the attack is
asymmetric: the damaging direction is a storefront record graded on the admin's
three hours, never the reverse. So no customer is logged out by the change — only
admins, once. It also closes the one case the guard cannot: records written before
the stamp exists are adopted, so they stay destroyable until each is touched once.

I prototyped it far enough to see it is viable — both save handlers, and the
Backend suite on a clean database gives the same 4 failures as `main` — but not far
enough to propose. The new subdirectory does not inherit the parent's permissions,
which would silently stop cron reaping admin sessions; `mkdir()` is unchecked and
racy; and the reaper's subdirectory scan has no test. Happy to work it up properly
if you want it.

hirale:
Correction to what I sent: the guard stops the record being destroyed, but not its TTL being rewritten — a forged admin read still resets a 30-day storefront record to 10800. Verified: TTL 2592000 → 10800 with the guard, 2592000 → 2591999 with per-namespace keyspaces. 

**So the keyspace split isn't the optional extra I framed it as; it's the part that actually closes it.**

fballiano:
You're right on both counts, and the mistake was mine. I removed the guard reasoning that "cookies are named per namespace, so an id can only come from the cookie" — which only holds for a browser. The cookie *name* selects which header the server reads; it does not constrain the value, and an attacker sending the request themselves has no such constraint. Thanks for reproducing it rather than just describing it.

Your correction is also right, and it matches something an earlier review round on this branch established empirically: `session_regenerate_id(false)` writes the current data back to the *old* id through the *current* handler before minting the new one. So the guard was always going to leave the ttl rewrite open. That rules out the guard as the fix rather than making it a partial one.

So I went straight to the keyspace split (`ba9c77b`) rather than landing the guard first:

- Storefront keeps the bare keyspace, byte-identical prefix and save path. No customer is logged out.
- Every other session name gets `<key_prefix><name>:` on Redis and `var/session/<name>/` on the filesystem. A foreign id misses instead of reading, so neither the record nor its ttl is touched. Admins sign in once after deploy.
- The guard stays removed. With separate keyspaces a record read under name X is always an X record, so `isForeignNamespace()` could never fire, and it would be the same dead-looking code that got deleted last time. This also closes the case you noted the guard could not: pre-stamp records are no longer adoptable, because they are not in the admin keyspace to begin with.

On the three gaps you flagged in your prototype:

- **Permissions.** `prepareSessionSaveDir()` mirrors the parent directory's mode instead of taking Symfony's `0777 & umask`. `mkdir()` subtracts the umask from its mode argument and `chmod()` does not, so the mode is set explicitly afterwards. Verified: subdir `750`, parent `750`.
- **Unchecked and racy mkdir.** `!@mkdir(...) && !is_dir($path)` so a concurrent request winning the race is not an error, and a genuine failure throws `Mage_Core_Exception` rather than silently falling back to the shared path.
- **Reaper.** `_getSessionSaveDirs()` returns the save path plus each readable subdirectory, with two tests covering it: an old record in a subdirectory is reaped, a recent one is kept.

Verified the separation directly: storefront resolves to `var/session-probe`, admin to `var/session-probe/maho_admin_session`, a third-party name to its own. 43 tests pass in `tests/Backend/Unit/Core/Model/Session/`, `composer lint` clean.

Two things worth your eye, since you have the reproduction harness and I do not:

1. I verified the keyspace separation at the handler level, not over HTTP. If you still have the curl setup, a run against `ba9c77b` would be worth more than my probe — specifically that a forged admin request leaves both `EXISTS` and `TTL` untouched, which is the case your correction was about.
2. Existing admin sessions are dropped on deploy. That is intended and I noted it in the PR description, but say so if you would rather it were migrated.

hirale:
Ran it over HTTP against `a62f9f6b3`. The case your correction was about is clean:

```
BEFORE   EXISTS=1  TTL=2592000     storefront maho_session, idle 4h
ATTACK   curl -b 'maho_admin_session=<storefront id>' /admin/   → 200
AFTER    EXISTS=1  TTL=2591999     untouched
```

Both EXISTS and TTL survive — the 1s delta is elapsed time, not a rewrite.

Also checked what the split could plausibly have broken, same harness:

- admin's own record idle 4h (>3h policy) → destroyed, idle timeout intact
- storefront record idle 40d (>30d policy) → destroyed
- file backend: records land in `var/session/` and `var/session/maho_admin_session/`,
  and the subdirectory is created with the parent's mode

On dropping existing admin sessions: leave it. Migrating them would mean reading the
old shared keyspace, which is the exact read this change closes — not worth reopening
to save admins one sign-in.

Thanks — good call skipping the guard rather than landing it first.

fballiano:
Thank you @hirale!!

PR #1208, opened 2026-08-01
Fixed composite SKU in Purchase events and tax-exclusive prices in Meta Pixel and GA4 tracking

fballiano:
Fixes #1207.

- Purchase events (Meta Pixel and GA4) now send the catalog product SKU instead of the order item's composite SKU (e.g. `basesku123-custom option`), falling back to the item SKU when the product no longer exists, so IDs match Meta Commerce / Google Merchant Center feeds.
- All monetary values are now tax-inclusive: product-based prices (view/list/cart/checkout events and the add/remove-from-cart observer) go through `Mage::helper('tax')->getPrice(..., true)`; quote/order item prices use `getBasePriceInclTax()` / `getBaseRowTotalInclTax()`. `tax_class_id` is added to the cloned category collections so the tax helper can compute the rate there.
- While touching the line: `begin_checkout` value now multiplies by item quantity, matching `view_cart`.
- The per-order purchase payload was extracted unchanged into a public `getPurchaseEventData()` on both blocks so it's testable with an in-memory order; tests cover SKU selection, incl-tax prices, and the deleted-product fallback.

> [!NOTE]
> **Developed with the help of AI.**<br>As part of our commitment to GenAI transparency, we flag pull requests produced with AI assistance alongside human work. As with every change in Maho, a maintainer reviews and validates it before merge, we never merge purely AI-generated changes. See the [GenAI transparency](https://github.com/MahoCommerce/maho#genai-transparency) section for details.

PR #1216, opened 2026-08-03
Fixed composite SKU sent in analytics events for products with custom options

fballiano:
Follow-up to #1208, fixes the issue reported in https://github.com/MahoCommerce/maho/pull/1208#issuecomment-5164266313.

`Mage_Catalog_Model_Product::getSku()` composes SKUs on cart products: custom options append their SKU suffix (`basesku123-mycustom option`), bundles append selection SKUs, and quote items re-attach custom options to the product, so every quote-based event (`begin_checkout`, `view_cart`, `InitiateCheckout`, add/remove-from-cart) sent composite SKUs; Purchase events sent the parent SKU instead of the variant SKU for configurables.

- Added `Mage_GoogleAnalytics_Helper_Data::getTrackingSku()`: the raw catalog SKU, resolved to the selected variant for configurables, never composed. Used in all SKU call sites of `Ga.php`, `Metapixel.php`, and the cart observer.
- Purchase events prefer the order item's stored `simple_sku` option so configurables report the variant SKU from freshly-loaded order items.
- Tests now build products that actually carry `option_ids` (the #1208 ones didn't, which is why they passed), plus configurable-variant cases and helper coverage.

> [!NOTE]

### DIFF
--- a/app/code/core/Mage/GoogleAnalytics/Block/Ga.php
+++ b/app/code/core/Mage/GoogleAnalytics/Block/Ga.php
@@ -214,7 +214,7 @@ gtag('set', 'user_id', '{$customer->getId()}');
                 $_product = $productInCart->getProduct();
                 $productPrice = (float) ($productInCart->getPriceInclTax() ?? $helper->getPriceInclTax($_product));
                 $_item = [
-                    'item_id' => $_product->getSku(),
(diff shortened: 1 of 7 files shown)


fballiano:
@empiricompany :-)

PR #1229, opened 2026-08-05
Fix gift card admin ACL resource

claudiobianco:
# Problem

`Maho_Giftcard_Adminhtml_Giftcard_PrintController` and `ProductController` check the ACL resource `giftcard/manage`, but `Maho/Giftcard/etc/adminhtml.xml` declares `admin/sales/giftcard/manage` (already used correctly by `GiftcardController` and `HistoryController`).

For an undeclared resource the Laminas ACL throws, and `Mage_Admin_Model_Session::isAllowed()` falls back to `$acl->isAllowed($role, null)`, the role's root rule. Administrators have `all => allow` there and never notice; every restricted role has `all => deny` and is denied, no matter which gift card permissions it was granted.

So a non-admin user with full gift card rights can edit gift cards but gets "Access denied" on:

- gift card PDF print (`giftcard_print/pdf`)
- mass PDF print (`giftcard_print/massPdf`)
- gift card email (`giftcard_print/email`)
- gift card product creation (`giftcard_product/create`, `.../save`)

# Fix

Use the declared resource `sales/giftcard/manage` in both controllers, in `ADMIN_RESOURCE` and in `_isAllowed()`.

# Verification

ACL loaded from the DB (`Mage::getResourceModel('admin/acl')->loadAcl()`):

```
admin/giftcard/manage        hasResource=false  -> throws, falls back to root rule
admin/sales/giftcard/manage  hasResource=true   -> admin=true, restricted role inheriting admin/sales=true
```

> [!NOTE]

### DIFF
--- a/app/code/core/Maho/Giftcard/controllers/Adminhtml/Giftcard/HistoryController.php
+++ b/app/code/core/Maho/Giftcard/controllers/Adminhtml/Giftcard/HistoryController.php
@@ -45,10 +45,4 @@ class Maho_Giftcard_Adminhtml_Giftcard_HistoryController extends Mage_Adminhtml_
         $this->loadLayout();
         $this->renderLayout();
     }
-
-    #[\Override]
-    protected function _isAllowed(): bool
-    {
-        return Mage::getSingleton('admin/session')->isAllowed('sales/giftcard/history');
-    }
 }
--- a/app/code/core/Maho/Giftcard/controllers/Adminhtml/Giftcard/PrintController.php
+++ b/app/code/core/Maho/Giftcard/controllers/Adminhtml/Giftcard/PrintController.php
@@ -10,7 +10,7 @@ declare(strict_types=1);
 
 class Maho_Giftcard_Adminhtml_Giftcard_PrintController extends Mage_Adminhtml_Controller_Action
 {
-    public const ADMIN_RESOURCE = 'giftcard/manage';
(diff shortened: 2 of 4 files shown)


claudiobianco:
Use sales/giftcard/manage for all actions or make different ACLs for the actions?

fballiano:
Pushed a follow-up commit removing the four `_isAllowed()` overrides in the gift card admin controllers. Each was identical to `Mage_Adminhtml_Controller_Action::_isAllowed()`, which already checks `static::ADMIN_RESOURCE`, so `ADMIN_RESOURCE` is now the only place the resource is named. That duplicated literal is what let the constant and the check drift apart in the first place. Did `GiftcardController` and `HistoryController` too, otherwise the same pattern stays around.

On your question (single resource vs per-action ACLs): keeping `sales/giftcard/manage` for all four actions. Flat is the dominant convention here, ~80% of admin controllers use one resource, and among Maho modules Blog, FeedManager, CatalogLinkRule, ContentVersion and MediaCleaner are all flat including their save/delete actions. Where the codebase does split (Revocation, Queue, sales order) it is along a privilege boundary, not per action, and Giftcard already has that split with `manage` vs `history`. Print and email are outputs of an entity the role can already edit outright.

The one action that arguably crosses a boundary is `ProductController` creating catalog products from a sales-scoped resource, but that is a real restriction on existing roles so it deserves its own PR.

fballiano:
Thank @claudiobianco!!

PR #1230, opened 2026-08-05
Fixed `GET /api/rest/v2/stores/currencies` returning 500 once any currency rate is saved

hirale:
The built-in `Currency` API resource returns HTTP 500 as soon as `directory_currency_rate` holds a
row for any allowed currency.

**The exchange rate is a string. The DTO property it is assigned to is a float.**

`Mage_Directory_Model_Resource_Currency::_getRatesByCode()`
(`app/code/core/Mage/Directory/Model/Resource/Currency.php:212-230`) returns the `rate` column
verbatim, with no cast. It is a `DECIMAL`, and the driver hands `DECIMAL` back as a PHP string:

```php
foreach ($rowSet as $row) {
    $result[$row['currency_to']] = $row['rate'];   // DECIMAL column → PHP string
}
```

So `getCurrencyRates()` returns quoted strings, not numbers — here on a store with base `USD` and
allowed `AUD,CAD,EUR,USD`:

```json
{ "AUD": "1.377220000000", "CAD": "1.323100000000", "EUR": "0.875880000000", "USD": "1.000000000000" }
```

`Mage\Directory\Api\Currency::$exchangeRate` is declared `?float`, and both files declare
`strict_types=1`, so no coercion happens. `CurrencyProvider::provide()` assigns the string
straight into the float property:

```php
// app/code/core/Mage/Directory/Api/CurrencyProvider.php:34
$dto->exchangeRate = $rates[$currencyCode] ?? null;   // string → ?float
```

```
TypeError: Cannot assign string to property Mage\Directory\Api\Currency::$exchangeRate
of type ?float
```

 The ?? null guard only covers the case where no rate row exists for that currency code. That state does not exist after a normal install: Mage/Directory/data/directory_setup/data-install-1.6.0.0.php:293 seeds EUR→EUR, EUR→USD, USD→EUR, USD→USD, and the shipped defaults are currency/options/base = USD, currency/options/allow = USD,EUR (Mage/Directory/etc/config.xml:117). Both allowed currencies therefore carry a rate row out of the box, so the endpoint returns 500 on a fresh install with no admin action at all. Saving rates in admin is not a prerequisite; the Reproduce steps below just make the failure explicit.

The 500 carries no detail, because `public/rest.php` boots the kernel with `debug=false`:

```json
{"error":"internal_server_error","message":"An internal error occurred","code":500}
```

### Reproduce

Tested on:

| | |
| --- | --- |
| Maho | `main` at `76327c891` (26.9.0) |
| PHP | 8.5.7 |
| MySQL | 8.4.9 |
| DB adapter | `Maho\Db\Adapter\Pdo\Mysql` — `pdo_mysql` + `mysqlnd`, `doctrine/dbal` 4.4.4 |

Only that combination was tested, but nothing in it looks load-bearing: `pdo_mysql` surfaces
`DECIMAL` as a PHP string to preserve precision, independently of emulated prepares, so any
install with a saved currency rate should hit this.

Prerequisite for both cases below: REST v2 is opt-in, so set `apiplatform/protocols/rest_v2 = 1`.
With it off, `ProtocolToggleListener` answers 404 `{"error":"protocol_disabled"}` before the
provider ever runs, which masks the bug.

Multi-currency:

1. `currency/options/base = USD`, `currency/options/allow = USD,EUR`.
2. Save rates in Admin → System → Currency Rates, so `directory_currency_rate` has `USD → EUR`.
3. `curl -s https://<host>/api/rest/v2/stores/currencies` → HTTP 500.

Single currency — same result:

1. `currency/options/base = USD`, `currency/options/allow = USD`.
2. Save rates once, so the self-rate row `USD → USD = 1.000000000000` exists.
3. Same call → HTTP 500.

The call returns 200 only while no row in `directory_currency_rate` matches an allowed currency
code — `$exchangeRate` is then never assigned a non-null value.

> [!NOTE]

### DIFF
--- a/app/code/core/Mage/Directory/Api/CurrencyProvider.php
+++ b/app/code/core/Mage/Directory/Api/CurrencyProvider.php
@@ -31,7 +31,10 @@ class CurrencyProvider extends \Maho\ApiPlatform\Provider
 
             $dto = Currency::fromModel($currency);
             $dto->symbol = $currency->getCurrencySymbol();
-            $dto->exchangeRate = $rates[$currencyCode] ?? null;
+            // Rates come from a DECIMAL column, so they arrive as strings. Under strict_types
+            // the assignment to ?float would throw for every currency that has a rate row.
+            $rate = $rates[$currencyCode] ?? null;
+            $dto->exchangeRate = $rate === null ? null : (float) $rate;
             $currencies[] = $dto;
         }
 
--- a/app/code/core/Mage/Directory/Api/CurrencyProvider.php
+++ b/app/code/core/Mage/Directory/Api/CurrencyProvider.php
@@ -31,10 +31,9 @@ class CurrencyProvider extends \Maho\ApiPlatform\Provider
 
             $dto = Currency::fromModel($currency);
             $dto->symbol = $currency->getCurrencySymbol();
-            // Rates come from a DECIMAL column, so they arrive as strings. Under strict_types
-            // the assignment to ?float would throw for every currency that has a rate row.
-            $rate = $rates[$currencyCode] ?? null;
-            $dto->exchangeRate = $rate === null ? null : (float) $rate;
+            // MySQL and PostgreSQL return the DECIMAL rate as a string, which strict_types
+            // refuses to assign to ?float. SQLite returns a number, so this never fails locally.
+            $dto->exchangeRate = isset($rates[$currencyCode]) ? (float) $rates[$currencyCode] : null;
             $currencies[] = $dto;
         }
 
--- a/tests/Api/V2/Read/CurrencyTest.php
+++ b/tests/Api/V2/Read/CurrencyTest.php
@@ -1,0 +1,96 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: 2026 Maho <https://mahocommerce.com>
+ * SPDX-License-Identifier: OSL-3.0
+ * @package Tests
+ */
+
+declare(strict_types=1);
+
+/**
+ * API v2 Currency Tests
+ *
+ * Tests for the currency endpoint - PUBLIC ACCESS (no auth required).
+ * All tests are READ-ONLY (safe for synced database).
+ *
+ * @group read
+ */
+
+describe('API v2 Currencies', function (): void {
+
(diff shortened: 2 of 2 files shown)


fballiano:
Pushed a regression test plus a small tweak to the fix.

The diagnosis is right, but the severity is understated. `directory_currency_rate` is seeded at install (`Mage/Directory/data/directory_setup/data-install-1.6.0.0.php:293`) with EUR/USD and both self-rates, and the default store is base EUR / allow EUR. So this returns 500 on a fresh install with no admin action at all. The "endpoint stays up only where rates were never saved" case does not really exist post-install. Worth updating the description.

It is also backend dependent, which is the part worth documenting in the code. MySQL and PostgreSQL hand back the DECIMAL as a string, but SQLite returns a number, and `int` widens to `float` even under `strict_types` (only `string` throws). So the endpoint works fine when developing against SQLite and the cast reads as redundant there. I reworded the comment to say that and folded the assignment back to one line:

```php
// MySQL and PostgreSQL return the DECIMAL rate as a string, which strict_types
// refuses to assign to ?float. SQLite returns a number, so this never fails locally.
$dto->exchangeRate = isset($rates[$currencyCode]) ? (float) $rates[$currencyCode] : null;
```

`tests/Api/V2/Read/CurrencyTest.php` needs no fixture thanks to the seeded self-rate. Besides the 200 it asserts `exchangeRate` is not a string and is numeric, which guards both dropping the cast and "fixing" the 500 by widening the property to `?string`. Note it can only go red against the unfixed code on the MySQL and PostgreSQL CI legs, not on SQLite.

I also swept every API DTO for the same hazard (uncast assignments into `int|float|bool` properties, with the pre-fix line here used as a control to confirm the check actually catches it). This was the only live instance. Everything else casts at the definition site, goes through a typed helper, or survives a JSON cache round trip. One unrelated leftover for a separate PR: `Mage/Cms/Api/MediaProcessor.php:164` assigns `filesize()` (`int|false`) to a `?int`, so a stat failure would be a TypeError rather than a null size.

hirale:
Thanks, the tweak and the test both look right to me. Description updated.

fballiano:
Thank you @hirale!!

PR #1240, opened 2026-08-06
Refused web requests when the database is behind the declared schema

### LINKED ISSUE #1239: Bug: maho_queue_message doesn't exist
## Desciption

After updating main to the latest version and running `composer install`, the following error message occurs when any page is rendered:
```
Error in file: "/var/www/html/app/code/core/Mage/Core/data/core_setup/data-upgrade-2.0.0-2.1.0.php" - An exception occurred while executing a query: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'db.maho_queue_message' doesn't exist

#0 /var/www/html/app/code/core/Mage/Core/Model/Resource/Setup.php(733): Mage::exception()
#1 /var/www/html/app/code/core/Mage/Core/Model/Resource/Setup.php(501): Mage_Core_Model_Resource_Setup->_modifyResourceDb()
#2 /var/www/html/app/code/core/Mage/Core/Model/Resource/Setup.php(300): Mage_Core_Model_Resource_Setup->_upgradeData()
#3 /var/www/html/app/code/core/Mage/Core/Model/Resource/Setup.php(284): Mage_Core_Model_Resource_Setup->applyDataUpdates()
#4 /var/www/html/app/code/core/Mage/Core/Model/App.php(325): Mage_Core_Model_Resource_Setup::applyAllDataUpdates()
#5 /var/www/html/app/Mage.php(663): Mage_Core_Model_App->run()
#6 /var/www/html/public/index.php(71): Mage::run()
#7 {main}
```

fballiano:
Fixes #1239.

The declarative schema (`sql/schema.php`) is applied only by `./maho migrate` and the installer, but legacy setup scripts still auto-apply on a web request. An upgraded store that hadn't run `migrate` ran `Mage_Core`'s `data-upgrade-2.0.0-2.1.0.php` against a database with no `maho_queue_message` table, so every page fatalled.

Rather than applying DDL implicitly (convergence can drop an index or FK no module declares, which is why it's a reviewed step), the gap is now detected:

- `Maho\Db\Schema\Status` fingerprints every active module's `sql/schema.php` plus the table prefix and compares it to a `declarative_schema` row in `core_resource`. On a mismatch the planner decides: an empty plan means the edit was cosmetic, so the new fingerprint is recorded and nothing blocks; a non-empty plan is pending.
- `Mage_Core_Model_App::isSchemaUpdatePending()` caches the verdict under its fingerprint, tagged with the config cache. `_initModules()` holds back `applyAllUpdates()` when pending (which also stops `applyAllDataUpdates()`), and `run()` serves a 503 page pointing at `./maho migrate`. `rest.php` and `api.php` return 503 as well.
- The CLI is deliberately not blocked, or `migrate` couldn't run; `BaseMahoCommand` prints a warning instead. `skip_process_modules_updates` installs are never held back.
- `migrate` and `Installer::installDb()` record the fingerprint after applying.

Measured on a 367-table install: 0.9 ms per boot in the steady state, 390 ms for the one-time introspection after a schema file changes.

> [!NOTE]


fballiano:
@tmewes This should help avoiding what you found with the unmigrated database schema ;-)

PR #1252, opened 2026-08-11
Fixed REST shipment creation never persisting qty_shipped

### LINKED ISSUE #1251: REST shipment create never persists qty_shipped — orders can be shipped repeatedly
## Summary

`POST /orders/{orderId}/shipments` registers the shipment, but `qty_shipped` is never persisted on the order items. As a result `canShip()` remains `true` — the same order can be shipped repeatedly, and invoiced orders never transition to `complete`.

## Root cause

The shipment is registered, but the order itself has no changed fields at save time. The order's save therefore short-circuits on `_hasModelChanged()`, and `_afterSave()` never runs — so the items' `qty_shipped` is never written.

## Steps to reproduce

1. Create an order with invoiceable/shippable items
2. `POST /orders/{orderId}/shipments` via the REST API
3. Observe the shipment is created, but the order items' `qty_shipped` is unchanged
4. Repeat step 2 — a second, duplicate shipment is created (`canShip()` still `true`)
5. If the order is invoiced, it never reaches status `complete`

## Expected behaviour

- `qty_shipped` is persisted after shipment creation
- `canShip()` returns `false` once fully shipped
- Fully invoiced + shipped orders transition to `complete`

## Proposed fix

In `ShipmentProcessor::buildAndRegisterShipment()`, add after `$shipment->register();`:

```php
$shipment->getOrder()->setIsInProcess(true);
```

This marks the order as changed so its save proceeds and `_afterSave()` persists the item quantities — matching the existing behaviour of the Adminhtml `ShipmentController` and of `InvoiceProcessor`.

## Verified

- Reproduced on `26.7.2` and `main@c674c40`
- With the fix applied: `qty_shipped` persists, invoiced orders reach `complete`, and re-shipping is blocked

## Minor related note

`Shipment::$items` is declared `writable: false`, so the OpenAPI spec marks it `readOnly` — but the POST body's `items` **is** honoured. Worth making it writable in the docs to match actual behaviour.

fballiano:
`ShipmentProcessor` registered the shipment but left the order with no changed data, so `Mage_Core_Model_Abstract::save()` short-circuited on `_hasModelChanged()` and never ran `_afterSave()`, where the order items `register()` had just updated get written. `qty_shipped` stayed 0, `canShip()` stayed true so an order could be shipped repeatedly, and `_checkState()` never ran so invoiced orders never reached `complete`.

Marking the order as in process before the transaction save fixes it, matching the adminhtml shipment controller and the SOAP API. Covers REST and GraphQL, both go through `doCreateShipment()`.

The shipment write tests also seeded nothing: `findShippableOrderId()` scanned the database and skipped when it found nothing, so the file was dead on a clean sample-data install. It now places a fresh cash-on-delivery order per test. That also removes the coupling that hid this bug, since the old scan kept returning the same order precisely because no shipment ever consumed it.

Fixes #1251

> [!NOTE]

### DIFF
--- a/app/code/core/Mage/Sales/Api/Shipment.php
+++ b/app/code/core/Mage/Sales/Api/Shipment.php
@@ -20,6 +20,8 @@ use ApiPlatform\Metadata\Link;
 use ApiPlatform\Metadata\GraphQl\Query;
 use ApiPlatform\Metadata\GraphQl\QueryCollection;
 use ApiPlatform\Metadata\GraphQl\Mutation;
+use ApiPlatform\OpenApi\Model\Operation as OpenApiOperation;
+use ApiPlatform\OpenApi\Model\RequestBody;
 use Maho\ApiPlatform\CrudResource;
 
 #[ApiResource(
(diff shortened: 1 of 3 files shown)


fballiano:
Thanks for the verification. Good catch on the spec: every `Shipment` DTO property is `writable: false`, so API Platform advertised no request body at all for `POST /orders/{orderId}/shipments`, even though the processor reads `items`, `tracks`, `comment` and `notifyCustomer` off the raw body. The same gap applied to the add-track and add-comment operations.

Pushed e37c391: explicit `openapi: new Operation(requestBody: ...)` on all three, matching the pattern in `Mage_Cms_Api_Media`. Added tests for the partial shipment path and for the spec itself, so the documented shape stays true.

PR #1264, opened 2026-08-12
Fixed the currency defects across the store model, storefront and APIs, and added display-currency selection

hirale:
A sweep for currency defects after #1248. 

**It is a big PR, but a single subject: everything in it is currency. One PR rather than several, because the fixes share machinery and each one surfaced the next.** 

## Test ENV
| | |
| --- | --- |
| Maho | [1480fd2](https://github.com/MahoCommerce/maho/tree/1480fd2610c03ed02b4af921ae891b259b098502) |
| Database | MySQL |
| Stores | one website; store views `default` and `eurview` |
| `currency/options/base` | `USD` |
| `currency/options/allow` | `USD,EUR` |
| `currency/options/default` | `EUR` |
| Rate | `USD -> EUR = 0.85` |
| Cart | one product at `29.99` base, qty 2 |

## 1. The display currency had two answers

`getCurrentCurrency()` converts the amounts; `getCurrentCurrencyCode()` labels them. Each resolved the currency on its own, and when they disagreed, amounts in one currency went out under the name of another. Two things made them disagree:

- **The memo.** A switch wrote the session but left the `current_currency` memo standing, so everything already converted kept the old currency.
- **The fallback.** `getCurrentCurrency()` falls back to base when the display currency has no usable rate; the code accessor never asked whether a rate existed.

Measured, with `allow = USD,GBP`, `default = GBP` and no `USD -> GBP` rate:

```
getCurrentCurrencyCode()   GBP      <- labels
getCurrentCurrency()       USD      <- converts
convertPrice(10.00)        10.00    <- unconverted, i.e. USD
```

Everything reading the code labeled those unconverted amounts `GBP`: analytics events, `Locale::getJsPriceFormat()`, layered nav, the admin order-create grid, block-cache keys. `Mage_Checkout_Model_Session::getQuote()` chased a code that could never match, recollecting and re-saving the cart on every request. Reaching this state takes no shopper, only configuration: rate import ships disabled, and the admin never checks that an allowed currency has a rate. Core already knew: the switcher block carried the comment "do not use `Mage::app()->getStore()->getCurrentCurrencyCode()` because of probability to get an invalid (without base rate) currency from code saved in session", and fixed only itself.

Fix: 3ad05686b (the memo), 25ac80f75 (the fallback).

## 2. The converted price lost its rate after the first read

`getPriceFilter()` converts and formats in one step, and memoises the filter it builds. The `elseif` chain hangs off the memo test instead of the currency test: only the first read got the display currency's filter, and every later read replaced it with the default currency's, whose rate is 1. Measured, base `USD`, display `EUR` at 0.85:

```
call 1   filter(100)   €85.00
call 2   filter(100)   €100.00
call 3   filter(100)   €100.00
```

Checkout renders through this filter twice on one page (`Mage_Checkout_Block_Onepage_Progress` and `Mage_Shipping_Model_Rate_Result::getRates()`), so whichever ran second was already wrong; the admin order-create shipping form is the third caller. Pre-existing, surfaced by reading the memo this PR started clearing.

Fix: 787a32261.

## 3. The currency cookie was write-only, so it is gone

`setCurrentCurrencyCode()` wrote a year-long `currency` cookie that no path in the tree reads back. The API made the cost visible: the no-rate fallback ran through the same setter on every `convertPrice()` call, so a stateless, unauthenticated API read answered with a `Set-Cookie`, client state the API will never read, on a response no shared cache will store:

```
GET /api/rest/v2/guest-carts/{maskedId}   X-Store-Code: eurview

before   Set-Cookie: currency=USD; expires=Wed, 11 Aug 2027 07:36:22 GMT; Max-Age=31536000; path=/; HttpOnly; SameSite=Lax
after    (no Set-Cookie)
```

Two defects went with it: the delete branch compared the currency code against a Currency object, a comparison that is never true in PHP 8, so switching back to the default re-wrote the cookie instead of deleting it; and admin order creation reached the same cookie write, though the admin has no use for a currency cookie at all.

The setter splits along that line. `setCurrentCurrencyCode()` records an explicit shopper choice and persists it to the session; the new `setRequestedCurrencyCode()` applies a currency for the current request and persists nothing, for callers whose currency belongs to the order or the request rather than to the shopper (admin order creation, the `X-Currency-Code` listener of section 8).

Fix: 58638c3e4 (the cookie), cfafbfd1e (the admin write), 396c5f7d4 (the setter split).

## 4. The cart reported one currency and amounts in another

`CartMapper` took the label from the stamped `quote_currency_code` while the amounts beside it are recomputed on every read at the live display currency:

```
sales_flat_quote row, unchanged throughout:
    quote_currency_code=EUR  base_to_quote_rate=0.8500  subtotal=50.98  base_subtotal=59.98

display USD   before   currency=EUR  price=29.99  rowTotal=59.98  subtotal=59.98
              after    currency=USD  price=29.99  rowTotal=59.98  subtotal=59.98
```

The `before` row served 59.98 USD under an EUR label, matching neither the label nor the database. The same split surfaced three more times:

- GraphQL `estimateShippingMethods`: the identical defect, live amounts under the stamped label;
- `applyGiftcard()`: the inverse, it read the client's amount as though it were in the stamped currency;
- the storefront gift-card widget: the card's stored value went through the stale stamp next to a total from the live currency, printing 50.00 and 42.50 for the same 50.00 card.

Fix: 91b69b0fd (the mapper), 8e3e9d601 (the estimate), 906d25a56 (the gift card amount), b610cbfd0 (the widget).

## 5. An order's currency answered differently depending on who asked

Orders can carry no `order_currency_code` (imports, migrations). Reading one, each surface improvised its own answer:

- invoices and credit memos: the viewer's display currency;
- order reads: the store's configured default;
- GraphQL: a bare `null`.

So the same invoice reported EUR to one viewer and USD to another while its amounts never moved. The currency is a property of the order, not of whoever is reading.

Fix: acce543ec (store, not viewer), 256853017 (one resolver for every reader), 5da73c52f (never an empty code), 3765e5c1d (store-less rows).

## 6. Orders were stamped in a currency their amounts were not in

`placeAdminOrder()` collected totals at the live display currency but never saved the quote. The quote-to-order conversion copies `quote_currency_code` and `base_to_quote_rate` from the stored columns, and only `_beforeSave()` refreshes those, so the order got whatever stamp the last save happened to record: an immutable label in the old currency over amounts in the new one, inherited by invoices, credit memos and reports. REST was safe by accident (`OrderProcessor` saves first); the GraphQL cartId-only path was not.

Fix: 35bbecd28, with d92f77ba4 giving the quote one way to restate its currency.

## 7. The MCP instruction line contradicted the tool output

`Kernel::mcpInstructions()` named the default view's display currency, but the line is compiled into the container, so it can never follow the store a tool call targets, and a backend token reads base currency regardless:

```
before   Prices and totals are in EUR unless a tool says otherwise.
after    Read the "currency" field of any response that carries one: cart and order amounts are
         in the currency named there, which is not always USD. Where no currency is given,
         amounts are in USD, the base currency of the default website; other websites may differ.
```

There is no multi-website test; standing up a second website in the suite is not worth the fixture cost.

Fix: 017dfaf2a, 596fda6b8, 8d7a8061d.

## 8. The gap: the API could not select a display currency

`currency/options/allow` is a multiselect and the storefront switcher already offers several display currencies on one store view. The API published that set through `GET /stores/currencies` but read no currency signal from the request, so every entry except the default was unreachable and a headless storefront needed one store view per currency.

`X-Currency-Code` now selects one. The decisions, and why:

- **A header, not a query parameter.** `Vary` cannot express a query parameter; `HttpCacheListener` already downgrades such responses to `private`, which would leave every priced read uncacheable by a CDN. `Vary` and the CORS allow list gained the header.
- **Never mislabel.** A currency the store cannot serve is a `400`, not base prices under a label the caller did not ask for.
- **No persistence.** Applied through `setRequestedCurrencyCode()`: an API request has no session, and the header speaks for one request only. `StoreContext::reset()` drops the store's currency memos afterwards, which matters wherever one process serves many requests (the test harness today; php-fpm production does not reuse it).
- **A cart adapts in its own store.** A cart is served by the store it was created in, not the API context, so the currency is re-applied there, on both load paths: `CartService::getCart()` and the `loadAdminQuote()` behind GraphQL's `placeOrder` and shipping estimate. Adapting rather than refusing follows the storefront, where `Mage_Checkout_Model_Session::getQuote()` recollects a cart stamped in one currency against the currency asked for. A write under the header stamps the cart the same way; the `base_*` columns never move, and the shopper's next storefront load recollects it back into their session currency.

Gift card presets stay in base and now carry `giftcardAmountCurrency` to say so: the client posts those values back to add-to-cart, where they are matched against the stored base-currency list, so converting them on the way out would break the round trip. A backend token also still reads base; `ProductProvider` keeps prices lossless for read-modify-write.

Feature: 2c8d78a02, with 3958a2d77 (the Vary constants), 5e11ab480 (the reset), c5dc60b20 and 678c59587 (the cart adaptation), 40e5d8eb3 (the gift card field).

## 9. Three answers to which currencies a store can serve

Which display currencies can this store serve? Four places asked, each with its own test:

- the storefront switcher: `isset()` on the rate;
- `GET /stores/currencies`: `(float) $rate <= 0`;
- the `X-Currency-Code` listener: a rate check of its own;
- `getCurrentCurrency()`: PHP truthiness on the string the `DECIMAL` column hands back, where `"0.0000"` is true.

Four tests, three definitions of one question. The visible symptom closes section 1's story: a `select` with nothing selected, the accessor saying `GBP` while the list offered only `USD`.

`Store::getServeableCurrencyRates()` is the single definition now: the allowed set minus anything without a usable rate, cast numeric once at the source; base needs no rate row but has to be allowed. No configuration reaches the zero-rate disagreement today (`Resource/Currency::saveRates()` skips zeros and every live writer goes through it); the point is that the surfaces cannot drift apart again.

Fix: 0bef2433a, pinned by a3289ed58.

## Compatibility

- `Mage_Core_Model_Store::getCurrentCurrencyCode()` now resolves through the no-rate fallback, so it can return the base currency where it used to return a code with no rate behind it. A caller that wants the choice itself rather than what it resolved to should use the new public `getRequestedCurrencyCode()`.
- No `currency` cookie is written any more. Nothing in the tree read it, but a custom theme or an edge rule outside the tree might. `Mage_Core_Model_Store::COOKIE_CURRENCY` stays, marked deprecated, so a module referencing the constant does not fatal on upgrade.
- `GET /stores/currencies` no longer lists an allowed currency that has no imported rate, matching the storefront switcher and the header.
- The `currency` field on orders, invoices and credit memos comes from the row rather than from the reader, so a row with no stamped code reports a different value than before.
- `Mage_Directory_Block_Currency::getCurrencies()` returns an empty array where only one currency is serveable, rather than that one, and `getCurrencyCount()` reports `0` instead of `1`. The stock template gates on `> 1`, so no switcher rendered in either case; a theme that reads either directly should check it.


fballiano:
@hirale Thank you so much for this! I made a last round of review, could you check if I didn't break anything? 🙏🙏

hirale:
@fballiano 

Thanks for the fixes, both are right.

One thing for later: **this PR gave most currency questions a single source of truth, but the old entry points are still open, and conversion never got one: `currencyConvert()`, `Currency::convert()/getRate()`, and raw `getCurrencyRates()` each read rates their own way (the pattern behind #1230).**  
The goal is every currency question gets exactly one answerer.

If the direction works for you, I can open a follow-up PR that reroutes the in-tree callers to the single source of truth, gives conversion one entry point, and deprecates the old ways.

fballiano:
@hirale every consolidation/singlesourceoftruth is always a good thing so yeah of course

I was trying to wrap my head around this problem and came up with this plan, what do you think?


>   Step 1: fix the rate at the source
> 
>   Everything else depends on this one.
> 
>   Mage_Directory_Model_Resource_Currency is the only code that touches directory_currency_rate, and it answers one question with three types: the integer 1 when the two codes match, a
>   string when the DECIMAL column comes back from MySQL or PostgreSQL, and false when no row exists. That is the reason #1230 is a 500, and the reason #1264 had to add a (float) cast in
>   five separate places with the same comment attached.
> 
>   Make getRate(), getAnyRate() and _getRatesByCode() all return ?float, cast at the point of the fetch. null means "there is no rate". It never means one.
> 
>   Two more things are wrong in the same file and are cheapest to fix while you are in there:
> 
>   getAnyRate() falls back to a reverse lookup that computes 1/rate in SQL. A stored zero divides by zero. That value is reachable: saveRates() rejects a rate that compares equal to zero,
>   but a rate below the column precision passes that check and then rounds to zero on insert.
> 
>   Nothing ever clears the static $_rateCache. After the admin saves rates, or after the import cron writes them, the rest of that process keeps serving the old value. The cache key also
>   skips the strtoupper() that the query bind applies, so usd and USD get separate entries and one of them can go stale on its own.
> 
>   Step 2: carry the type up through the model
> 
>   Mage_Directory_Model_Currency::getRate() and getAnyRate() become ?float, and the getRates() memo stores floats.
> 
>   While there: convert() throws a bare \Exception when the rate is missing. No Mage_Core_Exception handler catches that, so a missing rate surfaces as a white screen instead of a message.
>   Use Mage::throwException().
> 
>   Step 3: give conversion its answerer
> 
>   Two methods on Mage_Directory_Helper_Data:
> 
>   public function getRate(string $from, string $to): ?float;
>   public function convert(float $amount, string $from, string $to): ?float;
> 
>   Both currencies are named by the caller. There is no hidden default, and null is the honest answer when no rate exists.
> 
>   The helper's own $_currencyCache goes away. It is the third cache over the same table, after the model's getRates() memo and the resource's static array, and three caches over one table
>   is how the answers drift apart in the first place.
> 
>   currencyConvert() stays, marked #[\Deprecated], delegating, and keeping its hidden default of "the current display currency" so nothing changes underneath existing callers.
> 
>   Step 4: stop guessing a rate of 1
> 
>   This is the part that costs merchants money today.
> 
>   Six places convert a price into a website's currency when the admin saves a product, or when the price index rebuilds. Each one checks first that the website currency actually differs
>   from the base currency. Then, if the rate turns out to be missing, each one does this:
> 
>   if (!$rate) {
>       $rate = 1;
>   }
> 
>   So the base amount gets written to the database, unconverted, labelled as a different currency. A product priced at 100 USD is stored as 100 GBP. Nothing warns anybody.
> 
>   - Catalog/Model/Product/Attribute/Backend/Price.php:76
>   - Catalog/Model/Product/Attribute/Backend/Groupprice/Abstract.php:62
>   - Catalog/Model/Resource/Product/Option.php:114
>   - Catalog/Model/Resource/Product/Option/Value.php:92
>   - Catalog/Model/Resource/Product/Indexer/Price.php:553
>   - Downloadable/Model/Resource/Link.php:116
> 
>   The correct behaviour is to skip the write. A missing price is visibly wrong and a merchant fixes it. A confidently wrong price is invisible and ships orders. The tree already knows
>   this: Mage/Weee/Model/Attribute/Backend/Weee/Tax.php:80 drops the row rather than guess. Copy that, and add an admin warning on the save path so skipping is not silent either.
> 
>   Two more sites in the same family, with different shapes:
> 
>   CatalogSearch/Model/Advanced.php:140 uses the rate without checking it at all. Today a missing rate is false, so the price filter multiplies its bounds by zero and the advanced search
>   quietly answers on price >= 0. It needs the null check and a message to the shopper.
> 
>   Catalog/Model/Layer/Filter/Price.php:404 has the same $rate = 1 line, but it is display scope. After #1264 getCurrentCurrencyRate() can no longer return an unusable rate, so this one
>   just routes through the store and the fallback goes away.
> 
>   Step 5: retire the hand-rolled readers
> 
>   Mage/Usa/Model/Shipping/Carrier/Dhl/International.php:1003 looks up the rate, and when that fails, runs a second query the other way round and divides. That is getAnyRate(),
>   reimplemented inline. One helper call replaces it.
> 
>   CatalogSearch/Block/Advanced/Form::getAvailableCurrencies() and Adminhtml/Block/Sales/Order/Create/Data::getAvailableCurrencies() both rebuild getServeableCurrencyRates() by hand, which
>   is exactly what #1264 introduced. Each carries a bug that disappears with it: the first reads its memo as _currencies but writes it as currencies, so the memo never hits and the query
>   runs on every call; the second reads the currency codes from $this->getStore() but the base currency from Mage::app()->getStore(), which in the admin area is store 0, so the two can
>   describe different scopes.
> 
>   The rest are small: Adminhtml/Block/Widget/Grid/Column/Filter/Price::_getRate() loads a currency model on every call and should use the helper; Maho/Giftcard/Model/Giftcard.php:155 moves
>   from currencyConvert() to convert(); and the two admin search grids (Bundle/.../Option/Search/Grid.php:121, Adminhtml/.../Order/Create/Search/Grid.php:126) are display scope and should
>   go through the store.
> 
>   Step 6: pin the historical rule down with a test
> 
>   The quote stamps store_to_base_rate, store_to_quote_rate, base_to_global_rate and base_to_quote_rate when the order is placed. Nothing recomputes them, and nothing proves that either. A
>   comment does not survive a refactor; a test does.
> 
>   Place an order, change the rate table, reload the order, assert that the totals and the stamped rates have not moved.
> 
>   Step 7: deprecate, do not remove
> 
>   #[\Deprecated] on Mage_Directory_Helper_Data::currencyConvert() and on Mage_Directory_Model_Currency::convert(), getRate() and getAnyRate(). They all keep working as delegates.
>   Third-party modules call them, and Maho is what those modules extend.

hirale:
@fballiano 
The plan looks right. The `?float` return types worth a release-note mention too, since they can break third-party subclasses that override these methods without a return type.

I’ll keep it as one PR with one commit per step. open it soon.

PR #1268, opened 2026-08-13
Fixed admin form ajax validation errors redirecting to the dashboard

### LINKED ISSUE #1267: Saving products with existing / duplicate SKU causes 302 to Dashboard.
In OpenMage, it used to return:

from admin/catalog_product/validate/ via ajax

> {"error":true,"attribute":"sku","message":"The value of attribute &quot;SKU &quot;must be unique"}

and you'd get a small warning under the SKU field. In Maho 26.07, we're not getting this, rather getting kicked back to the Dashboard with no messages. 

admin/catalog_product/validate/id/1044/key/ - will return an empty 200 and then we get a 302 redirect.

I've had a bit of a look at it today but can't locate the issue and wanted to log it here before I move on since I'm out of time right now.

fballiano:
Fixes #1267.

`mahoFetch()` treats a JSON `{ error: true }` payload as a failure and throws, so `varienForm._validate()` always landed in its `.catch()` and called `_processFailure()`, which sets `location.href = BASE_URL`. `catalog_product/validate` reports a duplicate SKU exactly that way, so saving such a product dropped the form and the message and left the admin on the dashboard. The `if (response.error)` branches in `form.js` and in the product edit template were dead code.

Added a `rejectOnError` option to `mahoFetch()`, default `true` so no other caller changes, and opted `_validate()` out of it. This also covers customer edit, system variable, system design, the product mass-attribute update and widget instance.

The two product attribute validation endpoints needed a second fix to benefit: they wrote the encoded payload with `setBody()`, leaving the content type as text/html, so `mahoFetch()` returned a string and reading `.error` off it gave `undefined`. Both now use `setBodyJson()`.

Broken since #319 removed prototypejs, whose `Ajax.Request` passed the payload to `onComplete`.

New browser test saves a product with a taken SKU and asserts the inline advice shows and the browser stays on the form.

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Catalog/Product/Action/AttributeController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Catalog/Product/Action/AttributeController.php
@@ -269,6 +269,6 @@ class Mage_Adminhtml_Catalog_Product_Action_AttributeController extends Mage_Adm
             $response->setMessage($this->getLayout()->getMessagesBlock()->getGroupedHtml());
         }
 
-        $this->getResponse()->setBody($response->toJson());
+        $this->getResponse()->setBodyJson($response);
     }
 }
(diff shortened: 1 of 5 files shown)


mageaustralia:
Hi @fballiano  - thanks for the quick response as always. I've been meaning to test but have been under the pump and not had an opportunity yet. I'll let you know when I can.

PR #1277, opened 2026-08-15
Updated TipTap to 3.30.1 and fixed the drag handle appearing over the admin header

### LINKED ISSUE #1276: Bug Adminhtml: TipTap drag handle breaks out on CMS page edit
## Description

When playing around with the TipTap editor in adminhtml CMS pages, I found an optical bug which I'd like to report.

## Steps to reproduce

* Open Adminhtml (also happens in the demo)
* Click on `CMS > Pages`
* Open any page
* Click on tab `Content`
* The TipTap drag handle appears in the header

Image

fballiano:
Fixes #1276.

**Drag handle position.** DragHandle appends a `position: absolute; top: 0; left: 0` wrapper into `editor.view.dom.parentElement`, which is `.tiptap-content`. That element was not positioned, so the wrapper anchored to the page and landed over the admin header. Added `position: relative`. It is also the scroll container, so the handle now stays correct while the content scrolls. Fullscreen mode never showed the bug because `.tiptap-fullscreen` is `position: fixed`.

The extension also only positions and reveals the handle on the first `mousemove`, so before that it rendered visible and unpositioned. The handle now starts `visibility: hidden`; the extension's own `showHandle()` clears that inline value on first hover.

**TipTap 3.29.2 → 3.30.1.** No code changes were needed. The new `addDecorations()` hook in 3.30.0 only replaces hand-written decoration plugins, and our single ProseMirror plugin uses `appendTransaction`. StarterKit now pins its bundled `@tiptap/*` versions exactly, and 3.30.0 fixes `insertContent`/`setContent` when prosemirror-model loads twice, which applies here because we import `@tiptap/pm/state` next to StarterKit. One behavior change to know about: `ListKeymap` now binds `Tab`, so Tab at the start of a paragraph after a list moves that paragraph into the list.

Verified in Chrome against 3.30.1: before, the handle sat at `top=3 left=0` and visible; after, it is hidden on load and hover still reveals it beside the correct paragraph.

> [!NOTE]

### DIFF
--- a/public/js/mage/adminhtml/wysiwyg/tiptap/extensions.js
+++ b/public/js/mage/adminhtml/wysiwyg/tiptap/extensions.js
@@ -1,13 +1,13 @@
 // SPDX-FileCopyrightText: 2025-2026 Maho <https://mahocommerce.com>
 // SPDX-License-Identifier: AFL-3.0
 
-import { Editor, Node, Mark, Extension, mergeAttributes, ResizableNodeView } from 'https://esm.sh/@tiptap/core@3.29.2';
-import StarterKit from 'https://esm.sh/@tiptap/starter-kit@3.29.2';
-import Image from 'https://esm.sh/@tiptap/extension-image@3.29.2';
-import TextAlign from 'https://esm.sh/@tiptap/extension-text-align@3.29.2';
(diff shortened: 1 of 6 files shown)


fballiano:
@tmewes could you give this one a quick test?

tmewes:
@fballiano I tested the branch locally an can confirm that the issue is gone. Thank you! :)

PR #1285, opened 2026-08-17
Hid the checkout login and register links when customer accounts are disabled

### LINKED ISSUE #1284: Bug: Checkout contains links to login and register when customer accounts are diabled
## Description

When `customer/account/enabled_in_frontend` is set to `No`, the checkout still displays links to the login and register pages:

Image
Image

Related template:
`app/design/frontend/base/default/template/checkout/onestep.phtml`

fballiano:
The one-step checkout rendered the "Have an account? Login / Register" bar even when `customer/account/enabled_in_frontend` was `No`. It now uses the same flag check as `page/html/header.phtml`.

Every other account link in the theme was already guarded with `ifconfig="customer/account/enabled_in_frontend"`, so this template was the only leak.

Fixes #1284

> [!NOTE]


fballiano:
Thank you @tmewes!!

Issue #1269, opened 2026-08-13
Website prices are seeded once from a rate that may not exist, and never refreshed


hirale:
Hi @fballiano, there are a few decisions to be made in step 4.

> https://github.com/MahoCommerce/maho/pull/1264#issuecomment-5278137738

With `catalog/price/scope = website`, two mechanisms determine what a product costs on a website whose base currency differs from the default-scope base currency. Neither has an answer for a missing rate.

- **Base price is converted once, at product creation, then frozen.**
  `Backend/Price.php:59-62` returns early once the attribute has original data, so the conversion below it runs only for a new product. Changing the rate afterwards does not affect existing products; they keep the value they were created with.
- **On read, the base price is not converted at all.** The index reads the attribute at the  website's default store and falls back to the default-scope row (`Indexer/Price/Default.php:230`), handing an amount in the default-scope base currency to a website that prices in another one.
- **Tier and group prices are converted on every reindex**, from the same rate table (`Indexer/Price.php:400,447`).

So base prices are merchant data seeded by a conversion, while tier and group prices are derived from a conversion. Both models are defensible; carrying both, with no rule for a missing rate, is the problem.

## Example

Default-scope base `USD`, website B prices in `EUR`, no `USD -> EUR` rate. Product created at store 0 with `price = 100` and a global tier price of `90`.

| | website A (USD) | website B (EUR) |
| --- | --- | --- |
| base price | $100.00 | **€100.00**, a USD amount relabelled |
| global tier price | $90.00 | **€90.00**, recomputed as `90 * 1` on every reindex |
| tier price website B set itself | n/a | €92.00, correct, no rate involved |

Two reasons this is easy to miss:

- #1264's guard does not cover it. That one is about display currencies; this rate is between the default-scope base currency (`App.php:1141`) and a website's base currency. On website B the base currency is EUR, EUR against itself is 1.0, so `Store::getServeableCurrencyRates()` is satisfied.
- Importing the rate later fixes nothing, because the seeding never runs again.

## How other platforms avoid this

At the architecture level, everyone either guarantees the rate exists or does not convert at all. Nobody converts and falls back to 1.

| Platform | Where a price lives | Who converts | A missing rate |
| --- | --- | --- | --- |
| OpenMage / Magento 2 | per website, seeded once by conversion | product save (base), reindex (tier, group) | converts at 1:1, silently |
[...]

PR #1294, opened 2026-08-17
Centralized currency rate lookups and prevent incorrect price writes

hirale:
Follow-up to #1264, implementing the seven-step plan from that PR.

## 1. Make rate lookups return one type

`Mage_Directory_Model_Resource_Currency` answered one question with three types: `1` for matching currencies, a string when the DB returned `DECIMAL`, `false` when no row existed. Rate methods now return `?float`, where `null` means there is no rate. Callers lose their repeated `(float)` casts.

### Other fixes

* Currency codes are normalized through `Mage_Directory_Helper_Data::normalizeCurrencyCode()`, including the static cache keys and duplicate pair spellings before upsert.
* Rate caches are cleared after `saveRates()` and imports.
* `getAnyRate()` no longer performs `1 / rate` in SQL.
* `saveRates()` rejects values that cannot fit the DB column.
* Admin rate matrix handles invalid, zero and whitespace-padded values, and reads a cleared cell in the admin's own decimal notation.
* Currency rate imports write through `saveRates()`, and an import that fetched nothing stays a no-op.
* `directory_currency_rates_save_after` is dispatched only when a save stored rows.
* The group price backend drops its own rate memo; the resource cache below it already has one.

Fixes: `49f54dc83`, `eeaa769a8`, `0368ae469`, `b84d2235c`, `c22990c47`, `edccdd003`, `5fcc90213`, `8a562d4a9`, `1f873b336`, `cbddec3a8`

## 2. Keep the currency model consistent

Model rate methods take the same `?float`. `getRates()` no longer overwrites a rate the caller provided, and `convert()` throws `Mage_Core_Exception` instead of a bare `Exception`, so existing handlers catch it.

Fixes: `772244b35`, `fde57cc8c`, `067d1cd94`

## 3. Add one answerer for currency-pair conversion

```php
Mage::helper('directory')->getRate(string $from, string $to): ?float
Mage::helper('directory')->getAnyRate(string $from, string $to): ?float
Mage::helper('directory')->convert(float $amount, string $from, string $to): ?float
```

Both currencies are explicit at the call site; there is no hidden default. The helper's duplicate `$_currencyCache` is gone, leaving one cache at the resource layer.

Callers that must continue after a miss use `getRateOrWarn($from, $to, $subject)` or `warnMissingRate($from, $to, $subject)`, which report once per `from/to/subject` on the canonical spelling. Admin messages are added only when an admin session already exists, so CLI processes do not create one, and a blank rate cell is treated as an intentional "no rate".

Fixes: `b071d1b6f`, `e1be3bfa2`, `45a8433d5`, `314afa314`

## 4. Stop using `1` for missing rates

The most important behavioral change. Product and indexer paths did `$rate = 1` when no rate existed, storing 100 USD as 100 GBP under the other currency's label. Those write paths now **skip the derived website price**; the indexer instead writes a `null` rate, so the website row stays in the index while the derived price is treated as unavailable. Advanced search and layered navigation no longer fall back to `1` either.

Skipping a write does **not** delete an existing wrong price, and scope resolution can still fall back to the default price. This PR only stops creating new derived values. #1269 covers the long-term behavior, where website prices are derived from the global price at read time; that work lands as its own PR.

Fixes: `a3443c85d`, `314afa314`, `81f2df530`, `a62581c31`, `95444c8d8`
Tests: `4e0d8d6ee`, `146b8fa51`

`a3443c85d` stands alone and can be dropped if the guard should land with the derivation work instead.

## 5. Remove hand-written rate lookups

Callers that resolved rates on their own now use the centralized APIs.

* **Shipping**: DHL and UPS drop their own lookup logic. A method that cannot be converted surfaces the reason instead of vanishing, and the carrier's error takes precedence over the generic `specificerrmsg`.
* **Catalog search**: `getAvailableCurrencies()` uses the store's serveable currencies, and its cache-key typo is fixed.
* **Admin order creation**: available currencies come from the order's store, including that store's base currency. Deliberately different from the storefront switcher.
* **Price filter**: no currency model per lookup, and an unconvertible display currency is recorded rather than treated as parity.
* **Gift cards**: rate rows are stored `base → allowed`, so `allowed → base` divides by the stored rate. The GraphQL balance query probes the same direction the conversion uses, so it no longer reports a card in its own currency while the cart reports it converted.
* **Credit memos**: an adjustment on an order whose stamped rate is `null` still records the base amount, stores `null` rather than `0` in the order currency, and reports the miss. An operator sees the message; a gateway refund notification does not, having no session and no use for a rejection it would retry forever.

Fixes: `58036280f`, `26b3b4be9`, `2e913979a`, `5a0adf9e3`, `d0c5fa645`, `91cabee0a`, `c7c54bd1b`, `8696e0822`, `0ad702d15`

## 6. Treat order rates as historical data

Orders stamp their rates at placement and invoices and credit memos inherit them, so a later change to the rate table must not move them: an order placed at 1.2 stays at 1.2 after the table goes to 1.0. Missing rates are stored as `null`, not `0`. `Mage_Sales_Model_Quote::setForcedCurrency()` now rejects a currency that cannot be priced, so the failure happens when the currency is selected rather than during invoicing.

Test: `1c1bc85b3`

## 7. Deprecate legacy APIs and enforce the architecture

`currencyConvert()` and `Mage_Directory_Model_Currency::getRate()`, `getAnyRate()` and `convert()` stay for third-party compatibility as deprecated delegates, and `phpstan-deprecation-rules` keeps new callers off them. An architecture test pins the rest: only the resource model may touch `directory_currency_rate`, and only the currency model may touch the resource.

Fixes: `51a77d1b3`, `426519146`, `b5f5a1d54`, `11fb22d8d`

# Module devs

| Need                         | Use                                                              | Avoid                                           |
| ---------------------------- | ---------------------------------------------------------------- | ----------------------------------------------- |
| Rate from A to B             | `Mage::helper('directory')->getRate($from, $to)`                 | Currency model / rate table                     |
| Rate in either direction     | `Mage::helper('directory')->getAnyRate($from, $to)`              | Currency model / rate table                     |
| Convert an amount            | `Mage::helper('directory')->convert($amount, $from, $to)`        | `currencyConvert()`, currency model `convert()` |
| Rate required for pricing    | `Mage::helper('directory')->getRateOrWarn($from, $to, $subject)` | Fallback `1`                                    |
| Currencies a store can serve | `$store->getServeableCurrencyRates()`                            | Rebuilding the list from the rate table         |
| Store's current currency     | `$store->getCurrentCurrencyCode()`                               | Session / `currency/options/default`            |
| Price in store currency      | `$store->convertPrice()`                                         | Manual conversion                               |
| Order currency               | `order_currency_code` + `base_to_order_rate`                     | Current store currency                          |

Legacy APIs still work, but are deprecated. The first two rules are enforced by the architecture test.

# Compatibility

* Resource and currency-model rate methods now return `?float`. Third-party overrides without a return type may fail on upgrade.
* `Mage_Directory_Model_Resource_Currency::saveRates()` now returns a row count and declares `: int`. Same caveat.
* `Mage_Sales_Model_Quote::setForcedCurrency()` now throws when the quote cannot be priced in the requested currency.
* UPS no longer uses `$_baseCurrencyRate`; a subclass assigning it creates an unused dynamic property.
* On PHP 8.4+, deprecated APIs emit `E_USER_DEPRECATED`.
* Missing rates during save or reindex no longer create derived website prices. A warning is logged, and shown in admin where applicable.
* Credit memo adjustments on an order with a `null` stamped rate store `null` in the order currency instead of `0`. The base amount is unchanged and the refund still goes through. Orders with previously stored zero rates are unaffected.
* Advanced search reports an unconvertible currency instead of silently searching with `price >= 0`.
* `directory_currency_rates_save_after` no longer fires for a save that stored no rows.
* Saving currency rates clears the in-process rate memos but not the price index, as before, so a product page reads the new rate while an index-served listing reads the old one. #1269 covers index invalidation.

> [!NOTE]


fballiano:
Thank you @hirale, amazing!

I'll wait for green and merge!

PR #1296, opened 2026-08-17
Fixed the singleton registry error that broke bootstrap when the OpenTelemetry module is active

hirale:
With `Maho_OpenTelemetry` active and the Configuration cache type disabled, every request after
the first one following a cache flush dies during bootstrap:

```
Mage registry key _singleton/core/resource already exists
#1 app/Mage.php(476): Mage::register('_singleton/core...', Object(Mage_Core_Model_Resource))
#2 app/code/core/Mage/Core/Model/Resource/Setup.php(127): Mage::getSingleton('core/resource')
#3 app/code/core/Mage/Core/Model/Resource/Setup.php(260): Mage_Core_Model_Resource_Setup->__construct('core_setup', 'Mage_Core')
#4 app/code/core/Mage/Core/Model/App.php(452): Mage_Core_Model_Resource_Setup::applyAllUpdates()
#5 app/code/core/Mage/Core/Model/App.php(328): Mage_Core_Model_App->_initModules()
#6 app/Mage.php(699): Mage_Core_Model_App->run(Array)
```

### Environment

| | |
| --- | --- |
| Maho | `963fad89d`, clean checkout, no local or community modules |
| PHP | 8.5.7 |
| Database | MySQL 8.4 |
| `open-telemetry/sdk` | 1.15.0, installed |

### Reproduce

Start from a working install of `963fad89d` with `Maho_OpenTelemetry` left at its shipped
`<active>true</active>`, tracing switched off (no `dev/opentelemetry/enabled`, no `OTEL_*`
variables) and nothing else customized.

1. Converge the declarative schema, so `isSchemaUpdatePending()` returns false. While it returns
   true, the 503 schema guard refuses requests before this bug can surface.

   ```
   ./maho migrate
   ```

2. Turn off the Configuration cache type, in System > Cache Management or directly:

   ```sql
   UPDATE core_cache_option SET value = 0 WHERE code = 'config';
   ```

3. Flush, so both the config cache and the cached schema verdict start cold.

   ```
   ./maho cache:flush
   ```

4. Request the same page twice, with a cache buster so no FPC or edge cache can answer.

   ```
   curl -s "https://example.test/?nocache=1" | head -3
   curl -s "https://example.test/?nocache=2" | head -3
   ```

Request 1 renders normally. Request 2 and every request after it dies with the trace above, until
the Configuration cache type is turned back on. Measured 1 pass then 3 failures over the 4
requests following a flush, then 5 failures out of 5 on a further round.

Read the response body, not the status code. A health check on `%{http_code}` alone reports the
site as fine: depending on the error display settings the failure is either 200 with the trace
inline, or 503 with a generic page whose real exception is in `var/report/`.

### Cause

`Mage::getSingleton()` builds the model first and registers it second, with nothing guarding the
window between. `Mage_Core_Model_Config::getModelInstance()` profiles the construction and
`\Maho\Profiler::start()` asks for the tracer before the object exists, so building
`core/resource` reaches back into `core/resource`:

```
Resource_Setup::applyAllUpdates()               App.php:452, sets update mode
  Mage::getSingleton('core/resource')           Setup.php:127, outer, key not registered yet
    \Maho\Profiler::start(...)                  Config.php:1407, before `new $className()`
      Mage::getTracer()                         Profiler.php:100
        Mage::helper('opentelemetry')           Tracer.php:144, the same window on helper()
        $helper->isEnabled() -> store config    Tracer.php:152
          Mage::getSingleton('core/resource')   inner, registers the key outer register() -> "already exists"
```

`Mage::$_tracerInitializing` only guards recursion back into `getTracer()` itself, not re-entry
into `getSingleton()`. The inner read runs long before the tracer can find out it is switched off,
because `isEnabled()` reads store config before `initialize()` reaches the
`class_exists(TracerProvider::class)` guard, and update mode is what lets that read reach the
database at all: it makes `getStore()` fall back to the default store rather than throw.

Two conditions have to line up, which is why the issue is normally hidden. First,
`getCacheSaveLock()` must return at Config.php:464 without registering `core/resource` at
line 468. This happens only when the Configuration cache type is disabled. Second,
`isSchemaUpdatePending()` must use its cached verdict rather than opening its own connection
at App.php:519. That is the state from the second request after a cache flush onward.

Although both conditions also apply to the CLI, the CLI does not hit the problematic path:
`Mage::app()` initializes through `Mage_Core_Model_Config::init()` and never calls
`_initModules()`, so schema updates are not applied during CLI bootstrap.

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -473,7 +473,7 @@ final class Mage
     {
         $registryKey = "_singleton/$modelAlias";
         if (!isset(self::$_registry[$registryKey])) {
-            self::register($registryKey, self::getModel($modelAlias, $arguments));
+            self::register($registryKey, self::getModel($modelAlias, $arguments), $arguments === []);
         }
         return self::$_registry[$registryKey];
     }
@@ -525,7 +525,7 @@ final class Mage
     {
         $registryKey = "_resource_singleton/$modelAlias";
         if (!isset(self::$_registry[$registryKey])) {
-            self::register($registryKey, self::getResourceModel($modelAlias, $arguments));
+            self::register($registryKey, self::getResourceModel($modelAlias, $arguments), $arguments === []);
         }
         return self::$_registry[$registryKey];
     }
@@ -557,7 +557,7 @@ final class Mage
     {
         $registryKey = "_helper/$helperAlias";
         if (!isset(self::$_registry[$registryKey])) {
-            self::register($registryKey, self::getConfig()->getHelperInstance($helperAlias));
+            self::register($registryKey, self::getConfig()->getHelperInstance($helperAlias), true);
         }
         return self::$_registry[$registryKey];
     }
@@ -576,7 +576,7 @@ final class Mage
     {
         $registryKey = "_resource_helper/$moduleAlias";
         if (!isset(self::$_registry[$registryKey])) {
-            self::register($registryKey, self::getConfig()->getResourceHelperInstance($moduleAlias));
+            self::register($registryKey, self::getConfig()->getResourceHelperInstance($moduleAlias), true);
         }
         return self::$_registry[$registryKey];
     }
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -473,6 +473,8 @@ final class Mage
     {
         $registryKey = "_singleton/$modelAlias";
         if (!isset(self::$_registry[$registryKey])) {
+            // A constructor can re-enter its own alias and register it first; that
+            // instance wins. Arguments still fail loudly, the caller wants its own.
             self::register($registryKey, self::getModel($modelAlias, $arguments), $arguments === []);
         }
         return self::$_registry[$registryKey];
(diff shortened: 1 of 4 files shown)


fballiano:
@hirale thank you so much for this, the writeup was perfect, I could follow every step of the trace and confirm it in the code 🙏

I pushed a commit on top, let me know what you think:

`Profiler::start()` now checks `_spanPrefixes` before asking for the tracer. `CORE::create_object_of::Mage_Core_Model_Resource` does not match any prefix so it can never open a span, but it was still booting the tracer, and that is what triggers the store config read that re-enters `core/resource`. So this kills the trigger, and as a bonus it removes a `getTracer()` call from every single model instantiation. Added a test in `TracingSafetyTest` for it.

I kept your registry change of course, it is the right safety net: a timer that *does* match a prefix (`cache.` for example) can still fire inside a constructor and blow up the same way. I just added a line of comment on the `$arguments === []` flag because it is a bit cryptic otherwise.

One thing I did not do: moving `class_exists(TracerProvider::class)` to the top of `Tracer::initialize()`. It would save a db read on installs without the SDK, but we would lose the "SDK not installed" warning, so that deserves its own PR.

Does it look good to you?

hirale:
@fballiano LGTM, good job.

PR #1297, opened 2026-08-18
Fixed SegmentMatchingTest website fixtures

hirale:
`SegmentMatchingTest` assumed website IDs were always reused. When another test had already created website ID 2, the fixture created `website2` with a different ID but still inserted customers with `website_id = 2`, causing foreign key failures and duplicate-code errors.

The fix is to look up the website by its unique `code` instead of assuming its ID is 2.


fballiano:
Thank you @hirale!!

PR #1263, opened 2026-08-12
Product prices in the admin now show up to 4 decimals instead of rounding to 2

fballiano:
Fixes #1262.

The admin price form element (`Mage_Adminhtml_Block_Catalog_Product_Helper_Form_Price::getEscapedValue()`) and the custom option price renderer (`Option::getPriceValue()`) formatted stored values with `number_format($value, 2)`, while storage keeps 4 decimals (DECIMAL(12,4)). A stored `10.6557` displayed as `10.66`, and any save (even of an unrelated field) wrote the rounded value back.

Both now render through a shared `Mage_Adminhtml_Helper_Data::formatPriceForInput()`: full 4-decimal precision, trailing zeros trimmed, never fewer than 2 decimals. So `10.6557` renders as `10.6557` and `10.6600` still reads `10.66`.

Written test-first: the new Pest tests in `tests/Backend/Unit/Adminhtml/` fail against the unfixed code (verified: the element returned `10.66` for a stored `10.6557`).

> [!NOTE]
> **Developed with the help of AI.**<br>As part of our commitment to GenAI transparency, we flag pull requests produced with AI assistance alongside human work. As with every change in Maho, a maintainer reviews and validates it before merge, we never merge purely AI-generated changes. See the [GenAI transparency](https://github.com/MahoCommerce/maho#genai-transparency) section for details.

Issue #476, opened 2026-01-01
Deprecate and eventually remove Flat Catalog feature


fballiano:
## Summary

Deprecate the Flat Catalog feature (for both products and categories) in favor of modern database optimization strategies, with plan to remove entirely in a future major version.

## Background

Flat Catalog is a legacy Magento 1 optimization that denormalizes EAV (Entity-Attribute-Value) data into wide flat tables to improve query performance. This was necessary in the early 2000s when database performance was limited, but is no longer needed with modern database engines.

### 1. **Magento 2 Already Deprecated It**
- Officially deprecated in Magento 2.3+ (2019)
- Adobe states it "causes performance degradation and indexing issues"
- Replaced with ElasticSearch + optimized EAV queries
- Planned for complete removal in future releases

### 2. **Maintenance Burden**
- Complex reindexing logic
- Data synchronization bugs (see current SQLite type conversion fix)
- Doubles storage requirements
- Increases code surface area for bugs
- Confusing for developers ("why do I need to reindex?")

### 3. **Modern Databases Don't Need It**
- MySQL 8.0+/MariaDB 10.5+ handle EAV joins efficiently with proper indexing
- PostgreSQL (which Maho now supports!) excels at complex queries
- SSDs and modern hardware make the performance delta negligible
- Query planners are vastly better than in 2008

### Phase 1: Deprecation (2026 Q1-Q2)
- [x] Mark as deprecated in admin UI (warning when enabled)
- [x] Add deprecation notices in code
- [ ] Document in changelog and upgrade notes
- [ ] Create migration guide with performance alternatives
- [ ] Run benchmarks: EAV + cache vs Flat Catalog
- [ ] Document proper database indexing strategies
- [x] Default to disabled for new installations

### Phase 2: Removal (2027 or next major version)
- [ ] Remove flat catalog code entirely
- [ ] Remove reindexers
- [ ] Clean up database tables
- [ ] Simplify indexing system
- [ ] Update documentation

## Migration Path for Users

Documentation should include:

1. **How to disable flat catalog safely**
2. **Database optimization guide:**
   - Recommended indexes for EAV tables
   - Query optimization tips
   - PostgreSQL-specific optimizations
3. **Caching strategies:**
   - Built-in file/APC cache configuration
   - Optional Redis setup
   - Block caching best practices
4. **Performance monitoring:**
   - How to measure catalog query performance
   - Identifying slow queries
   - When to consider external search (ElasticSearch)

## Benefits of Removal

[...]

PR #1319, opened 2026-08-22
Made website prices follow the currency rate instead of storing a converted copy

### LINKED ISSUE #1269: Website prices are seeded once from a rate that may not exist, and never refreshed

hirale:
Closes #1269, implementing the four decisions agreed there. Follow-up to #1294.

## The problem

When a product was created, Maho wrote a copy of its price into every website that prices in another currency, converted at that day's rate. That copy never changed again, and in the database it looked exactly like a price the merchant had typed for that website. So the rate moved and prices did not, and a product created while the rate was missing sold at the EUR number under the GBP label.

## The rule now

> A price the merchant set for a website is used as is. Otherwise the price is the default-scope price times the current rate, computed when it is read. No rate, no price.

Nothing is written by the conversion any more. Tier and group prices already worked this way; now the base price, special price, msrp, custom option prices and downloadable link prices do too. Percentage prices (an 80% special price, a 10% option) are never converted. Derived values keep 4 decimals; 2-decimal rounding stays in the display layer (#1263, #1274).

### 1. Several websites, each with its own store view and base currency

Setup:

* Catalog price scope is **Website** (System > Configuration > Catalog > Price).
* Two websites, each with one store view: **EU** with base currency EUR, **UK** with base currency GBP. EUR is also the install's default base currency, so the default-scope price is in EUR.
* One currency rate, EUR to GBP: `0.85`.
* One product. In the admin, with the scope switcher on "Default Config", the merchant enters the price `10.6557`. That is the default-scope price, the one every website starts from.

In the tables, "default" is that default-scope row and "UK" is a row the merchant entered with the scope switcher on the UK store view, with "Use Default" unchecked.

**Before this PR**

| Situation | `price` rows stored | UK storefront shows | after the rate moves to `0.90` |
|---|---|---|---|
| Merchant never touched the UK price | default `10.6557`, UK `9.0573` (written by Maho on creation) | `9.0573` GBP | still `9.0573` |
| Merchant enters `12.00` at UK scope | default `10.6557`, UK `12.00` | `12.00` GBP | `12.00` |
| No rate when the product was created | default `10.6557` | `10.6557` GBP, the EUR number | still `10.6557`, even once a rate exists |

**After this PR**

| Situation | `price` rows stored | UK storefront shows | after the rate moves to `0.90` |
|---|---|---|---|
| Merchant never touched the UK price | default `10.6557` | `9.0573` GBP, computed on read | `9.5901` |
| Merchant enters `12.00` at UK scope | default `10.6557`, UK `12.00` | `12.00` GBP, as entered | `12.00` |
| No rate | default `10.6557` | nothing: not salable, out of the index | priced the moment a rate exists |

The second situation is identical before and after: a price the merchant enters at UK scope is stored and used exactly as before. What is gone is the UK row Maho wrote on its own in the first situation, which could not be told apart from a merchant's `12.00` and was frozen at the old rate. Saving the product at UK scope without changing the price writes no new row.

### 2. One website, several store views with different display currencies

Setup:

* **One** website with base currency USD.
* Two store views under it: the EU view allows only EUR, the UK view allows only GBP, as display currencies.
* Rates USD to EUR and USD to GBP.

There is one price, stored and charged in USD; each view converts it for display at its rate, and the order is placed in that display currency. Example 1 does not apply here: the conversion happens on the way to the screen, as it always has, rounded to 2 decimals.

| EU view, price `10.6557` USD | Before this PR | After this PR |
|---|---|---|
| USD to EUR rate `0.92` exists | `9.80` EUR | `9.80` EUR, unchanged |
| No USD to EUR rate | `10.6557` USD, although USD is not in the view's allowed currencies | the view refuses to serve until a rate exists or USD is allowed |

## What changed

1. **Seeding removed.** `Backend/Price::afterSave()`, the option resources and the link resource no longer write converted rows. They store what was typed, nothing else.
2. **Price index derives.** The fallback is `ROUND(default * cwd.rate, 4)`, for product prices and fixed custom option prices alike. A website with no rate gets no row in the website table every price indexer joins, so its whole catalog leaves the index in one place.
3. **Model derives.** `Product::getPriceAttributeValue()` converts on read and never writes the result into the product data, so `load()` then `save()` cannot turn a derived price into a website price. `getPrice()`, `getSpecialPrice()`, `getMsrp()`, the bundle and gift card price models, custom options, downloadable links and the REST/GraphQL read model all go through it. The REST write response derives `msrp` the same way, so a POST or PATCH and a following GET agree. Collection-loaded products remember which store their prices were loaded for, so a cron or CLI process does not price them against whatever store it has current.
4. **No rate means not for sale.** `isSalable()` is false, the product is out of the index, and a gift card returns no price instead of its default-currency amount list.
5. **Rates and currencies invalidate what they touch.** Saving a rate a website prices in marks the price index for reindexing and flags catalog rules for reapplication; changing a website's base currency is registered with the price indexer like the price scope setting, and flags the catalog rules as well, because a new base currency selects a different rate.
6. **Catalog rules discount the website price**, derived the same way, instead of the raw default value; a website with no rate gets no rule prices.
7. **A store sells only in a currency it is allowed to use and has a rate for.** `Store::getCurrentCurrency()` no longer falls back to a base currency the merchant did not allow. It keeps base as the first choice, takes the first allowed currency with a rate when base is not on the allow list, and refuses only when no allowed currency has a rate. Admin requests are exempt so the screen that fixes the configuration stays reachable.

## Rows the old seeding left behind

They cannot be told apart from prices a merchant typed, so they are kept and now count as merchant prices. The upgrade to `Mage_Catalog` 2.0.1 counts them and, if there are any, leaves an admin inbox notice naming `catalog:price:website-overrides`. That command lists every price row scoped to a store or website (product, option and link prices) and flags the ones equal to `default price × current rate`, which is what the seeding would have written. Deleting the flagged rows is the merchant's call.

## Out of scope

* `cost` is not a selling price and is not converted.
* Merchant-added price attributes are read through the magic getter, which cannot derive; use `$product->getPriceAttributeValue('<code>')` to convert one.
* Flat catalog (deprecated, #476) gets no derivation. This is a regression: flat used to copy the seeded rows and now shows the default-scope amount on a website in another currency.
* Configurable attribute price deltas and gift-card amount lists stay in the default currency, as on `main`.

## Behavior changes

* **`Store::getCurrentCurrency()` can throw** on the storefront when nothing in a store's allow list has a rate. Where base is off the allow list but another allowed currency has a rate, that currency is served instead of base. Such a store silently sold in base before, so the error only appears after upgrading.
* `Product::getPrice()` returns `float|null`; `null` means no usable rate.
* `getExistsStoreValueFlag()` is now correct for collection-loaded products and is set by a save at store scope without a reload; `unsExistsStoreValueFlag()` is new.
* `Mage_Catalog` moves to 2.0.1; its data upgrade raises the inbox notice described above and marks the price index for reindexing. `Mage_CatalogRule` moves to 2.0.1 and marks its own rules for reapplication. Both were built under the old rule. Prices on a website in another currency change on upgrade only for products that never got a converted row (none was written while the rate was missing); seeded rows stay as they are until the merchant deletes them.
* New command `./maho catalog:price:website-overrides`: lists every price row scoped to a store or website with its default value, the current rate and whether the row equals `default × rate`. `--suspect-only` shows only those, `--limit N` stops after N rows. Read-only.
* For extension code, `getData('price')` and `getPrice()` are no longer the same value: the first is the stored default-scope amount, the second is what this website charges. A price model that reads the attribute data directly bypasses the conversion.
* `./maho health-check` gains a scan of `app/code/local`, `app/code/community` and project themes for exactly that: `getData('price')`, a price model of its own, and direct reads of `catalog_product_entity_decimal`.

> [!NOTE]


fballiano:
Thank @hirale for taking care of this!!!

PR #1324, opened 2026-08-23
Added a store setting that makes 2FA mandatory for admin users

### LINKED ISSUE #1323: Feature: Enforce Admin 2FA
### Summary

Admin 2FA already exists in Maho, but enrolment is optional. For most stores, 2FA on admin accounts should be mandatory. This requests a store-level setting that requires all admin users to enrol in 2FA, with a configurable grace period (as Google and other platforms do) — admins are prompted to enrol from first sign-in and must complete it within the grace window, after which access is blocked until they do.

### Why

- **The admin panel is the highest-value target in the application.** A compromised admin account exposes every customer's data, all orders, and can inject store-wide malicious code — unlike a customer account, where the blast radius is one person. Optional 2FA leaves the most critical logins the most likely to be unprotected.
- **Optional security is inconsistently applied.** When 2FA is opt-in, a portion of admins never enrol — usually the busy or less technical ones, exactly the accounts most at risk. Enforcement closes that gap.
- **It's a baseline expectation and often a requirement.** Magento 2 has enforced admin 2FA since 2.4; PCI-DSS and many cyber-insurance policies require MFA on privileged accounts. A store currently can't guarantee all admins have 2FA on, even though the feature exists.
- **The feature is already built.** This is an enforcement layer on top of existing admin 2FA, not new 2FA work — so the lift is small.

### Proposed behaviour

- A store-level config setting: **"Require 2FA for all admin users"**, off by default.
- A companion **grace period** setting (e.g. days, or number of logins) for how long an admin may continue signing in before enrolment becomes blocking.
- When enforcement is on, a non-enrolled admin is prompted to enrol at each sign-in during the grace window, with the remaining time shown, but can still work.
- Once the grace period expires, the admin is redirected to enrolment on sign-in and blocked from the rest of the backend until enrolled.
- A grace period of `0` means enrol immediately (no grace) for stores that want hard enforcement.
- Off by default so existing installs are unaffected on upgrade.
- The existing 2FA recovery/reset path continues to apply, so enforcement can't permanently lock out a legitimate admin.

### Acceptance criteria

- [ ] A store-level setting exists to require 2FA for all admin users, defaulting to off.
- [ ] A configurable grace period setting exists, with `0` meaning immediate enforcement.
- [ ] During the grace window, a non-enrolled admin is prompted to enrol on sign-in but can still access the backend, with remaining time indicated.
- [ ] After the grace period expires, a non-enrolled admin is routed to enrolment and blocked from other backend pages until enrolled.
- [ ] Admins who already have 2FA configured are unaffected.
- [ ] With enforcement off, behaviour is unchanged (2FA stays optional).
- [ ] The existing 2FA recovery/reset mechanism still works under enforcement.
- [ ] Behaviour is covered by tests.

### Notes

The grace period mirrors how Google and similar platforms roll out mandatory 2FA — it lets store owners turn enforcement on without locking out active admins mid-session, giving everyone a defined window to enrol. Combined with off-by-default, it means no existing install changes behaviour unexpectedly on upgrade. Happy to help refine scope or test against a real store.

fballiano:
Closes #1323.

Admin 2FA was per-user opt-in, so a store could not guarantee that every backend account had it on. Two new fields in System > Configuration > Advanced > Admin > Security control it:

- **Require 2FA for All Admin Users**, off by default.
- **Require 2FA From**, a cutover date.

Before the date, a user without 2FA sees one reminder per session and keeps full access. From the date on, the user is sent to My Account and blocked from the rest of the backend until they enroll. An empty date blocks at once. Only the enrolment page, its POST handler, the passkey registration call, and logout stay reachable while blocked.

A registered passkey satisfies the requirement. `Mage_Admin_Model_User::authenticate()` already skips the TOTP check for a passkey sign-in, and a passkey is phishing-resistant, so forcing those users to add TOTP would be a step down.

### Why a fixed date instead of a per-user countdown

The issue asked for a grace period measured from first login. That needs a new column, an upgrade script, and rules for a user who never logs in during the rollout, who would then get a fresh full window months later. A fixed date has no such hole, needs no schema change, and shows the store owner the exact deadline.

### Recovery

`./maho admin:user:twofa-reset` clears the 2FA of a user who lost their device. No recovery path existed before, so enforcement without it could lock an owner out of their own store. `./maho config:set admin/security/twofa_required 0 --scope default --scope-id 0` remains the escape hatch.

### Related fix

The account save accepted `twofa_enabled=1` with an empty verification code. That stored a secret the user never scanned and locked them out at the next login, because `authenticate()` then started asking for a code. `Mage_Admin_Model_User::applyTwofaChange()` now demands a valid code, on the off-to-on transition only.

### Testing

13 new tests in `tests/Backend/Unit/Admin/`. `composer lint` passes. The full Backend suite passes locally (3434 tests). The manual walk-through in a live admin panel was not run.

> [!NOTE]


fballiano:
tested and it works correctly, I think the fixed date is an ok midground for functionality and simplicity

jparker1986:
Great work @fballiano. The idea of the countdown was to replicate what Google did, which was give each new user a grace period, but understand the constraints you have to work within. I personally feel a user who kicks the can down the road will likely do it until he is "forced" anyway, but it did give a bit of breathing space for them to get setup.

One note I would make is when I enabled it for my own account, the process was a bit fiddley because I scanned the QR code, entered the 2FA code returned by Authy and clicked "Save", only to be prompted for the admin password again after a page reload. 

I do view having to re-enter the current admin password as a security step to ensure people can not make changes to an account which is already logged in, but can be annoying at times when trying to make changes to API Users, etc. 

I do wonder if a better approach for UX would be to ask for that password prior to form submit, or getting rid of it completely.

PR #1346, opened 2026-08-28
Templates: Align whitespaces, use echo shortcut, remove unnecessary semicolons

tmewes:
## Description

This PR is just a small cleanup/fix of some annoying inconsistencies in the template files:
* Align whitespaces, e.g. replace `<?php if(...):?>` with `<?php if (...): ?>`
* Use `<?=` instead of `<?php echo`
* Remove unnecessary semicolons, e.g. replace `<?php endif; ?>` with `<?php endif ?>`


fballiano:
Thank you Tino!!

PR #1349, opened 2026-08-30
Fixed revocation emails failing once the config section is saved in the admin

claudiobianco:
`adminhtml/system_config_source_email_template` derives its option value from the
field's config path, so `revocation/email/received_template` offers
`revocation_email_received_template`. The templates are registered as
`revocation_email_received` / `revocation_email_merchant`, so saving Sales >
Revocation stores a code no template answers to and both mails die in
`sendTransactional()` with "Invalid transactional email code". The customer still
gets the success page, so the receipt the directive requires silently never
arrives.

Renamed both nodes to the path-derived codes (the convention the other 44 core
email template selects follow) and updated the defaults. Shops that already saved
the section need no migration: the stored value is the new node name.

The added test walks every such select and asserts the offered code is registered
and the config default resolves. Against unfixed `main` it fails with
`no template registered as "revocation_email_received_template" for config field
revocation/email/received_template`.

### DIFF
--- a/app/code/core/Maho/Revocation/etc/config.xml
+++ b/app/code/core/Maho/Revocation/etc/config.xml
@@ -44,16 +44,16 @@ SPDX-License-Identifier: OSL-3.0
         </resources>
         <template>
             <email>
-                <revocation_email_received translate="label" module="revocation">
+                <revocation_email_received_template translate="label" module="revocation">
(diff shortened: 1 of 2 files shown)


fballiano:
Thank you @claudiobianco!!

PR #1351, opened 2026-09-02
Fixed loadByUsername() leaving admin/API users partially loaded

hirale:
`Mage_Admin_Model_User::loadByUsername()` populated the model directly with `setData()`, bypassing the resource `_afterLoad()` and `setOrigData()` logic normally performed by `load()`.

This caused two issues on subsequent saves:

* **`extra` was JSON-encoded repeatedly.** Each `loadByUsername()` → `save()` added another JSON layer, causing `extra` to grow until MySQL rejected the write. String-shaped `extra` also broke System Config collapse state handling and could discard existing states.
* **The stored password hash was re-hashed.** Without orig data, `_beforeSave()` treated the password as changed and bcrypt-hashed the existing hash again. As a result, `admin:user:enable`, `admin:user:disable`, and `admin:user:twofa-reset` could lock the admin out. `admin:user:change-password` avoided the password issue but still corrupted `extra`.

`Mage_Api_Model_User::loadByUsername()` had the same orig-data problem, causing an existing `api_key` hash to be re-hashed when saved.

Both methods now delegate to `load()`, ensuring `_afterLoad()` and orig data are initialized correctly.

## Reproduce

Against `b14d40e81` without the patch:

```bash
./maho db:query "SELECT password, extra, LENGTH(extra) FROM admin_user WHERE username = 'admin'"
printf 'admin\n' | ./maho admin:user:disable
printf 'admin\n' | ./maho admin:user:enable
./maho db:query "SELECT password, extra, LENGTH(extra) FROM admin_user WHERE username = 'admin'"
```

The password hash changes and the old password no longer works; `extra` gains another JSON layer on each save.

### DIFF
--- a/app/code/core/Mage/Admin/Model/Resource/User.php
+++ b/app/code/core/Mage/Admin/Model/Resource/User.php
@@ -59,27 +59,6 @@ class Mage_Admin_Model_Resource_User extends Mage_Core_Model_Resource_Db_Abstrac
     }
 
     /**
-     * Load data by specified username
-     *
-     * @param string $username
-     * @return false|array
-     */
-    public function loadByUsername(#[\SensitiveParameter] $username)
-    {
-        $adapter = $this->_getReadAdapter();
-
-        $select = $adapter->select()
-            ->from($this->getMainTable())
-            ->where('username=:username');
-
-        $binds = [
-            'username' => $username,
-        ];
-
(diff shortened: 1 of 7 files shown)


fballiano:
Thank you @hirale 🙏🙏🙏

PR #1376, opened 2026-09-09
Hardened OAuth callback URL matching, escaped the consumer name on the authorize login page, and deprecated Mage_Oauth

fballiano:
- A consumer with a registered callback URL now only accepts a callback whose scheme, host and effective port match exactly, whose path continues the registered path at a segment boundary, and which carries no userinfo. The path check runs on the decoded path, so `%2e%2e` and backslash segments are refused too. Every registered query parameter must be present with the same value. Before, a plain string prefix check let hosts such as `legit.com.evil.com` or `legit.com@evil.com` pass, and any other well-formed URL fell through to a generic URL check, so the authorized token and verifier could be redirected to a foreign host.
- The check lives in `Mage_Oauth_Helper_Data::isCallbackUrlOnAllowlist()` and is used by both the initiate request and the request token save. The out-of-band marker short-circuits it. A consumer without a registered callback keeps the generic URL validation.
- The consumer model refuses a callback URL or rejected callback URL without a scheme or with userinfo at save time, so the admin sees the error instead of a silent lockout.
- The request token no longer instantiates `core/url_validator` through `getSingleton()`, which passed an array to a typed constructor and crashed. The class had no other caller and is removed.
- The admin OAuth authorize login templates now escape the consumer name, matching the sibling button templates.
- Every `Mage_Oauth` class is marked `@deprecated since 26.9` in favor of `Maho_ApiPlatform`, the same as `Mage_Api2`.
- Backend tests cover the bypass shapes, the token and server validation, the consumer validation, and the escaped login page.


fballiano:
**Shape.** The PR does two related things in the OAuth authorize flow. It replaces a string prefix match on the callback URL with a check on scheme, host, effective port and path. It also escapes the consumer name that the admin login page prints. The rule lives in `Mage_Oauth_Helper_Data::isCallbackUrlOnAllowlist()`. It runs from one place, `Mage_Oauth_Model_Server::_validateCallbackUrlParam()`, beside the other protocol parameter checks of that class. The shape fits the platform: no new module, no new configuration key, and one new public helper method. One cost remains. The model no longer checks the column on save, so a future writer of a token callback URL must call the helper itself.

**Fixed.**

- `app/code/core/Mage/Oauth/Model/Token.php:222` Deleted the callback URL check from `validate()`. `_beforeSave()` calls `validate()` on every save, not only on an insert. So the check ran again long after the token was issued. An admin changed the consumer callback URL, and then failed to revoke an older token. The throw also left the `foreach` in `AuthorizedTokensController::revokeAction()`, so every other selected token stayed active. `Mage_Oauth_Model_Server::_validateCallbackUrlParam()` already applies the rule on the only path that writes the column. 11 lines deleted, 0 added.
- `app/locale/en_US/Mage_Oauth.csv:105` Deleted the two strings that only the removed block used. This also turned the "Check CSV files are sorted alphabetically" job green. That job failed on the previous head.
- `tests/Backend/Unit/Oauth/Model/CallbackUrlValidationTest.php` Deleted the `Mage_Oauth_Model_Token::validate` group and the helper `oauthRequestTokenFor()`. Both tested the removed block.
- `tests/Backend/Integration/Oauth/AuthorizeLoginTemplateTest.php:31` The test rendered empty HTML. It failed on all seven database backends. It set the design area to `adminhtml`, but it left the design package on the frontend package `base`. Only the `default` package holds the admin templates. So `getTemplateFilename()` returned an empty string, and `fetchView()` logged "Template file not found". The test now sets the package through the existing `Mage_Core_Model_Design_Package::setAllGetOld()`. The platform already uses that method for this save and restore pair. I ran the test against sqlite: both datasets pass. I also took the `escapeHtml()` call out of the template, and both assertions failed again.

**Advice.**

- `app/code/core/Mage/Oauth/Helper/Data.php:117` Cut the four line docblock to one line. Its second sentence restates the `if` conditions below it.
- `app/code/core/Mage/Oauth/Helper/Data.php:135` Replace `hash_equals($registeredUrl, $callbackUrl)` with `===`. Neither value is a secret. Every other `hash_equals()` call in this module compares a verifier, a signature or a key, so the constant time compare misleads the reader.
- `app/code/core/Mage/Core/Model/Url/Validator.php` Delete the file. This PR removed its last caller. The model factory also cannot build it, because it passes an array to a `?string` constructor parameter. The fix reaches outside the files the PR touches.
- `tests/Backend/Integration/Oauth/` The removal of the check in `Token::validate()` has no test. Red: save a consumer with `callback_url` `https://legit.example/cb`. Create a request token on that URL. Change the consumer to `https://other.example/cb`. Load the token fresh, then assert `expect(fn() => Mage::getModel('oauth/token')->load($id)->setRevoked(1)->save())->not->toThrow(Mage_Core_Exception::class)`. The reload matters. `getConsumer()` caches the consumer on the token instance, so the first object hides the defect.

**Rejected.**

- CSV sort order, wrongly rejected. A verifier tested with shell `sort`, found the file already unsorted, and called the finding false. The `locale-csv-sorting` workflow sorts with `strnatcasecmp` instead, and it failed on the previous head. The finding was real. The round 1 deletion resolved it.
- `@param array<string, string|int>` disagrees with the `?int` return. `parse_url()` always returns an integer port, and AGENTS.md endorses an array shape docblock.
- The new method duplicates `Maho_ApiPlatform_Model_Oauth_Client::hasRedirectUri()`. Two protocols, two tables, two deliberate policies. That method matches an exact entry in the `redirect_uris` list of an OAuth 2.1 client.
- `getEffectivePort()` duplicates `Maho_AccessibilityScan_Helper_Data::getUrlAuthority()`. That method is protected, so the OAuth helper cannot call it. It also returns null for any scheme other than http and https, so it breaks a callback with a custom scheme.
- The test uses reflection on protected names. The public `initiateToken()` needs a signed request, a persisted consumer and a valid signature. 20 test files in this repository already use reflection.
- The model tests pass for the wrong reason. The old code fell back to `isValidUrl()`, which accepted `https://legit.com.evil.com/cb` silently. So the tests do turn red for the right reason.
- `expect(true)->toBeTrue()` asserts nothing. If the method throws, the bare calls already fail the test. The proposed replacement has the same blind spot, and nine other test files use the same shape.
- `setIsSimple(true)` on the block removes the need for the design change in the test. The verifier returned a rejected verdict, so the loop left the design change in place.

**Needs a decision.**

A consumer with an empty registered callback URL still accepts any valid URL. The field is optional on the admin consumer form. In that case `_validateCallbackUrlParam()` falls back to `Mage::helper('core')->isValidUrl()`. A holder of the consumer key and secret then sends `oauth_callback=https://evil.com/steal`. An admin authorizes, and the browser carries the token and the verifier to that host. That is the exact result the PR wants to stop. The behavior predates the PR, so the loop left it alone. When the consumer holds no registered callback URL, the fix refuses the initiate request. The fix also makes the field required. That fix changes behavior for every existing install.

The admin save path never checks `callback_url`. `Mage_Oauth_Model_Consumer::validate()` checks only the key length and the secret length. The form field carries no `validate-url` class. An admin saves `app.example.com/cb` with no scheme, or a URL that carries a user name. `parse_url()` then reports no host, so the new helper demands a byte identical callback. It refuses every real client callback with `parameter_rejected`. Before this PR the code fell back to "any valid URL". Such an install worked, and now it stops working, with no message that points at the consumer record. The fix belongs in `Mage_Oauth_Model_Consumer::validate()` or the admin form block. The PR touches neither file.

**Noticed, not fixed.**

- `app/code/core/Mage/Oauth/Model/Consumer.php:58` `_beforeSave()` calls `trim()` on `callback_url` and on `rejected_callback_url` without a null check. So saving a consumer without either value raises a deprecation, which is fatal in developer mode. I found this while reproducing the test failure.

**Checks.** php-cs-fixer, PHPStan level 6 and Rector are clean on every changed PHP file. php-cs-fixer is clean on the two changed templates. When they are passed by hand, PHPStan reports `method.internalClass` on the Pest test files. `.phpstan.dist.neon` scopes analysis to `app`, `lib`, `maho` and `public`. Six rounds. CI on the last commit: green, all 28 checks pass, including the seven database test suites (run 34689056784).

PR #1378, opened 2026-09-09
Stopped concurrent orders from going past coupon, promotion and download usage limits

fballiano:
Every usage counter was read into PHP, raised there, and written back. Two orders placed at the same time both passed the check in the cart, so a coupon went past `uses_per_coupon` or `uses_per_customer`.

Each counter now moves inside the database, in one `UPDATE` that refuses to go past the limit.

- Added `incrementTimesUsed()` and `decrementTimesUsed()` to the coupon, the rule, the coupon usage and the rule customer resources. Each one moves its counter in a single `UPDATE`. That `UPDATE` matches no row once the limit is reached, so the counter stops there.
- The order placement observer runs every counter in one transaction. It takes the rule rows in sorted id order, so two orders queue instead of deadlocking. The order itself still stands once a limit is reached. `Mage_SalesRule_Model_Validator` already refuses an exhausted coupon in the cart. `place()` also calls the payment before this observer runs, so a refusal here only destroys an order the shopper already paid for.
- The per-customer coupon counter and the per-customer rule counter first insert a row with a count of zero. That insert does nothing for a row that already exists. The guarded `UPDATE` then runs again, so a second order cannot count twice.
- `salesrule_customer` gains a unique key on `rule_id` and `customer_id`. Without that key the per-customer rule counter had to read the row before the insert. That read missed a row another order had just written.
- A downloadable link now takes one download from the purchased count before the file is sent. A shutdown callback gives the download back for any transfer that does not finish. The count used to move after the response, so parallel requests went past the number of downloads bought.
- The coupon model and the rule model are no longer saved to move a counter. `salesrule_coupon_save_*` and `salesrule_rule_save_*` therefore no longer fire on order placement. Nothing in the tree listens for them.

Every statement uses the DBAL adapter helpers, so the code runs on MySQL, PostgreSQL and SQLite.

## Release note

This release adds a unique key on `rule_id` and `customer_id` to `salesrule_customer`. Most installs hold no duplicate row and need no action.

An install that already lost this race holds two rows for one customer and one rule. The database refuses the unique key over those rows, so `./maho migrate` stops with a duplicate entry error. `Migrate` applies the declarative schema before every setup script, so no script cleans the rows first. Merge them by hand before the upgrade:

```sql
UPDATE salesrule_customer t
  JOIN (SELECT rule_id, customer_id, MIN(rule_customer_id) AS keep_id,
               SUM(times_used) AS total
          FROM salesrule_customer
         GROUP BY rule_id, customer_id HAVING COUNT(*) > 1) d
    ON t.rule_customer_id = d.keep_id
   SET t.times_used = d.total;

DELETE t FROM salesrule_customer t
  JOIN (SELECT rule_id, customer_id, MIN(rule_customer_id) AS keep_id
          FROM salesrule_customer
         GROUP BY rule_id, customer_id HAVING COUNT(*) > 1) d
    ON t.rule_id = d.rule_id AND t.customer_id = d.customer_id
 WHERE t.rule_customer_id <> d.keep_id;
```


fballiano:
## Review report

**Shape.** This pull request moves four usage counters out of read-then-write PHP code. The coupon, the promotion rule, and the purchased download link now move their counters with one guarded `UPDATE`. The shape fits the platform. Each counter moves inside the resource model that owns its table. The change adds no new module, no new event, and no new configuration key. Two problems stay open. Each one needs a change that is wider than this pull request. They are in "Needs a decision" below.

**Fixed.**

- `app/code/core/Mage/Downloadable/controllers/DownloadController.php:215-228`: the reservation now comes back from a shutdown callback. The release only ran in the `catch` block. A customer who loses the connection never reaches that block, so a dropped transfer cost a paid download for good. The callback also covers a script that reaches `max_execution_time`. Net 11 added lines.
- `app/code/core/Mage/Downloadable/controllers/DownloadController.php:225`: the download counts as sent only for a `connection_status()` of `CONNECTION_NORMAL`. A host that sets `ignore_user_abort = On` runs the output loop to the end after the customer leaves. The transfer then looked complete. Net 0 lines.
- `tests/Backend/Integration/Downloadable/Model/Resource/DownloadReservationTest.php:14-30`: the fixture creates a real order and a real order item. All three tests in this file failed on every database backend. `Mage_Downloadable_Model_Link_Purchased::_beforeSave()` refuses a null order id, and both columns carry a foreign key. Net 16 added lines.

I made two other changes and then reverted them. A later round proved that each one cost more than the defect it removed. Both problems are in "Needs a decision".

**Advice.**

- `app/code/core/Mage/Downloadable/Model/Resource/Link/Purchased/Item.php:56`: `releaseDownload()` writes `status = available`. Its only guard is `number_of_downloads_used > 0`. A refund that expires the link during the transfer is then undone, and the shopper gets one more download. Do not just drop the `status` write. `reserveDownload()` sets the status to `expired` on the last download. A release that skips the status then leaves an expired link with a free download. The fix must remember the status that the reservation replaced.
- `app/code/core/Mage/SalesRule/Model/Resource/Coupon/Usage.php:26`: `updateCustomerCouponTimesUsed()` has no caller left in the tree. A third-party module can still call it, so do not delete it. Mark it `@deprecated`, the way `Mage_Core_Helper_Data::ipRateLimiter()` is marked.
- `app/code/core/Mage/Downloadable/controllers/DownloadController.php:13`: `$_downloadSent` holds state for one action only. A local variable that the closure captures by reference is smaller. PHPStan then reports `Left side of && is always true`, because it cannot follow a by-reference write. The property is the form that passes the checks.
- `app/code/core/Mage/Downloadable/controllers/DownloadController.php:218`: no other call site registers a shutdown callback inline. The house pattern puts it in a small named method, with the reason in its docblock. See `Mage_Cron_Model_Observer::_registerJobShutdownHandler()`.
- `app/code/core/Mage/Downloadable/controllers/DownloadController.php:225`: the line reads the connection state, and nothing says why a plain `true` is wrong. One comment line carries it: `ignore_user_abort = On` lets the loop reach the end of the file after the customer leaves.
- `tests/Backend/Integration/SalesRule/Model/Resource/UsageLimitTest.php`: no test refuses an order placement through the coupon per-customer limit. The observer passes `(int) $coupon->getUsagePerCustomer()` to the resource, and no test checks that wiring.
  Red: build a coupon with `usage_limit = null` and `usage_per_customer = 1`. Dispatch `sales_order_place_after` once for the customer. Then assert on the second order:
  `expect(fn() => Mage::dispatchEvent('sales_order_place_after', ['order' => $second]))->toThrow(Mage_Core_Exception::class, 'You have reached the usage limit for this coupon code.');`
- `tests/Backend/Integration/SalesRule/Model/Resource/UsageLimitTest.php`: the same gap for the rule `uses_per_customer` limit.
  Red: build a rule with `uses_per_customer = 1` and a coupon with no limit. The second dispatch must throw `Mage_Core_Exception` with the text `You have reached the usage limit for this promotion.`, and `$rule->getTimesUsed()` must stay `1`.
- `tests/Backend/Integration/SalesRule/Model/Resource/UsageLimitTest.php:86,122,158,181`: four `toThrow()` calls name only the class. Any other `Mage_Core_Exception` makes them pass for the wrong reason. Add the expected text to each one.
- `tests/Backend/Integration/Downloadable/Model/Resource/DownloadReservationTest.php`: no test covers the `status = ?` condition of `reserveDownload()`. Delete that line, and every test stays green. A pending link then becomes downloadable.
  Red: build the item with `LINK_STATUS_PENDING`. Then assert `expect($resource->reserveDownload($itemId))->toBeFalse();` and `expect((int) $item->load($itemId)->getNumberOfDownloadsUsed())->toBe(0);`
- `tests/Backend/Integration/Downloadable/Model/Resource/DownloadReservationTest.php`: no test covers the `number_of_downloads_used > 0` guard of `releaseDownload()`. The column is unsigned, so a release with no reservation raises a range error.
  Red: on a fresh item, call `$resource->releaseDownload($itemId);` then assert `expect((int) $item->load($itemId)->getNumberOfDownloadsUsed())->toBe(0);`

**Rejected.**

- Delete `updateCustomerCouponTimesUsed()`, because it forwards with the limit switched off. The method never applied a per-customer limit before this pull request. The wrapper keeps the old contract exactly.
- The observer opens a transaction that gives no rollback, and its docblock is false. `Mage_Sales_Model_Service_Quote::submitOrder()` runs `place()` as a commit callback inside a real transaction. The docblock is correct, and the rule row lock lasts until the outer commit.
- The line `array_unique(array_filter(array_map(intval(...), ...)))` is too dense. The same pattern already sits at `app/code/core/Maho/CustomerSegmentation/Helper/Data.php:25`. The line is also correct for every input, including an empty `applied_rule_ids`.
- The old decrement path created a row with `times_used = 1` and blocked the customer. The decrement only runs from `sales_order_payment_cancel`. That event always follows an increment for the same pair, so the branch is unreachable.
- `forUpdate(true)` on the per-customer rule select fixes the stale read. It does nothing on SQLite, where the adapter drops the clause. It does nothing on PostgreSQL, where a select that returns no row locks no row. On MySQL it takes a gap lock on a table with no unique key. Two unrelated checkouts then deadlock on the insert that follows. I applied it in round 1 and reverted it in round 2.

**Needs a decision.**

1. The limit error lands after the payment.

   `Mage_Sales_Model_Order::place()` calls `_placePayment()` first. It dispatches `sales_order_place_after` next. The new `Mage::throwException()` calls therefore run after the gateway took the money. `Mage_Core_Model_Resource_Transaction::save()` rolls the order back, and nothing voids the charge. Trigger: a coupon with `usage_limit = 1` that two shoppers hold in their carts. The second shopper pays and then sees the usage limit error, with no order. Before this pull request that shopper got an order and lost no money.

   I moved the observer to `sales_order_place_before` in round 1 and reverted it in round 2. The move works, and a failed payment still rolls the counters back. The `UPDATE` on the shared `salesrule` row then holds its lock across the whole gateway round trip. A store with one cart rule for every order writes that single row on every checkout. Checkouts then queue at the speed of the gateway, not at the speed of the order save. An offline payment method adds no wait at all.

   Two repairs that I did not make. Move the observer to `sales_order_place_before` and accept the longer lock. Or split the work, so a read-only limit check runs before the payment and the counters still move after it. The second repair needs about 25 new lines, and it puts the limit in two places.

2. The per-customer rule counter can insert two rows.

   `Mage_SalesRule_Model_Resource_Rule_Customer::incrementTimesUsed()` runs a guarded `UPDATE`, then a plain select, then an insert. `salesrule_customer` has no unique key on `rule_id` and `customer_id`. MySQL runs at REPEATABLE READ. The transaction fixes its read view before it takes the rule row lock. The plain select can then miss a row that another placement just wrote. Two concurrent first uses by one customer then insert two rows. `loadByCustomerRule()` reads only the first row, and the customer passes `uses_per_customer`. The coupon path has no such hole, because `salesrule_coupon_usage` has a composite primary key for `insertIgnore()`.

   The repair that I did not make: add a unique key on `rule_id` and `customer_id` in a new upgrade script. Then delete the select branch and copy the `insertIgnore()` shape from `Mage_SalesRule_Model_Resource_Coupon_Usage::incrementCustomerTimesUsed()`. That is a schema change, which is wider than this pull request.

**Checks.** Four review rounds. Round 4 found no defect. `php-cs-fixer`, PHPStan and Rector report no error on the changed files. CI on `322dd1708` is green. All 28 checks pass, including the seven Pest suites on SQLite, MariaDB 10.11, MariaDB latest, MySQL 8.4, MySQL latest, PostgreSQL 15 and PostgreSQL latest. The three earlier SQLite failures came from the test fixture of this pull request, not from the review. The last two commits repair them.

PR #1379, opened 2026-09-09
Contained every filesystem path built from request or stored data

fballiano:
Adds `\Maho\Io::getPathWithinDir()`, one helper that resolves a path inside a base directory and refuses anything that leaves it (dot segments, backslashes, absolute paths, stream wrappers, null bytes, and symlinks that point out). Every place that builds a filesystem path from a request parameter or a stored value now goes through it.

- Custom option file downloads (storefront, wishlist, REST) resolve the stored quote and order paths inside their own target directories instead of a string prefix check; the same resolver guards the quote to order copy and the cleanup unlinks in `Mage_Sales_Model_Quote` and `Quote_Item`.
- The admin `{{media}}` directive, the WYSIWYG thumbnail resolver, `getCurrentPath()` and the image view, delete and edit actions contain their paths inside the media storage root. The editor preview accepts one `{{media}}` or `{{skin}}` directive and refuses every other directive before it runs.
- `\Maho\Io\File` now refuses to write, copy, move, delete or open a write stream on a destination outside the working directory passed to `open()`. A null byte or a `..` segment in the destination is refused even with no working directory open.
- The legacy REST and SOAP product image endpoints reduce a caller supplied file name to a bare name, and the image update path stays inside the product media folder.
- Downloadable link and sample files, admin customer file downloads, EAV file deletes, blog image deletes and media storage file info are contained the same way. A downloadable file name whose tmp path is not a file, or whose destination leaves the files directory, stops the product save.
- `php8`, `phar` and `phps` are added to the protected upload extensions.

Tests cover the helper (traversal, absolute, backslash, symlink and wrapper shapes), the working directory guard, the option path resolver, the media directive, the thumbnail resolver, the REST image file name and the extension list. The API suite test needs the installed test database, so it runs in CI.

> [!NOTE]


fballiano:
## Review report

**Shape.** The PR adds one helper, `\Maho\Io::containedPath()`, and sends about fifteen path builders through it, so several hand written `realpath()` and `str_starts_with()` pairs disappear from controllers and models. It also hardens `\Maho\Io\File` so that a write, copy, move, delete or write stream cannot leave the directory that `open()` received; every core caller of those methods writes inside its own working directory, so the guard is safe for core. The seam on `Mage_Catalog_Model_Product_Option_Type_File::resolveStoredPath()` fits, because the quote and order directories already live on that class and `Mage_Sales` already reached for that model. The shape problem is duplication, not behavior: `\Maho\Io` now offers three containment helpers with no deprecation, and the CMS media API keeps two hand written copies of the algorithm that the new helper contains. The PR also carries an upload extension change that its title does not mention.

**Fixed.**
- `Mage/Catalog/Model/Product/Attribute/Media/Api.php:216`: strip the leading separator before `containedPath()`. A gallery value is `/i/m/photo.jpg`, so the check read it as an absolute path and `catalog_product_attribute_media.update` always failed with "Can't create image." (1 line)
- `Mage/Eav/Model/Attribute/Data/File.php:210`: same strip. The stored value is `/f/i/name.pdf`, so the old customer file was never deleted from the disk. (1 line)
- `Mage/Wishlist/controllers/IndexController.php:797`: test the value with `is_array()` before `resolveStoredPath()`. A value that does not unserialize to an array raised a `TypeError`, which `catch (Exception)` does not catch, so the customer got HTTP 500. (2 lines)
- `Mage/Sales/Model/Quote.php:414` and `Mage/Sales/Model/Quote/Item.php:798`: build the option file model once, outside the delete loops, as `ProductCustomOptionProvider` and `DownloadController` already do. (2 lines)
- `tests/Api/V2/Write/ProductCustomOptionFileTest.php:57,100`: store the option value with `jsonEncode()`. `prepareForCart()` writes JSON, so `serialize()` is the legacy format. (2 lines)

**Advice.**
- `Mage/Catalog/Model/Api2/Product/Image/Rest.php:153` and `Mage/Catalog/Model/Product/Attribute/Media/Api.php:122` hold the same six line file name block. The single home is `lib/Maho/File/Uploader.php`, which this PR does not touch. See "Needs a decision".
- `Mage/Catalog/Model/Api2/Product/Image/Rest.php:156`: a file name of `///` makes `Mage_Core_Model_File_Uploader::getCorrectFileName()` emit "Undefined array key extension". The bug sits in `lib/Maho/File/Uploader.php:332` and predates the PR, but the new code is the first path that lets a caller reach it with a request value. The new unit test passes this exact input.
- `Mage/Downloadable/Helper/File.php:121`: `ltrim($file, DS)` keeps a leading backslash on Linux, while the three other call sites strip with `'\\/'`. The result is still fail safe, so this is an inconsistency only.
- `Mage/Downloadable/Helper/File.php:122`: the two line comment says every caller checks the result with `is_file()`. `_moveFileFromTmp()` does not.
- `Mage/Core/etc/config.xml:474`: the `php8`, `phar` and `phps` entries are upload validation, not path containment. A revert of the path work also removes the upload block.
- `tests/Backend/Unit/Cms/Model/MediaPathContainmentTest.php:53`: the case "refuses an absolute path outside the current directory" passes against the unfixed code too, because the old `resizeOnTheFly()` built a path that is not a file either. It proves nothing. Delete it; the traversal case above it is the real one.
- Missing tests, with the line that fails against the unfixed code:
  - `Mage_Cms_Helper_Wysiwyg_Images::getCurrentPath()`. Red: set the `node` parameter to a directory outside the storage root whose name contains the storage root text, then `expect(getCurrentPath())->toBe($storageRoot)`. The old test `stripos($path, $currentPath) !== false` accepts it.
  - `Mage_Downloadable_Helper_File::getFilePath()`. Red: `expect(getFilePath($linksDir, '../../../../app/etc/local.xml'))->toBe($linksDir . DS)`. The old code returns a real readable file that the download controller streams.
  - `Mage_Core_Model_File_Storage_Abstract::collectFileInfo()`. Red: `expect(fn() => collectFileInfo('../../app/etc/local.xml'))->toThrow(Mage_Core_Exception::class)`. The old code returns the content of that file.
  - `Maho_Blog_Model_Post_Attribute_Backend_Image::_deleteFile()`. Red: with the stored value `../keep.png`, `expect(is_file(Mage::getBaseDir('media') . '/keep.png'))->toBeTrue()`. The old code deletes it.
  - The two fixes above. Red: an image update through `catalog_product_attribute_media.update` returns the new content; and `compactValue(['delete' => 1])` on a file attribute leaves no file on the disk. Both fail against the PR before this review.

**Rejected.**
- `Mage_Cms_Model_Wysiwyg_Images_Storage::deleteDirectory()` keeps `substr()` after the removed prefix test. The only caller always passes an absolute resolved path that starts with the root.
- `Mage_Downloadable_Helper_File::getFilePath()` failure value plus `_moveFileFromTmp()`. The `mv()` guard that this PR adds already stops any write outside the base directory, and `rename()` cannot write onto an existing directory.
- `Mage_Core_Model_File_Storage_Abstract::collectFileInfo()` duplicate message. The smaller fix turns a handled exception into a fatal `TypeError`, because `getFilteredPath()` lives in a strict types file.
- The protected extension list is a configuration default only. The field is hidden in `system.xml` and no code writes it to the database.
- `ltrim($value, '\\/')` weakens the guard. It does not. Tested values such as `/../../app/etc/local.xml`, `//..//..//x` and `\..\..\app\etc\local.xml` still return false.
- The wishlist ternary should become `catch (Throwable)`. That would also hide a real error raised inside `_prepareDownloadResponse()`.
- `ltrim($file, DS)` in `Mage_Downloadable_Helper_File` gives a wrong result. A stored downloadable name never starts with a backslash, and `containedPath()` normalizes backslashes itself.

**Needs a decision.**
- The same file name cleaning block now exists in `Mage/Catalog/Model/Api2/Product/Image/Rest.php:153` and in `Mage/Catalog/Model/Product/Attribute/Media/Api.php:122`, and a narrower copy already sits in `Mage/Cms/Api/MediaProcessor.php:136`. The cost is three places to correct when the rule changes. The fix that I did not make: add one static method next to `getCorrectFileName()` in `lib/Maho/File/Uploader.php` that returns a safe base name with a default, then call it from all three sites. I left it because `lib/Maho/File/Uploader.php` is outside this PR, and `lib/Maho/Io.php` is the wrong home for a file name helper.

**Noticed, not fixed.**
- `Mage_Wishlist_IndexController::downloadCustomOptionAction():796` and `Mage_Sales_Model_Quote::_beforeDelete():420` decode the custom option value with the raw `unserialize()`. `prepareForCart()` writes that value with `jsonEncode()`, so the decode returns false for every value that the current code writes. The wishlist download and the quote file cleanup are both dead, and an uploaded file stays on the disk. `Mage_Sales_Model_Quote_Item::_beforeDelete()` already uses `Mage::helper('core/string')->unserialize()`, which accepts JSON. The behavior predates this PR.
- `Mage/Cms/Api/MediaProcessor.php:87` and `Mage/Cms/Api/MediaProvider.php:74` hold a hand written copy of the containment algorithm. They protect correctly today, but they do not use the new helper.
- `lib/Maho/Io.php` now holds three overlapping helpers, `allowedPath()`, `validatePath()` and `containedPath()`. `Mage_Core_Model_File_Validator_AvailablePath` and `Mage_ImportExport_Helper_Data` still point developers at the weakest one, `allowedPath()`.

**Checks.** Two rounds. `php-cs-fixer`, PHPStan level 6 and Rector pass on every changed file. CI on `19d3bfd52` is green: all 28 checks pass, including the seven database test suites.

fballiano:
One gap next to the `{{media}}` containment in this PR.

`directiveAction()` in `Mage_Adminhtml_Cms_WysiwygController` still passes the request parameter `___directive` straight to `filter()`. Every admin directive runs on that text, not only one `{{media}}` construction, and the stream-wrapper check after it is a denylist. The path containment added here closes the traversal case, but the endpoint still evaluates arbitrary directive text from the request.

Suggested change: before the filter call, match the decoded text against `\Maho\Filter\Template::CONSTRUCTION_PATTERN`, require exactly one match, and require its directive name to be `media`. Refuse anything else. That is the shape of the OpenMage fix for GHSA-7vww, and it removes the need for the stream-wrapper denylist.

PR #1381, opened 2026-09-09
Hardened security tokens and encrypted config paths

fballiano:
Security tokens now come from the CSPRNG instead of `md5(uniqid())`. This covers the shared `uniqHash()` helper (customer and admin password reset tokens, the magic login link, wishlist sharing codes, newsletter subscriber codes), the customer confirmation key, the legacy API session idand the customer segmentation coupon hash. All of them keep their previous shape, so the stored columns and the emailed links do not change.

The endpoints that consume an emailed token (customer account confirm, reset password, magic link login, admin reset password) now count a rejected token against a rate limiter. The customer endpoints share the existing `system/rate_limit/reset_password` budget (default 10 per hour per client), and the admin endpoint reads a new `system/rate_limit/admin_reset_password` budget with the same default. Both are fields under System > Rate limit, and 0 disables the limit. The ApiPlatform module repeated ten `system/rate_limit` defaults that other modules already declare, and its copy loaded last, so the effective limit for seven endpoints differed from the value the admin field documents. Those copies are gone. The two keys that only ApiPlatform held (`change_password`, `giftcard_balance`) moved to the Customer and Giftcard modules, and they, together with `coupon_validate`, now have an admin field. The gift card balance web endpoint reads the same key instead of two class constants. A valid token never counts against the client. The customer reset password token is now compared in constant time with `hash_equals`, as the admin and magic link checks already were.

The list of config paths stored encrypted was built from `system.xml` only, and it ignored `<config_path>` overrides. Paths declared encrypted in a `config.xml` default, website or store node (PayPal credentials, UPS and USPS account data) were therefore readable in plain text from a transactional email or newsletter template through `store.getConfig()`. The list is now built from the merged config tree as well, it resolves `config_path`, and it is cached per request. The same list also refuses encrypted paths in the admin variable allowlist that backs the `{{config}}` directive, and it makes the encryption key rotation command re-encrypt those values. The three USPS payment fields (account number, CRID, MID) were declared encrypted in `config.xml` but not in `system.xml`, so the admin form saved them in clear text; the `system.xml` fields now declare the encrypted backend model and the obscure frontend type. The encrypted backend model also takes the raw stored value when the form posts the mask: before, it decrypted the stored value first, which returned an empty string for a clear-text row and for a store scope row, so an untouched save stored the literal mask over the secret.

The admin login observer and the legacy API `login()` handler now accept only string credentials. A client that posted `login[username][]=x` reached the typed session and user model methods with an array, so the request ended with a `TypeError` instead of the normal failed login. The admin observer passes empty strings for any other shape, and the API handler returns `invalid_request_param`; it also accepts the username `0`, which the previous `empty()` check rejected. Ported from OpenMage/magento-lts#5723.


fballiano:
## Review

**Shape.** The PR replaces clock-derived tokens with `random_bytes()` across the customer, admin, API, newsletter and coupon generators, adds a shared rate limiter to the four endpoints that consume an emailed token, and rebuilds the encrypted config path list from the merged config tree instead of `system.xml` alone. The shape fits the platform: it reuses the existing `Mage::helper('core')->rateLimiter()` factory instead of writing a new limiter, and it keeps the encrypted path list in the one class that already owns config sections. Round 1 corrected the one behavioural error in the limiter work: the customer actions counted a hit even when the client was already blocked, so the 15 minute window never drained, and the customer controller now checks the limiter before the `try`, as the admin controller already did. Two open items stay with you: the rate limit budgets are class constants instead of `system/rate_limit/<key>` store config, and the encrypted path list lives in `Mage_Adminhtml_Model_Config`. After round 1 the branch is consistent in behaviour, and the remaining gap is test coverage, not code.

**Fixed** (commit f23684e01, 15 lines added, 12 deleted)

- `Mage/Customer/controllers/AccountController.php:638,863,1052`: the `tooManyAttempts()` check moved out of the `try` block in `confirmAction`, `resetPasswordAction` and `magicLinkLoginAction`. The `catch` calls `hit()`, so a blocked request added a new hit and the 15 minute window never drained. An attacker who filled the budget and then sent one request every 10 minutes kept a legitimate customer locked out for as long as the requests kept coming. Each action now shows "Too many attempts. Please try again later." on its own redirect, which also fixes the reset password action showing "Your password reset link has expired." to a blocked client.
- `Mage/Adminhtml/controllers/IndexController.php:468`: the limiter now throws the message that both callers already print. Both callers catch every exception and replace the text, so "Too many attempts. Please try again later." never reached an admin. The unused line is gone from `app/locale/en_US/Mage_Adminhtml.csv`.
- `tests/Backend/Unit/Core/Helper/SecurityTokenTest.php:47`: deleted the `Mage_Downloadable_Model_Observer::saveDownloadableOrderItem` entry. `->not` binds to `toMatch()` only, so `toContain('random_')` ran as a positive check, and that method body holds no `random_` call. The test failed on every run.

**Needs a decision**

1. Three USPS paths break the encryption key rotation. `Mage/Usa/etc/config.xml:147,149,150` mark `carriers/usps/payment_account_number`, `payment_crid` and `payment_mid` with the encrypted backend model, but the matching fields in `Mage/Usa/etc/system.xml:695,715,725` declare no backend model, so the admin saves those values in clear text. The new config tree scan adds the three paths to the list, and `sys:encryptionkey:regenerate` then fails to decrypt the clear rows and aborts with "No changes were made". `health-check` reports the same rows as undecryptable. Any store that filled a USPS payment field can no longer rotate its key. The fix is to add `<backend_model>adminhtml/system_config_backend_encrypted</backend_model>` to the three `system.xml` fields, or to drop the attribute from the three `config.xml` nodes. Both files sit outside this PR, so I left them alone.

2. The rate limit budgets are class constants. AGENTS.md says a public endpoint reads its budget from `system/rate_limit/<key>`, with a default in `config.xml` and a field in `system.xml`. `Mage_Sales_Helper_Guest`, `Mage_Downloadable_DownloadController`, `Mage_Shipping_TrackingController` and `Mage_Contacts_IndexController` all do that. `system/rate_limit/reset_password` already exists with a default of 10, and `Mage_Customer_Api_CustomerProcessor::resetPassword` already spends it, so the reset token now has two budgets, one tunable and one fixed. A merchant behind a proxy that reports one shared client address cannot lower or disable the new limit. The fix needs new keys in `Mage/Customer/etc/config.xml` and `system.xml`, which this PR does not touch.

**Advice**

- `Mage/Adminhtml/controllers/IndexController.php:483`: declare `_checkResetPasswordLinkToken` as `protected`. Every other underscore-prefixed method in this controller and in `Mage_Adminhtml_Controller_Action` is protected, and a subclass cannot reuse a private check.
- `Mage/Customer/Helper/Data.php:578`: the docblock still says the token is based on the customer id, but the body ignores `$customerId`. Delete the sentence, or drop the parameter and update the two call sites.
- `tests/Frontend/Integration/Customer/AccountTokenRateLimitTest.php`: no test locks the round 1 fix. Red: add `expect(tokenLimiter()->attempts())->toBe(Mage_Customer_AccountController::TOKEN_RATE_LIMIT_MAX_ATTEMPTS);` after the final dispatch in each of the three "after too many failed attempts" tests. Against the code before the fix `attempts()` returns 11 and the assertion fails.
- `tests/Backend/Unit/Adminhtml/Model/Config/EncryptedPathsTest.php:73`: the test cannot fail. `Store::setConfig()` writes a node with no backend model, so `getConfig()` returns the ciphertext and the store never holds `$secret`. Red: assert the rendered text equals `[][]`. Before the fix the render is `[<stored value>][<stored value>]`.
- `tests/Backend/Unit/Admin/Model/VariableEncryptedPathTest.php:45`: same problem. Red: assert the rendered text equals `[]`.
- The admin reset limiter has no test. Red: call `_validateResetPasswordLinkToken()` ten times with a wrong token, then once with the valid token, and expect `Mage_Core_Exception`. Before the PR the eleventh call returns without an exception.
- The website and store branch of `_collectEncryptedConfigTreePaths()` has no test. Only the `default` branch runs with core data. Red: extend the config with a `<websites><base>...` node that carries the encrypted backend model, then expect the path in the list. Before the PR the list came from `system.xml` only, so the path is absent.
- `tests/Backend/Unit/Adminhtml/Model/Config/EncryptedPathsTest.php:18` pins eight paths that belong to `Mage_Usa` and `Maho_Paypal`. Keep one path per source kind instead: one `config_path` case, one `config.xml` case, one plain `system.xml` case.

**Rejected**

- The four copies of `bin2hex(random_bytes(16))` should call `uniqHash()`. The PR test reads each method body and requires an inline `random_` call, so a delegation would turn it red.
- `count($parts) === 3` in the exploded branch drops paths. All 26 collected paths have three segments, and no caller outside the new test passes `true`.
- `Mage_Admin_Helper_Variable::isEncryptedPath()` duplicates `Mage_Adminhtml_Model_Email_PathValidator::isEncryptedNodePath()`. That method is private, and the public `isValid()` has a different contract.
- The global function names in the new frontend test file will collide. No collision exists, and unprefixed names are the norm in that suite.
- `throw new Exception` for a visitor message breaks the project rule. The same file already used that pattern at nine places before the PR.
- The two new docblocks are too long. Both carry a reason the code cannot show.
- Deleting the downloadable entry from the token test loses coverage. `tests/Backend/Unit/Downloadable/LinkHashTest.php` already calls `generateLinkHash()` 50 times and checks the format and the uniqueness.

**Noticed, not fixed**

- `Mage/Newsletter/Model/Subscriber.php:278`: `randomSequence()` repeats the character picker of `Mage_Core_Helper_Data::getRandomString()` and keeps a dead `$par` variable. Both predate this PR.
- `Mage/Customer/controllers/AccountController.php:1157`: the magic link token check uses `strcmp` where the admin uses `hash_equals`. The PR does not touch that line.
- `Maho/CustomerSegmentation/Helper/Coupon.php:109`: the coupon suffix keeps 6 hexadecimal characters, so the number of possible values does not change. The source is now unpredictable, which is the real gain.

**Checks.** php-cs-fixer, PHPStan level 6 and Rector all pass on the changed files. Two rounds. CI is green on f23684e01: every job passes, including the test suite on all seven database backends.

PR #1382, opened 2026-09-09
Added form key validation to storefront state-changing actions

fballiano:
- Every storefront action that changes state now validates the form key and accepts POST only: recurring profile state/update, billing agreement cancel, customer and guest reorder, cart shipping estimate, shipping method update and coupon, wishlist move-from-cart, wishlist item options, shared wishlist add-to-cart and add-all-to-cart, compare remove and clear, tag save and tag remove, price and stock alert signup, one-page checkout method, and the partial authorization cancel.
- The templates that offer those actions now submit a POST form with the `core/formkey` block, or post through `customFormSubmit()` where the control sits inside another form, instead of a GET link or a `window.location` jump. That also repairs several buttons that pointed at POST-only routes and answered 405: recurring profile, billing agreement cancel, shared wishlist add-to-cart and the cart "Move to wishlist" link.
- Reorder now also honors the `sales/reorder/allow` setting, which the storefront links already checked but the action did not.
- The shipping estimate error escapes the rejected country code where the message enters the cart page, so a value that reaches the session messages cannot carry markup. The country resource model keeps a plain message for the CLI and API callers.
- The wishlist share email links each item to the shared wishlist page, and no longer offers the "Add all items" link: an email cannot carry a form key. The controller no longer passes the unused `salable` variable. A merchant with a saved copy of the old email template must drop the `{{depend salable}}` block.
- The shared wishlist add-to-cart action refuses an item that does not belong to the wishlist named by the sharing code.
- A guest interrupted by a POST-only action returns to the page that held the form after login, whatever the "redirect to dashboard after login" setting says, the same as an interrupted GET already did.
- The product alert `testObserver` endpoint is gone. Nothing referenced it, and it let any logged-in customer run the full alert mail job.
- Public-by-design endpoints are unchanged: product alert unsubscribe links, newsletter unsubscribe, payment webhooks, OAuth endpoints and the API protocol endpoints.
- Frontend and backend tests cover a refused request per action, POST-only routing, the escaped country code, the shared item ownership check, the post-login target, and the templates that must post a form key.

A central storefront form key check is tracked in #1436.

> [!NOTE]

### DIFF
--- a/app/code/core/Mage/Catalog/controllers/Product/CompareController.php
+++ b/app/code/core/Mage/Catalog/controllers/Product/CompareController.php
@@ -96,6 +96,13 @@ class Mage_Catalog_Product_CompareController extends Mage_Core_Controller_Front_
     #[Maho\Config\Route('/catalog/product_compare/remove', name: 'catalog.product.compare.remove', methods: ['POST'])]
     public function removeAction(): void
     {
+        if (!$this->_validateFormKey()) {
+            if (!$this->getRequest()->getParam('isAjax', false)) {
+                $this->_redirectReferer();
+            }
+            return;
+        }
+
         $productId = (int) $this->getRequest()->getParam('product');
         if ($this->isProductAvailable($productId)) {
             $product = Mage::getModel('catalog/product')
@@ -139,6 +146,11 @@ class Mage_Catalog_Product_CompareController extends Mage_Core_Controller_Front_
     #[Maho\Config\Route('/catalog/product_compare/clear', name: 'catalog.product.compare.clear', methods: ['POST'])]
     public function clearAction(): void
     {
+        if (!$this->_validateFormKey()) {
+            $this->_redirectReferer();
+            return;
+        }
+
         $items = Mage::getResourceModel('catalog/product_compare_item_collection');
 
         if (Mage::getSingleton('customer/session')->isLoggedIn()) {
--- a/app/code/core/Mage/Checkout/controllers/CartController.php
+++ b/app/code/core/Mage/Checkout/controllers/CartController.php
@@ -585,6 +585,19 @@ class Mage_Checkout_CartController extends Mage_Core_Controller_Front_Action
         $region     = (string) $this->getRequest()->getParam('region');
         $isAjax     = (bool) $this->getRequest()->getParam('isAjax');
 
+        if (!$this->_validateFormKey()) {
+            if ($isAjax) {
+                $this->getResponse()->setBodyJson([
+                    'success' => false,
+                    'error' => true,
+                    'message' => $this->__('Invalid form key. Please refresh the page.'),
+                ]);
+                return;
+            }
+            $this->_goBack();
+            return;
+        }
+
         try {
             Mage::getModel('directory/country')->loadByCode($country);
         } catch (Mage_Core_Exception $e) {
@@ -642,6 +655,18 @@ class Mage_Checkout_CartController extends Mage_Core_Controller_Front_Action
         $code = (string) $this->getRequest()->getParam('estimate_method');
         $isAjax = (bool) $this->getRequest()->getParam('isAjax');
 
+        if (!$this->_validateFormKey()) {
+            if ($isAjax) {
+                $this->getResponse()->setBodyJson([
+                    'success' => false,
+                    'error' => $this->__('Invalid form key. Please refresh the page.'),
+                ]);
+                return;
+            }
+            $this->_goBack();
+            return;
+        }
+
         if (!empty($code)) {
             $this->_getQuote()->getShippingAddress()->setShippingMethod($code)->save();
             $this->_getQuote()->collectTotals()->save();
@@ -666,6 +691,18 @@ class Mage_Checkout_CartController extends Mage_Core_Controller_Front_Action
     public function couponPostAction(): void
     {
         $isAjax = (bool) $this->getRequest()->getParam('isAjax');
+
+        if (!$this->_validateFormKey()) {
+            if ($isAjax) {
+                $this->getResponse()->setBodyJson([
+                    'success' => false,
+                    'message' => $this->__('Invalid form key. Please refresh the page.'),
+                ]);
+                return;
+            }
+            $this->_goBack();
+            return;
+        }
 
         // Check for empty cart
         if (!$this->_getCart()->getQuote()->getItemsCount()) {
--- a/app/code/core/Mage/Checkout/controllers/OnepageController.php
+++ b/app/code/core/Mage/Checkout/controllers/OnepageController.php
@@ -349,6 +349,10 @@ class Mage_Checkout_OnepageController extends Mage_Checkout_Controller_Action
             return;
         }
 
+        if (!$this->_validateFormKey()) {
+            return;
+        }
+
         if ($this->getRequest()->isPost()) {
             $method = $this->getRequest()->getPost('method');
             $result = $this->getOnepage()->saveCheckoutMethod($method);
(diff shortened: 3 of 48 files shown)


fballiano:
**Shape.** The PR does one job and does it in the right place: the form key check sits in each controller action, the templates post a form or call the shared `customFormSubmit` helper, and no new class, event, table or configuration key appears. The review moved the country code escape to the one view that renders it and replaced a nested form with the existing JavaScript post pattern. The one shape question left is the after-login target: the PR turns many links into POST requests, and the customer session sends a guest who was interrupted by a POST to the dashboard when the store setting says so. That rule lives in `Mage_Customer_Model_Session::authenticate()`, outside this PR.

**Fixed.**
- `app/design/frontend/base/default/template/productalert/product/view.phtml:11`: the alert signup form rendered inside the product page add-to-cart form, so the browser dropped it and the button added the product to the cart. The link now posts through `customFormSubmit` with the form key, as the compare link on the same page does. 3 lines added, 6 removed.
- `app/code/core/Mage/Directory/Model/Resource/Country.php:87` and `app/code/core/Mage/Checkout/controllers/CartController.php:612`: the resource model returned HTML entities to the CLI importer and the JSON API. The escape now happens where the message enters the cart page. The backend test that locked the escaped model message is deleted. The estimate test still covers the page. 1 line changed, 1 reverted, 20 deleted.
- `app/code/core/Mage/Wishlist/controllers/IndexController.php:740`: the share email no longer reads the `salable` variable, but the controller still loaded the item collection to compute it. A merchant with a saved copy of the old email also got an add-all link with an empty href. 1 line deleted.
- `app/design/frontend/base/default/template/catalog/product/compare/sidebar.phtml:32`: the `btn-link` class overrode the legacy `.btn-remove` rule that hides the label and draws the X, so the full label showed in the small box. The class is removed. 1 line changed.

**Advice.**
- `app/code/core/Mage/Wishlist/Controller/Abstract.php:17`: `$_isCheckFormKey` has no remaining override, so delete the property and reduce line 46 to `if (!$this->_validateFormKey())`. The file is outside the PR.
- `app/code/core/Mage/Checkout/controllers/CartController.php:588`: the three new form key failure answers in one file use three JSON shapes; use one shape.
- `app/code/core/Mage/Catalog/controllers/Product/CompareController.php:99`: the AJAX branch answers 200 with an empty body, so the compare popup reloads as if the removal succeeded; return the JSON shape of `couponPostAction`.
- `public/skin/frontend/base/default/src/_components.css:180`: the base theme styles `.btn-link` but not `.btn-remove` outside layered navigation, so the compare sidebar remove button now shows as plain text there. Add `.btn-remove` to that group. Before the PR the same control was an unstyled link.
- `app/design/frontend/base/default/template/sales/order/info/buttons.phtml:16`: the three reorder buttons use two class sets (`btn-link link-reorder` and `link-reorder`), which needs two selector groups in the legacy skin; use one.
- `public/skin/frontend/legacy/default/css/styles.css:942`: the two-line CSS comment narrates the edit; delete it or keep one short line.
- `app/design/frontend/base/default/template/catalog/product/compare/list.phtml:181`: the `alert("ciao")` debug line sits in the function the PR edits; delete it.
- PR body, last bullet: it lists the shared wishlist "Add All to Cart" email link as unchanged, but the diff removes that link and makes the action POST-only.
- Tests: helper names such as `cefkController` and `bafkCreateCustomer` hide their meaning; six files copy the same customer factory and session reset, which belong in `tests/Helpers/`.
- Tests: `tests/Frontend/Unit/Core/StateChangingActionFormTest.php` asserts on template text, so a safe rewrite breaks it; prefer rendering the block and checking the output.
- Missing test, `saveMethodAction`: put one item in the quote, post `method=guest` without a form key, call `saveMethodAction()`. Red: `expect(Mage::getSingleton('checkout/session')->getQuote()->getCheckoutMethod())->toBeNull()`.
- Missing test, tag save: as a logged-in customer with frontend tagging on, post the tag name in the query string with no form key. Red: `expect(Mage::getModel('tag/tag')->loadByName($tagName)->getId())->toBeNull()`. A second case posts the name in the body with a valid key. Red: `->not->toBeNull()`.
- Weak test, `tests/Frontend/Integration/Catalog/CompareFormKeyTest.php:66`: the remove test runs on an empty list, so it passes without the fix. Add the product first. Red: `expect(cmpfkCompareCount())->toBe(1)`. Add a case with `isAjax => 1`. Red: `expect($controller->getResponse()->isRedirect())->toBeFalse()`.
- Weak test, `tests/Frontend/Integration/Wishlist/WishlistFormKeyTest.php:103`: `code => 'nope'` reaches `noRoute` with or without the fix. Use a real sharing code. Red: `expect(Mage::getSingleton('checkout/cart')->getQuote()->getItemsCount())->toBe(0)`.
- Missing test row, `tests/Frontend/Unit/Core/StateChangingActionFormTest.php`: add `'wishlist share email' => ['wishlist/email/items.phtml', ['getSharedWishlistUrl'], '/getSharedItemAddToCartUrl/']`.
- Missing test, AJAX estimate: post `country_id => '<b>x</b>'` with `isAjax => 1` and a valid key. Red: `expect($result['message'])->toContain('<b>x</b>')`, which records that the JSON branch stays raw for `alert()`.
- `tests/Frontend/Integration/Sales/ReorderTest.php:102`: the test sets `sales/reorder/allow` to 0 on the store singleton and never restores it.

**Reverted.**
- `app/code/core/Mage/ProductAlert/controllers/AddController.php:24` and `app/code/core/Mage/Tag/controllers/IndexController.php:28`: round 1 set the post-login target to the referer after `authenticate()` failed, so a guest returned to the product page. Round 2 removed it: the call overwrote the target that the session sets from the "redirect to dashboard after login" store setting, and the existing post-login redirect test locks that setting. The session already returns the guest to the referer when the setting is off.

**Rejected.**
- `testObserverAction` lets any visitor run the alert cron: `preDispatch` requires a logged-in customer, and the weakness predates the PR.
- `compare/list.phtml` and `paygate/form/cc.phtml` duplicate `mahoFetch`: the raw fetch calls predate the PR, and the compare popup has no `form_key` input for `mahoFetch` to read.
- `wishlist/index/allcart` has no POST-only route: pre-existing, and the abstract action checks the form key on any method.

**Needs a decision.**
- Guest interrupted by a POST-only action. The PR turns the alert signup, the tag form, reorder and the other links into POST requests. When a guest submits one and logs in, `Mage_Customer_Model_Session::authenticate()` sends the guest to the dashboard under the default store setting, and the action is lost with no message. Before the PR the GET link replayed after login. The review controller already sets the referer unconditionally, the wishlist controller sends the guest to the wishlist page, and the session honors the store setting. Pick one rule in `authenticate()`, which is outside this PR.
- No central form key check on the storefront. The admin area validates the form key once in `preDispatch`. The storefront now holds about ninety copies of the same five-line guard, and the next POST action can forget it. A check in `Mage_Core_Controller_Front_Action::preDispatch()` with an opt-out list is a follow-up.

**Noticed, not fixed.**
- `app/code/core/Mage/Wishlist/controllers/SharedController.php:86`: `cartAction` loads the item by id and never checks that it belongs to the wishlist named by `code`, so any visitor with a form key can add another customer's wishlist item, with its custom option values, to their own cart. Pre-existing.
- `app/code/core/Mage/ProductAlert/controllers/AddController.php:30`: `testObserverAction` lets any logged-in customer run the full alert mail job at will. Nothing references it. Delete it.
- `app/code/core/Mage/Directory/controllers/CurrencyController.php:13`: the currency switch still changes session state on a GET without a key.

**Checks.** cs-fixer, rector and PHPStan pass on every changed PHP and PHTML file. PHPStan does not analyze `tests/`. Two rounds. CI on 75f6da8bee: green, every check passed.

PR #1385, opened 2026-09-09
Replaced message markup with plain text and structured arguments

fballiano:
Builds on `hardening/output-escaping` (#1383); review that one first.

- Message text is now plain text that may carry `%s` placeholders and nothing else. The values behind those placeholders are data, and `Mage_Core_Block_Messages` escapes each one at render time, so no caller escapes anything and no caller writes markup. This answers the review point that a resource model such as `Mage_Admin_Model_Resource_Acl` had no business calling `escapeHtml`.
- A `\Maho\Message\Link` argument renders as an anchor whose label and url are both escaped, with script identifiers stripped from the url first, so a `javascript:` or `data:` scheme cannot survive. A newline in the text renders as a line break, which is how the API credentials notice lays out its three lines.
- `addErrorText` / `addWarningText` / `addNoticeText` / `addSuccessText` on `Mage_Core_Model_Session_Abstract` and `Mage_Core_Block_Messages` are the entry points. An argument count mismatch is logged and falls back to the raw text rather than throwing inside a renderer.
- The twelve core messages that used to be assembled markup now use these, which deletes their `escapeHtml`/`escapeUrl` calls and the `allowHtml` opt-in. The report notice said "click here" twice, so both of its links are now named. The import result appends its Import button below the message list instead of concatenating that markup into a message.
- The boolean `allowHtml` parameter stays on the existing `add*` methods so third-party code keeps working; core no longer passes it anywhere.


fballiano:
Folded into #1383; keeping the work in one pull request.

PR #1386, opened 2026-09-10
Added a warning for the HTML that a save removes from a content field

fballiano:
A user opens a static block that holds an inline `<svg>` icon. The user saves the block and makes no edit. Every icon is gone. The admin sees only the message "The block has been saved."

Two filters remove content, and both were silent:

- The save sanitizer allows the W3C baseline and removes the rest. This filter removes the icons.
- The rich text editor keeps only the markup that its schema declares.

## The change

`Mage_Core_Model_Input_Filter_MaliciousCode::describeRemoved()` compares the two documents and names each removal. It reports only removals. It reports only the outermost element, because a removed `<svg>` also removes its `<path>`.

- **After a save.** The resource model records the removals. An observer on `model_save_after` shows a notice. This covers CMS blocks, CMS pages, blog posts, products, categories and checkout agreements.
- **Before a save.** A new endpoint `cms_wysiwyg/sanitizePreview` answers the same question. The author can still change the content. A field must carry the url to get the warning, so a field that the save does not sanitize stays quiet. The check attaches to a plain textarea, so it also covers the product description.
- **In the editor.** The editor compares its own result. It stays closed for content that it cannot keep. It asks before it replaces the source, both in a CMS field and in the product popup.

## Security fix: checkout agreements

No filter ran on a checkout agreement. The checkout page prints an HTML agreement without escaping it. The ACL `sales/checkoutagreement` is separate from the CMS one, so an admin with only sales rights could put a script tag on every checkout page.

The save now filters both fields. An agreement that is not HTML is escaped at render, so the save keeps that value unchanged.

## Not included

Newsletter templates and email templates keep their current behaviour. They need `<style>` blocks and table markup that the W3C baseline removes. They already run the filter at preview, which is where the markup can execute.

## Not covered

A paste into the rich text editor. TipTap removes unsupported markup at paste time. None of these checks sees that.


fballiano:
@jparker1986 at the moment inline SVG is not supported by tiptap, we can develop a custom extension but it's not 100% simple cause there are attack vectors that can be injected via SVG so it has to be well thought and it will require more time.

But at least now we should have complete coverage when Maho removes part of the content because of sanitization.

PR #1389, opened 2026-09-11
Added inline SVG support to content fields and fixed the SVG upload validator

fballiano:
An author could not keep an icon, a logo or a diagram in a content field. Two layers removed it: the W3C baseline of the Symfony sanitizer holds no SVG element, so every save removed the graphic, and the TipTap schema had no node for `<svg>`, so the editor removed it as well. PR #1386 made that loss visible; this makes it stop happening.

Both layers now keep the subset in the new `Maho\Security\SvgAllowlist`: geometry, paint, text and animation. The editor reads the same list from the WYSIWYG config, so it keeps what a save keeps. There is no toolbar button. An SVG is source markup, not a text style, so it arrives by paste or through the source view.

Everything on the server goes through the public Symfony configuration API. No fork, no patch, nothing marked `@internal`.

## The upload validator was failing open

The first commit is a separate concern that surfaced while building this. `Mage_Core_Model_File_Validator_Svg` removed a fixed list of dangerous names, so it kept every name the list did not mention. Five payloads passed:

| Payload | What happened |
|---|---|
| An external entity in a document type | **A local file was read into the saved SVG.** `LIBXML_NOENT` turns entity replacement on, not off, and `LIBXML_NONET` does not stop a `file://` read. |
| `<animate attributeName="href" values="javascript:...">` | Kept. The element was not on the list. |
| `<set attributeName="onload" to="...">` | Kept. The attribute check looked only at names starting with `on`. |
| `xl:href` instead of `xlink:href` | Kept. The check compared the prefix as text. |
| `<use xlink:href="https://...">` | Kept. It reads a document from another server. |

Two features reach this code, the favicon and the catalog placeholder, and both write into `media/` where the file is served directly. The validator now uses the shared allowlist, refuses a document that declares a document type, and compares an attribute by its namespace instead of its prefix.

## Notes for a reviewer

- **The stored markup is lower case on purpose.** Symfony writes `viewbox` and `lineargradient`. The HTML5 rules for SVG give the browser the standard spelling again. Do not "correct" this; a mixed-case name in the config matches nothing. A test covers every name in the list.
- **`<title>` cannot work inline.** `HtmlSanitizer` keeps only the configured elements that its internal `HEAD_ELEMENTS` list does not contain, so `allowElement('title')` has no effect and reports no error. `aria-label` names an icon instead. A test pins this so nobody tries again.
- **`image` and `font` are dropped.** The sanitizer compares element names and ignores the namespace, so allowing `<svg>` also allows these two, which SVG uses for a different purpose. An SVG `<image>` reads a document from a URL. `a` is the third such name and is kept, because the baseline already removes `javascript:` from a link.
- **The `EXTRA_ATTRIBUTES` loop moved after the element loop.** `allowAttribute($attr, '*')` applies only to the elements allowed at that moment, so in the old order an SVG lost its `class` and its `aria-label`.
- **Animation needs two value rules, and one alone is not enough.** `attributeName` must name an attribute the allowlist already permits, which removes `href` and `onload`. The value attributes must carry no URL scheme and no reference to another document, which removes a paint value pointing at another server. An animation without `attributeName` changes nothing.
- **The existing tests used `<svg>` as the example of markup a save removes.** They now use `<iframe>`.

## What was measured

- 35 payloads through the real pipeline: 6 that must survive, 29 attacks. Every one reaches a fixed point, `sanitize(sanitize(x)) === sanitize(x)`, which is the property that rules out mutation XSS. Six context-flip payloads are included.
- All 6195 icons in `mahocommerce/icons` survive with no geometry loss.
- All 10 shipped theme SVG files survive, `loading.svg` with its animation.
- The five upload bypasses are closed.
- 161 tests pass across the eight affected suites, 732 assertions.

The Browser suite was not run here, as it needs Playwright. It holds one new test.

> [!NOTE]


fballiano:
## Review of PR #1389

**Shape.** This PR lets an author keep an inline SVG in a content field, and it puts the whole name policy in one class that the HTML sanitizer, the `.svg` upload sanitizer and the TipTap editor all read. One list behind three readers is the right shape for the platform, and the editor receives the list instead of holding a copy, so the two cannot drift. The fixed-point tests target the exact hazard that inline SVG creates, which is HTML5 foreign-content parsing, and I could not break that property in 40 payloads across three rounds. The remaining cost is the reuse itself: the same allowlist now governs the upload validator, so `Mage_Core_Model_File_Validator_Svg` narrows what a merchant may upload. That trade needs a human decision, and it is below.

**Fixed.**
- `lib/Maho/Security/SvgAllowlist.php:81` moved `desc` from `ELEMENTS` to `FILE_ONLY_ELEMENTS`. `<desc>` is an HTML integration point, so the parser puts its children in the HTML namespace and the W3C baseline keeps them, while the editor builds its tree from SVG names alone. The editor warned about a loss that the save does not cause, and removed the markup when the admin confirmed. A `.svg` file is read as XML and has no integration point, so a file still keeps `<desc>`. The move also closes a real hole: with `desc` inline, `<svg><desc></desc></svg>` kept the ``.
- `app/code/core/Mage/Core/Helper/Purifier.php:166` builds the three value filters through `Mage::getModel()` instead of `new`, and throws `Mage_Core_Exception` when a rewrite answers the wrong interface. Every other `Mage_Core_Model_Input_Filter_*` class comes from the factory at all 11 of its call sites; these three silently opted out, so a `<rewrite>` resolved but reached nothing.
- `tests/Backend/Unit/Core/Helper/PurifierTest.php` and `tests/Backend/Unit/Core/Model/File/Validator/SvgTest.php` cover both changes on both sides. Each new test is red against the previous commit.

**Rejected.** Ten findings did not survive verification.
- `SvgLink::SCHEMES` is the platform-wide link scheme policy on an SVG class. The list equals Symfony's own default, so it changes no behavior, and both docblocks state the coupling on purpose.
- The editor does not run the `SvgPaint`/`SvgAnimation`/`SvgLink` value rules client-side. The server is the boundary; content loaded from the database has already passed them, and the only unfiltered content is what the admin just typed.
- `svg.js` has no test for the inline-versus-block choice or for a nested `<svg>`. The sibling TipTap extensions carry no tests at all.
- `Mage_Core_Exception` should be `LogicException`, or should carry a translated message. 73 of the 80 `new Mage_Core_Exception(...)` calls in `app/` pass a plain literal, and the three nearest precedents for a rewrite resolving to the wrong interface all do the same.
- `ATTRIBUTE_SANITIZER_MODELS` should not be public. Three of the file's five other constants are public and read nowhere outside it.
- `Mage::getSingleton()` would match the neighbours better than `Mage::getModel()`. A singleton is shared process-wide, so the rewrite test would need to clear the singleton registry as well.
- The `FILE_ONLY_ELEMENTS` docblock states one reason for three entries, and the `desc` comment leads with the weakest reason. The comment leads with the security reason, not the editor reason.

**Needs a decision.**

The upload validator narrows what a merchant may upload, and it says nothing. `SvgAllowlist::ELEMENTS` is the inline-content list, and `Mage_Core_Model_File_Validator_Svg` now applies it to a standalone `.svg` file. The list holds no `mask`, `filter`, `pattern`, `marker`, `textPath`, `switch` or `use`, and `COMMON_ATTRIBUTES` holds no `mask`, `filter`, `vector-effect` or `paint-order`. The old denylist kept all of them. `validate()` rewrites the file in place and returns with no message, and the root `<svg>` always survives, so the `empty($sanitized)` guard never fires.

A plain Figma wordmark export run through the validator:

```
in : <g filter="url(#filter0_d)"><path .../><rect fill="url(#paint0_linear)"/></g>
     <defs><filter id="filter0_d" ...><feOffset dy="2"/><feGaussianBlur stdDeviation="2"/>...</filter>
     <linearGradient id="paint0_linear">...</linearGradient></defs>
out: <g><path .../><rect fill="url(#paint0_linear)"/></g>
     <defs><linearGradient id="paint0_linear">...</linearGradient></defs>
```

The admin sees "The configuration has been saved." A `<mask>` is worse, because the `mask` attribute goes but the masked rectangle stays and covers the artwork, and a `<use href="#symbol">` shape disappears outright. Two features reach this code, the favicon (`Mage_Adminhtml_Model_System_Config_Backend_Image_Favicon`) and the catalog placeholder, and both write into `media/`.

I wrote the fix and then reverted it on request. It added about 90 lines to `SvgAllowlist`, most of it the 24 filter primitives, which a `<filter>` needs because a filter with no primitives renders the referencing element fully transparent. None of the names is a security need: a `.svg` file is read as XML with no foreign-content parsing, and `<mask>`, `<filter>` and `<pattern>` draw nothing and fetch nothing once `SvgPaint` limits a reference to the same document. The review found three follow-up defects inside that 90 lines, which is fair evidence that the list is not free to maintain.

Three ways out, in rising cost:
1. Leave it. The favicon and the placeholder accept a smaller subset than before, and a merchant with a masked or filtered logo gets a wrong graphic with no message.
2. Report the loss. Compare the sanitized document with the parsed input and name the removed elements in a `Mage_Core_Exception`. This contradicts the current docblock ("Rewrites the file with the part that the policy allows") and would break the four committed tests that assert `sanitizeUpload()` returns a stripped string rather than throwing, so it is a design change, not a fix.
3. Widen the file half. `FILE_ONLY_ELEMENTS` is the seam that already exists for this difference.

**Noticed, not fixed.** An uploaded `.svg` keeps its `style` attribute verbatim, so `style="fill:url(https://evil.example/x)"` makes a store-hosted graphic fetch from a third party. `SvgDocumentSanitizer`'s class docblock already states this. The only actor who reaches it is an admin with System Config access, and CSS in an SVG document cannot run script in a current browser. It predates this PR.

**Checks.** php-cs-fixer and rector pass on the whole tree. PHPStan passes on every changed file; a full-tree run cannot complete in this environment, because its parallel worker is killed for memory on the unmodified tree as well. Three rounds. CI runs the tests.

fballiano:
@jparker1986 SVGs are now supported ;-)

PR #1399, opened 2026-09-12
Booted the application once per test process instead of once per test

fballiano:
Every test rebuilt the whole application in `setUp()` and threw it away in `tearDown()`, about 14 ms per test, most of it reloading the store structure. The Backend suite alone runs 3934 tests.

Tests now share one application. `tearDown()` calls the new `Mage::softReset()`, which clears the registry and the request-scoped state but keeps the configuration tree, the stores and the open connections. A test that writes to the configuration still gets the old full reset, because `Store::setConfig()` edits the shared tree and `saveConfig()` edits the database, and only a rebuild restores either; `Mage_Core_Model_Config` counts its writes so the base test case can tell the two cases apart. The three base test cases now share `Tests\MahoTestCase`, which also boots on the cached configuration: a boot that finds no cache keeps reading the tree it built in memory, where a cache flush inside a test cannot reach it.

Measured locally on SQLite with sample data, same pass/skip counts before and after:

| suite | before | after |
| --- | --- | --- |
| Backend | 482 s / 441 s | 310 s / 315 s / 303 s |
| Frontend | 20 s | 15 s |
| Api | 299 s / 301 s | 304 s / 302 s (HTTP bound) |
| peak RSS | 1603 MB | 1205 MB |

One test changed: `CmsImporterTest` read back a configuration value it had written by flushing the cache, which only worked because the flush made the next section load miss and rebuild the tree from the database. It now calls `resetConfig()` and passes on its own, which it did not before.

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -163,6 +163,35 @@ final class Mage
     }
 
     /**
+     * Drop the state one request built on top of an initialized application.
+     *
+     * The configuration tree, the store objects and the open connections stay,
+     * so this is only equivalent to reset() while nothing has written to the
+     * configuration. Callers watch Mage_Core_Model_Config::getWriteCount() and
+     * fall back to reset() once it moves.
+     */
+    public static function softReset(): void
+    {
+        if (self::$_app === null) {
+            return;
+        }
+
+        $resource = self::$_registry['_singleton/core/resource'] ?? null;
+        self::$_registry = [];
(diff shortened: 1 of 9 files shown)


fballiano:
Closing: the change cuts the Backend suite by about 30% locally (449s to 314s) but only about a minute off the CI job, which is one serial step dominated by Backend and Api together, and it reached that by putting test-only machinery into app/Mage.php, App.php and Config.php. Not worth it on either count.

PR #1401, opened 2026-09-13
Completed the PayPal payment at order placement instead of parking the order in pending_payment

### LINKED ISSUE #1400: PayPal: initialize() creates an unpaid pending_payment order that can never be paid
### Summary

`Maho_Paypal_Model_Method_Abstract::initialize()` places the order in `pending_payment` when the quote payment has a `paypal_order_id` but no capture or authorization ID. Nothing in the module ever completes that order, so it is always unpaid and it stays open indefinitely.

```php
} else {
    $stateObject->setState(Mage_Sales_Model_Order::STATE_PENDING_PAYMENT);
    $stateObject->setStatus('pending_payment');
    $stateObject->setIsNotified(false);
}
```

### Why this order can't be paid

- Placing the order deactivates the quote, so the customer can't retry payment on it. They start a new cart and a new order.
- `approveOrderAction()` normally authorizes or captures before placing the order, and the `CHECKOUT.ORDER.APPROVED` webhook does the same. The webhook skips any PayPal order that already has a Mage order, so it never picks this one up.
- The only way to pay it is an admin online capture (`capture()` falls back to `captureOrder($paypalOrderId)`). That only works while PayPal still holds the order (3 hours by default), and nothing in the admin prompts it.
- There is no cron that cancels stale `pending_payment` orders, so the order sits open with its stock reserved.
- If the buyer retries approval, `approveOrderAction()` finds the quote inactive, loads the existing order by `paypal_order_id` and returns the success page without checking the order was actually authorized or captured.

### This branch has been hit by two unrelated bugs

1. #750: `getOrder()` was called without `purchase_units.payments`, so already-completed PayPal orders had no capture or authorization ID. Paid orders landed in `pending_payment`. The fetch was fixed, but the branch stayed.
2. Custom checkout flows on 26.7+: a theme that submits `checkout/onepage/saveOrder` directly (instead of the review-step button that calls `approveOrder`) reaches `initialize()` with only `paypal_order_id` set. The order is created with no authorization. PayPal shows the order as `APPROVED` with `payments: null`.

In both cases the result was a silent unpaid order. An exception would have made either bug visible straight away.

### Proposed fix

Throw in `initialize()` when there is no capture or authorization ID, the same way it already throws when `paypal_order_id` is missing:

```php
if (!$captureId && !$authId) {
    Mage::throwException(
        Mage::helper('paypal')->__('Please complete the PayPal payment before placing the order.'),
    );
}
```

The quote stays active, the customer can retry, and the error is logged.

If a supported flow really does need to place the order before the payment exists, an alternative is to authorize or capture inside `initialize()`, plus a cron that cancels unpaid PayPal `pending_payment` orders after the PayPal order expires.

Separately, `approveOrderAction()` should check the existing order's payment state (an authorization or capture transaction) before returning success for an inactive quote.

### Environment

- Maho main (also seen on 26.7.x)
- Payment method: `paypal_standard_checkout`, payment action authorize

fballiano:
Fixes #1400.

`initialize()` created an unpaid `pending_payment` order when the quote payment carried only a `paypal_order_id`, and nothing in the module ever completed it. Any flow that skipped `approveOrder` (a theme posting to `saveOrder`, the API, an early `getOrder` without payments) hit this branch.

- `initialize()` now loads the PayPal order, checks that it belongs to this order (invoice id, currency, amount), and authorizes or captures it according to the payment action. A PayPal order that is not approved makes the placement fail, so the quote stays active and the buyer can retry. The `pending_payment` branch is gone.
- The quote binding check moved from the checkout controller into `Maho_Paypal_Helper_Data::assertPaypalOrderMatchesQuoteOrOrder()` so both paths run the same one.
- New test `PlacementCompletionTest` covers authorize, capture, an already completed PayPal order, a not approved order, a foreign order, and an amount mismatch.

> [!NOTE]


fballiano:
@jparker1986 could you try this one?

jparker1986:
Tested on our staging (5859109 + this PR, PayPal sandbox, capture). Placing an order with an unapproved PayPal order is refused and no order is created. A PayPal order whose amount no longer matches the cart is refused too. Both are logged. A browser approve and capture test is to follow

fballiano:
@jparker1986 seems correct right?

jparker1986:
@fballiano Yes. An unapproved PayPal order or a changed cart is refused with no order created, and a normal approve and capture in the browser places the order as processing with the invoice and capture recorded.

Can be closed.

fballiano:
closed or merged? 😅

jparker1986:
Merged would be better ;-)

PR #1405, opened 2026-09-14
Allowed the REST and GraphQL API to create an adjustment-only credit memo

### LINKED ISSUE #1404: REST/GraphQL: cannot create an adjustment-only credit memo (goodwill refund) — empty `items` refunds the whole order
### Summary

The admin can create a credit memo that refunds only an adjustment amount (every item qty set to 0, "Adjustment Refund" filled in). This is the standard way to record a goodwill / partial-value refund without consuming item quantities.

`POST /api/rest/v2/orders/{id}/credit-memos` (and the GraphQL mutation) cannot express this:

- Omitting `items` or sending `items: []` builds `['qtys' => []]`, and `Mage_Sales_Model_Service_Order::prepareCreditmemo()` treats an empty qtys map as "refund every refundable item". The adjustment is then added on top, the total exceeds the refundable balance, and the request fails with `400 Refund amount exceeds the order's refundable balance` — even for £1.
- Sending items with `qty: 0` (what the admin form posts) is rejected with `400 Each item must have qty > 0`.

The only workaround is to refund a whole item and cancel most of it out with `adjustmentNegative`. That marks the item fully refunded, so a later real refund of that item (e.g. the customer returns it after a goodwill gesture) can no longer be recorded.

### Reproduce

Order with one invoiced item, £116.84 paid, nothing refunded:

```
POST /api/rest/v2/orders/{id}/credit-memos
{"adjustmentPositive": 5}
→ 400 Refund amount exceeds the order's refundable balance

POST /api/rest/v2/orders/{id}/credit-memos
{"items": [{"orderItemId": 123, "qty": 0}], "adjustmentPositive": 5}
→ 400 Each item must have qty > 0
```

The model layer handles it fine (dry run, nothing saved):

```php
$svc->prepareCreditmemo(['qtys' => [], 'adjustment_positive' => 5]);
// 1 item, base_grand_total 121.84   <- what the API builds

$svc->prepareCreditmemo(['qtys' => [$itemId => 0], 'adjustment_positive' => 5]);
// 0 items with qty > 0, base_grand_total 5   <- what the admin builds
```

### Cause

`Mage/Sales/Api/CreditMemoProcessor.php` (`buildAndRegisterCreditMemo`, current `main`):

```php
$data = ['qtys' => []];
...
if ($qty <= 0) {
    throw new BadRequestHttpException('Each item must have qty > 0');
}
```

### Suggested fix

Either of:

1. Accept `qty: 0` for items (mirror the admin form), or
2. When `items` is explicitly an empty array and an adjustment is given, fill `qtys` with `0` for every order item so no item is refunded — keeping "items omitted = refund everything" for backward compatibility.

The existing refundable-balance cap and the admin controller's "Credit memo's total must be positive" rule would still apply.

Maho `main` @ 9cab897 (also present in 26.9.0).

fballiano:
Fixes #1404.

`Mage_Sales_Model_Service_Order::prepareCreditmemo()` reads an empty `qtys` map as "refund every item", and `CreditMemoProcessor` always built one. A goodwill refund therefore refunded the whole order and then failed the refundable-balance cap. Sending items with `qty: 0`, the shape the admin form posts, was rejected outright.

- An absent `items` key keeps its meaning: refund every refundable item. A present `items` key now means "refund exactly these", so an empty list refunds no item and leaves an adjustment-only memo. `qty: 0` is accepted.
- The shipping collector refunds the whole remaining shipping charge unless an amount is given, so a £5 goodwill refund on an order with £10 of shipping produced a £15 memo. A memo that refunds no item now defaults `shipping_amount` to 0, and a new `shippingAmount` input (base currency, like `adjustmentPositive`) makes that overridable. The admin form has the same field.
- A memo whose total is not positive is rejected, the same rule as the admin controller, with the same `getAllowZeroGrandTotal()` escape for free orders.
- Item validation now matches `InvoiceProcessor`: each entry must be an object with an `orderItemId` that belongs to the order. An unknown id was silently ignored before.

Six cases in `tests/Api/V2/Write/CreditMemoTest.php` cover this, including the reporter's real complaint: after an adjustment-only refund the order stays refundable and `qty_refunded` stays 0.

Behavior change for the release notes: `items: []` no longer means "refund the whole order".

> [!NOTE]


fballiano:
## Review report

**Shape.** The PR lets both REST and GraphQL create a credit memo that refunds an adjustment alone. It checks every `items` entry against the order, rejects a negative shipping amount, and rejects a memo whose grand total is not positive. The adjustment-only path drives `prepareCreditmemo()` with a quantity map instead of building a credit memo by a second route, so it adds no model, table, or event. One method now reads both transports, which matches `InvoiceProcessor` and the shared `normalizeGraphQlInput()` helper. The shape fits the platform.

**Fixed.**

- `CreditMemoProcessor.php:158` The zero-quantity map put every order item into the memo, including an item that was never invoiced. `Mage_Sales_Model_Order_Creditmemo_Item::calcRowTotal()` then divided by an available quantity of 0, and the request returned HTTP 500. The map now stores an entry only when the quantity is above zero. A key that matches no order item keeps the map non-empty, so `prepareCreditmemo()` adds no item.
- `CreditMemoProcessor.php:44` A `shippingAmount` below zero is now rejected. It wrote a negative `base_shipping_refunded` on the order, which let a later memo refund more shipping than the order charged. On a free shipping order it also made `Mage_Sales_Model_Order_Creditmemo_Total_Tax` divide by zero and return HTTP 500.
- `CreditMemoProcessor.php:24` The REST parse block and the GraphQL parse block were copies, and this PR added the same four lines to both. One method now serves both, through `Maho\ApiPlatform\Processor::normalizeGraphQlInput()`. This is the shape that `InvoiceProcessor` already uses.
- `CreditMemoProcessor.php:46` The new message now uses the "must" form that this file uses for an input rule. The "cannot" form is reserved for a state.

**Advice.**

- `CreditMemoProcessor.php:135` The item loop is the third near copy in this folder. `InvoiceProcessor.php:110-145` and `ShipmentProcessor.php:226-240` hold the other two, and the three already disagree on a non numeric id. See "Needs a decision".
- `CreditMemo.php:41` and `:67` Neither the `Post` description nor the `Mutation` description says that `items: []` now refunds no item. A client that sends that body today gets a full refund and after this PR gets an adjustment alone, with no warning.
- `CreditMemo.php:76` The new `shippingAmount` argument has no description. `_initCreditmemoData()` passes it to `setBaseShippingAmount()`, so it is a base currency amount, and the shipping collector reads it as a value that includes tax when `displaySalesShippingInclTax` is on.
- `CreditMemoProcessor.php:115` and `:194` Two comment blocks are longer than the fact they carry.
- `CreditMemoProcessor.php:60` and `:107` Each of the two private methods now takes seven positional parameters. Every new API field adds one more to both.
- `CreditMemoProcessor.php:190` The refundable cap reads `base_total_paid`, which is 0 for an order paid in full by a gift card. The cap therefore limits nothing there. No money leaks, because `Maho_Giftcard_Model_Total_Creditmemo:54` caps the refund at the gift card amount the customer paid, and an inflated adjustment still trips the cap on the part that the gift card does not cover. The comment above the cap overstates what it covers.

Missing tests, with the red line for each:

- A negative `shippingAmount`. Post `shippingAmount: -5` and expect 400. Without the new rule the memo total grows by the negative amount, or the request returns 500 on a free shipping order.
- A partly invoiced order. Refund an adjustment alone on an order whose second item was never invoiced, and expect 201. Without the quantity filter the request returns 500.
- `shippingAmount` on a flat rate order. Post every item at full quantity plus `shippingAmount: 0`, and expect `baseShippingAmount` 0. Without the field the shipping collector refunds the whole charge.
- The GraphQL mutation. Call `createCreditMemo(input: {orderId: <id>, items: [], adjustmentPositive: 5})` and expect 5.0. Before the fix the empty list refunded the whole order.
- A bad `items` body. Post `items: "all"`, `items: ["x"]` and `items: [{orderItemId: "abc", qty: 1}]`, and expect 400 for each. Before the fix the first raised a TypeError and returned 500.
- An all-zero quantity list on a flat rate order. Expect `baseShippingAmount` 0. The current all-zero test uses free shipping, so it stays green if the quantity filter is removed.
- `tests/Api/V2/Write/CreditMemoTest.php:80` Assert that the `items` key exists before you assert that it is empty. The `?? []` fallback hides a dropped key.

**Reverted.**

- Round 2 replaced `normalizeGraphQlInput()` with `Request::toArray()`, to reject a JSON body that is not an object. Undone. API Platform rejects a scalar body at `AbstractItemNormalizer.php:279-284` before the processor runs, so there was nothing to reject.

**Rejected.**

- "A quantity above the refundable quantity is not capped." `Mage_Sales_Model_Order_Creditmemo_Item::setQty()` already throws when the quantity exceeds `getQtyToRefund()`.
- "`shippingAmount` needs an `is_numeric` check like `items`." `adjustmentPositive` and `adjustmentNegative` use the same bare `(float)` cast, and an explicit cast never throws under `strict_types=1`.
- "A JSON scalar body now refunds the whole order." API Platform rejects such a body before the processor runs.
- "`OrderMutationHandler::buildProcessReturn()` duplicates the refund rules." The duplication predates this PR.

**Needs a decision.**

- The item validation loop exists three times: `CreditMemoProcessor`, `InvoiceProcessor`, and `ShipmentProcessor`. The three already disagree, because two reject a non numeric `orderItemId` and `ShipmentProcessor` casts it to 0. One shared method on `Maho\ApiPlatform\Processor`, taking a minimum quantity and a per item check, would delete about 60 lines and add about 25. It touches two files that this PR does not touch, so I did not make it.
- `Mage\Sales\Api\GraphQL\OrderMutationHandler::buildProcessReturn()` holds its own copy of the shipping default, the zero quantity filter, the balance cap, and the positive total rule. Two GraphQL paths named `createCreditMemo` therefore behave differently. This predates the PR and the fix is wider than it.

**Noticed, not fixed.**

- A full refund through the API on an order with a partly invoiced item returns HTTP 500. `prepareCreditmemo()` adds the uninvoiced item with quantity 0, and `calcRowTotal()` divides by zero. This is the same crash the PR path hit, on the pre-existing refund-all path, in files this PR does not touch.

**Checks.** `php-cs-fixer`, `phpstan` level 6, and `rector` all report no error on the changed PHP files. Two rounds. CI on `98b12bd846`: 29 checks pass, and "Backport to maintenance branch" is skipped, which is not a failure.

PR #1413, opened 2026-09-15
Kept newsletter, address and tax rule validation messages out of the exception log

### LINKED ISSUE #1412: Newsletter subscribe logs "This email address is already assigned to another user." as an exception
#### What happens
A guest submits the newsletter form (`POST /newsletter/subscriber/new`) with an email address that belongs to a registered customer. The request is correctly refused and the visitor sees:

> There was a problem with the subscription: This email address is already assigned to another user.

The same event is also written to `var/log/exception.log` as an error, with a full stack trace:

```
ERROR: Mage_Core_Exception: This email address is already assigned to another user. in app/Mage.php:616
```

It's an expected validation result, not a fault. But anything that watches the exception log or a Monolog handler (error trackers, alerting) records it as an application error. On our store it happens several times a day.

#### Why
`Mage_Newsletter_SubscriberController::newAction()` throws a `Mage_Core_Exception` for the validation failure (line 45), catches it (line 54) and passes it to the session:

```php
} catch (Mage_Core_Exception $e) {
    $session->addException($e, $this->__('There was a problem with the subscription: %s', $e->getMessage()));
}
```

`Mage_Core_Model_Session_Abstract::addException()` (line 650) always logs:

```php
public function addException(Exception $exception, string $alternativeText): self
{
    Mage::logException($exception);
    $this->addError($alternativeText);
    return $this;
}
```

So every user-facing validation message sent through `addException()` becomes a logged exception. A few other core controllers catch `Mage_Core_Exception` and call `addException()` in the same way.

#### Steps to reproduce
1. Register a customer account with `customer@example.com`.
2. Log out and allow guest newsletter subscription.
3. Subscribe with `customer@example.com` from the footer form.
4. The error message is shown, and a `Mage_Core_Exception` stack trace is added to `var/log/exception.log`.

#### Expected
The visitor sees the message, and nothing is logged, since this is normal validation and not an application error.

#### Suggested fix
In `newAction()` (and similar catch blocks), show `Mage_Core_Exception` messages with `$session->addError(...)` and keep `addException()` for the generic `Exception` branch, where logging is useful.

Separately, the behaviour itself may be worth another look: a registered customer who isn't logged in can't subscribe from the footer at all. They could be subscribed against their account, or asked to log in.

fballiano:
`Mage_Core_Model_Session_Abstract::addException()` always calls `Mage::logException()`. A `Mage_Core_Exception` carries a message meant for the visitor, so passing one to it wrote a stack trace to `var/log/exception.log` for an ordinary validation result. A store with a public newsletter form collected several of those a day, and error trackers reported each one as an application fault.

Five catch blocks now use `addError()` in the `Mage_Core_Exception` branch and keep `addException()` in the generic `Exception` branch. Every other catch block in the code base already works that way. The visitor sees the same message as before.

Closes #1412


fballiano:
## Review report

**Shape**

This PR changes five catch blocks that catch `Mage_Core_Exception`. Each block now calls `addError()` instead of `addException()`, so the visitor still sees the message but the platform no longer writes a stack trace to `var/log/exception.log`. The plain `Exception` branch next to each block still calls `addException()`, so a real fault still reaches the log. The shape fits the platform: `Mage_Core_Exception` carries a message that a person must read, and every other exception marks a fault, so the choice belongs at the catch site that knows the type, not inside the shared session code. The PR adds no new class, no new method, and no new configuration key, and it changes no shared signature. A scan of `app/code` found exactly five blocks with this pattern before the PR, and zero after it.

**Fixed**

`tests/Frontend/Integration/Newsletter/SubscribeValidationLoggingTest.php:112` - the assertion `->not->toBeEmpty()` was always true. `Mage_Core_Model_Session_Abstract::init()` binds the message data by reference into `$_SESSION`, and `Mage::reset()` between tests does not clear `$_SESSION`. The message that the first test adds therefore stayed in the list when the second test ran. A check proved this: with the dispatch call deleted, the second test still passed. The line now asserts the exact text `Invalid subscription confirmation code.`. One line changed, no line added.

**Advice**

`app/code/core/Mage/Adminhtml/controllers/Tax/RuleController.php:128` - the tax rule save block is reachable and has no test. A second rule that reuses the code of an existing rule makes `Mage_Core_Model_Resource_Db_Abstract::_checkUnique()` throw `Code already exists.`. Red line: `expect($appendedExceptionLogText)->not->toContain('Code already exists.');` in a Backend test. It fails against the old code, because `addException()` wrote `$e->__toString()`, and that string starts with the message.

`tests/Frontend/Integration/Newsletter/SubscribeValidationLoggingTest.php:59` - the `beforeEach` block replaces the Symfony session object but never clears `$_SESSION`, so the message list carries over between tests. Add `$_SESSION = [];` there before a third test joins this file.

**Rejected**

- "`dev/log/active` in `beforeEach` leaks into later tests." `Mage_Core_Model_Store::setConfig()` writes only to the in-memory node, and `Mage::reset()` in `tearDown()` discards the whole config tree.
- "The exception log byte size total is unreliable." The two saves run before the measurement, `composer test` passes no parallel flag, and CI sets `MAGE_IS_DEVELOPER_MODE=1`, so a stream handler runs, not a rotating one.
- "The three global test functions can collide." No other file under `tests/` uses those names, and 308 test files declare global helpers the same way.
- "Three of the five changed catch blocks need a test." A `Mage_Core_Exception` cannot reach two of the three. `$ruleModel->delete()` calls no `Mage::throwException()`, and the address path raises `Mage_Eav_Model_Entity_Attribute_Exception`, which extends `Exception`.

**Noticed, not fixed**

- `tests/Frontend/Integration/Newsletter/SubscribeValidationLoggingTest.php:45` - `dispatchNewsletterAction()` repeats the request wiring of `postUnsubscribe()` in `UnsubscribeOneClickTest.php:42`. The fix needs an edit to a file this PR does not touch. The same hand-built wiring appears in 27 other test files, so it is the house pattern today.
- `app/code/core/Mage/Newsletter/controllers/SubscriberController.php:55` - the guest subscribe endpoint answers with two different texts, so a visitor can learn which addresses belong to customers. It runs no rate limiter, although `system/rate_limit/newsletter_subscribe` exists and only the API reads it. Both conditions existed before this PR, and the old log line held no address and no request detail.

**Checks**

Two rounds. Round 2 confirmed no finding. On the changed files, php-cs-fixer, Rector, and PHPStan report no problem. CI on `7fbadb3` is green: all 28 checks pass, including the seven test suites.

PR #1421, opened 2026-09-17
Stopped the admin from creating a 301 when the url key redirect checkbox is cleared

### LINKED ISSUE #1420: Unchecking "Create Permanent Redirect for old URL" still generates a 301 redirect
### Description
When changing a product or category **URL Key** in the admin panel and unchecking the **"Create Permanent Redirect for old URL"** checkbox, the system still creates a permanent 301 redirect in URL Rewrite Management.

### Steps to reproduce
1. Go to **Catalog > Manage Products** (or **Manage Categories**).
2. Open any existing product/category.
3. Update the **URL Key** field with a new value.
4. Uncheck the checkbox **"Create Permanent Redirect for old URL"**.
5. Save the product/category.
6. Navigate to **Catalog > URL Rewrite Management** and search for the old URL key.

### Expected Result
No new 301 redirect should be generated for the old URL key.

### Actual Result
A new 301 Permanent Redirect entry is created in URL Rewrite Management pointing from the old URL to the new URL.

fballiano:
Confirmed: onUrlkeyChanged() looked for the hidden input as the immediate next sibling of the url key field, but the renderer puts a `<br>` between them, so the hidden input stayed disabled and an unchecked checkbox posted nothing. The controller then saw no value, kept the store default and wrote the permanent redirect.

The function now finds the hidden input by the name that it shares with the checkbox. The same code serves the product form and the category form, so both are fixed.

Closes #1420

> [!NOTE]


fballiano:
## Review

**Shape.** This PR fixes one defect and adds one browser test for it. In `public/js/mage/adminhtml/product.js`, `onUrlkeyChanged()` found the hidden "create redirect" input as the immediate next sibling of the url key field. The renderer puts a `<br>` between them, so the hidden input stayed disabled, the form posted no value, and the controller kept the store default and wrote the 301. The PR now finds the hidden input by name inside the same container, which repairs every url key field that uses this renderer instead of one form. The new test drives the real admin form, which is the only layer that can see this defect, and it follows the existing browser test conventions for admin login, secret keys, page waits and fixture cleanup. The shape fits the platform: one small change in shared code, one test beside the tests that already cover the admin, and no new module, helper, event or configuration key.

**Fixed.**

- `tests/Browser/AdminUrlKeyRedirectTest.php:140` - the test set the url key with `fill()`, which assigns the value without key events. The renderer binds `onUrlkeyChanged()` with `onkeyup` and `onchange`, so the handler never ran and the redirect checkbox stayed disabled. Playwright waits for an enabled checkbox before it unchecks one, so the browser wedged and all seven test jobs ran to the 45 minute limit. The test now clears the field and types the new key. 2 lines changed, 0 added.
- `tests/Browser/AdminUrlKeyRedirectTest.php:150` - the save waited for `.content-header h3:visible`, which the product edit page also holds, so the test could read the database before the save reached the server. It now waits for `#productGrid`, which only the product grid page holds. 3 lines changed, 1 added.
- `tests/Browser/AdminUrlKeyRedirectTest.php:171` - the test expected exactly one permanent redirect row. A product save writes one row per store view, and a local run measured four on the sample data install, so the test failed with a correct fix. It now expects more than zero rows. 1 line changed, 1 added.

**Advice.**

- `public/js/mage/adminhtml/product.js:960` - the selector `input[type=hidden][name="${CSS.escape(chbx.name)}"]` can be `input[type=hidden]`. The `<td class="value">` holds exactly one hidden input, and the line above already finds the checkbox with the plain selector `input[type=checkbox]`.
- `tests/Browser/AdminUrlKeyRedirectTest.php:165` - the second test is a guard, not a regression test. With the checkbox checked, the checkbox posts its own value, so the hidden input never matters, and the test passes against the unfixed code too. Only the first test carries the red line.
- `tests/Browser/AdminUrlKeyRedirectTest.php:162` - the first test only asserts that no redirect row exists. That also holds when the save never reaches the server. Red line: add `expect(countUrlKeyRewrites('pest-url-key-off-new'))->toBeGreaterThan(0)`, counting rows in `core/url_rewrite` where `request_path LIKE 'pest-url-key-off-new%'` and `options` is not `RP`. That assertion fails when the click or the client side validation stops the save.
- No test covers the category form, although the PR body says the fix serves it. Red line: create a category under a store root category, type a new `general[url_key]`, clear `general[url_key_create_redirect]`, save, then expect zero rows with `options = 'RP'` for the old key. Against the unfixed code the count is above zero. Two differences matter: the category form sets an html id prefix, so the field has no plain `#url_key` id, and the renderer is skipped when the category parent is the tree root.
- `tests/Browser/AdminUrlKeyRedirectTest.php:150` - the `->wait(2)` call is needless on its own, because the `#productGrid` locator already waits. I kept it. The locator timeout is 5 seconds, the save must refresh url rewrites over four store views, and five sibling browser tests use the same two lines.

**Rejected.**

- The cleanup deletes the products before the admin user, so a failed product delete leaves an active admin account. The same order is the house pattern in `AdminProductSkuValidationTest.php` and `AdminRelatedProductsGridTest.php`.
- `setResources(['all'])` grants more than the test needs. Every other browser test that creates a throwaway admin grants the same.

**Noticed, not fixed.**

- The throwaway admin user fixture is copied in three browser tests: this one, `AdminProductSkuValidationTest.php` and `AdminRelatedProductsGridTest.php`. The bodies match apart from the constants. Removing the copies needs `tests/Pest.php` and two files that this PR does not touch.

**Checks.** `php-cs-fixer` and `rector` report nothing on the changed files. A single file `phpstan` run reports Pest symbol errors, but an untouched sibling test reports the same errors, so they come from the single file run and not from this PR. Three rounds. Three local runs against a fresh test database: the original test timed out at `uncheck()`, the fixed test passes, and the fixed test fails with "4 is identical to 0" against the unfixed JavaScript, which is the red line the PR needed. Continuous integration on `1a89f0f0d3`: all 28 checks pass, including all seven test suites.

Issue #1426, opened 2026-09-17
Migrate the CLI commands to Symfony Console invokable commands


fballiano:
Symfony Console 8.1 (in use since the Symfony 8.1 upgrade) supports invokable commands: an `__invoke()` method with `#[Argument]` and `#[Option]` parameters replaces `configure()` plus `execute()`, and `#[AsCommand]` can sit on a method.

All 86 commands under `lib/MahoCLI/Commands/` still use the `configure()` / `execute()` style. Migrating them removes the option and argument boilerplate and gives typed, validated input for free.

Scope:

- Convert every command in `lib/MahoCLI/Commands/` to the invokable style.
- Update `dev:create-command` so the scaffold it writes uses the new style.
- Move the CLI tests to `CommandTester::run()` with `ExecutionResult` and the `assertCommandFailed()` / `assertCommandIsInvalid()` helpers where they simplify the test.
- Keep command names, options and output unchanged so scripts and cron entries keep working.

PR #1427, opened 2026-09-17
Migrated to Symfony 8.1

fballiano:
Every Symfony constraint moves from `^7.4` to `^8.1` and the `>=8.0` conflict block is gone. The composer plugin is now required as `^5.0` (v5.0.1), the release that accepts `symfony/routing` 8. Maho releases up to 26.9 keep the 4.x plugin line.

Code changes that the upgrade forced:

- `ApiUser` drops its `eraseCredentials()` override, removed from `UserInterface` in 8.0.
- `Kernel::registerBundles()` returns `array` typed against the new `DependencyInjection\Kernel\BundleInterface`.
- The `Url` constraint now requires a top-level domain by default and rejected `http://localhost`. `isValidUrl()` keeps the previous behavior so dev stores stay scannable and OAuth callbacks keep validating. A test covers it.

One 8.1 feature adopted: `Maho\Http\Client::create()` caps the connect phase at 10 seconds through `max_connect_duration`, so an unreachable carrier host no longer stalls a request for the whole idle timeout. Callers can override it. A test covers the default and the override.

I read the 8.0 and 8.1 changelog of every package in use and checked each behavior change against the code; nothing else applies. The invokable console command migration is tracked in #1426.

Verified locally: `composer lint` is green, the API kernel compiles with zero container deprecations, REST v2 and GraphQL routes answer 200, and the queue worker runs.

> [!NOTE]


fballiano:
**Shape.** The PR raises every Symfony constraint from `^7.4` to `^8.1`, removes the `>=8.0` conflict block, and moves the composer plugin to `^5.0`. It applies four small code changes: `ApiUser` drops `eraseCredentials()`, `Kernel::registerBundles()` returns `array`, `isValidUrl()` keeps the Symfony 7 behavior with `requireTld: false`, and `Maho\Http\Client::create()` sets a 10 second connect cap. The shape fits the platform: a dependency upgrade with the minimum code reaction, plus two tests. The reviewers booted the API kernel under Symfony 8.1, sent live REST requests, and checked each 8.0 and 8.1 removal against the code. No defect was found, so the loop made no change.

**Fixed.**
- `tests/Backend/Integration/MahoCLI/CatalogPriceWebsiteOverridesTest.php:68`: Symfony 8.0 removed `Application::add()`. The test now calls `addCommand()`. One word changed, commit 665e68b475.

**Advice.**
- `lib/Maho/Http/Client.php:33`: the connect cap is not forced by the upgrade. It changes every caller of `Client::create()`, and curl counts DNS, TCP and the TLS handshake inside the 10 seconds. The PR body states this on purpose, so this line only records the fact. A separate PR would make a later bisect land on the right change.
- `lib/Maho/Http/Client.php:33`: the comment says the cap stops a stall "for the whole idle timeout". For a caller with a timeout under 10 seconds, such as the admin notification feed (2 seconds) and the contact form (5 seconds), the idle timeout already ends the stall first. Shorten the comment to say that the cap helps the callers with a long or unset timeout.
- `tests/Backend/Unit/Maho/Http/ClientTest.php:6`: delete the `@package Maho_Http` line. No module of that name exists, and the sibling tests for `lib/Maho` code carry no `@package` line.
- `tests/Backend/Unit/Core/Helper/ValidationTest.php:103`: add one assertion for an IP host to the new test. Red: `expect($this->helper->isValidUrl('http://127.0.0.1:8080/'))->toBeTrue()` fails with the Symfony 8.1 default and passes with `requireTld: false`.

**Rejected.**
- `composer.json:129`: claim that the removed conflict block lets composer resolve `symfony/security-core` 7.x next to `security-bundle` 8.1, so `ApiUser` fails at class load. The removed entries were upper bounds, not lower bounds. The lock file pins `security-core` 8.1.6, and the installed `UserInterface` no longer declares `eraseCredentials()`.
- `lib/Maho/Http/Client.php:33`: claim that the comment word "black-holed" breaks a project rule. No rule in AGENTS.md restricts comment wording.
- `tests/Backend/Unit/Maho/Http/ClientTest.php:15`: claim that the reflection on `defaultOptions` errors when tracing wraps the client. Tracing is off in the test environment, and `Mage::reset()` clears the tracer after every test, so the traced client path is unreachable.
- `lib/Maho/Http/Client.php:33`: claim that the cap also limits reads on `NativeHttpClient`. That client is only selected without ext-curl, and `composer.json` requires ext-curl, so a supported install always uses `CurlHttpClient`.

**Noticed, not fixed.**
- `app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php:684`, `:916`, `:1483` and `UpsAuth.php:39` build raw curl handles with no connect timeout. `app/code/core/Maho/SocialLogin/Model/JwksClient.php:43` and `Provider/Facebook.php:45` call Symfony `HttpClient::create()` directly. These six sites miss the new connect cap and the tracing decorator.
- `app/code/core/Mage/Core/Helper/Data.php:1233`: the docblock of `isValidUrl()` does not say that it checks shape and protocol only. Every fetch caller already adds a host allow-list, but the rule is not written down. This gap existed before the PR.

**Checks.** php-cs-fixer, PHPStan and Rector are clean on the changed source files. PHPStan and Rector do not cover `tests/` in the project config. Two rounds: round 1 found no defect, round 2 fixed the CI failure above. CI on 665e68b475 is green on every job, including the seven test suite backends: https://github.com/MahoCommerce/maho/actions/runs/35248559669

PR #1428, opened 2026-09-17
Migrated the CLI commands to Symfony Console invokable commands

### LINKED ISSUE #1426: Migrate the CLI commands to Symfony Console invokable commands

fballiano:
Every command under `lib/MahoCLI/Commands/` now declares its input with `#[Argument]` and `#[Option]` parameters on `__invoke()` instead of `configure()` and `execute()`. The `dev:create-command` scaffold writes the same style, and the CLI tests use `CommandTester::run()` with `ExecutionResult` and the `assertCommand*` helpers.

Command names, options, defaults, descriptions, help and output are unchanged (checked by diffing every command definition before and after). Two deliberate differences:

- Options that took an optional value with a non-false default now require a value when given (`config:get --scope`, `email:queue:clear --status`, `dev:frontend:theme:create --theme`, `install --db_engine`, and a few more). The invokable API cannot express that mode. Defaults are the same.
- Options and arguments typed `int` reject a non-numeric value instead of casting it to zero.
- `paypal:webhook:simulate --resource-id` builds its `SIMULATED-<timestamp>` default in the body instead of the option definition, so the help screen no longer prints a frozen timestamp. `queue:work --index` and `--sleep` default to the integers 0 and 1 instead of the strings; the run-time result is the same.

`ListApiResourcesTest` stays on `execute()`: `run()` uses a test output that rejects the section a `SymfonyStyle` table needs.

Closes #1426


fballiano:
**Shape.** This PR moves every command under `lib/MahoCLI/Commands/` from the `configure()` plus `execute()` pair to a single `__invoke()` method, where `#[Argument]` and `#[Option]` attributes on typed parameters declare the input. It updates the `dev:create-command` scaffold and the CLI tests to the same style. The change is a net deletion of about 486 lines across 95 files, and it removes hand-written option parsing and casts in favour of the framework. The shape fits the platform: it follows the current Symfony Console style, keeps one style for all commands, and adds no new class, helper, or configuration key. Three reviewers built the input definition of all 88 commands before and after the change and compared them: only the differences that the PR body declares remain, plus the two items noted under Advice.

**Fixed.**
- `lib/MahoCLI/Commands/FrontendThemeCreate.php:90`: one branch still read `$input->getOption('package')` while the bound `$package` parameter held the same value. Replaced with `$package` (1 line). It was the last `getOption()` call under `lib/MahoCLI/Commands/`.

**Advice.**
- `lib/MahoCLI/Commands/Shell.php:27`: the command builds `new SymfonyStyle($input, $output)` by hand and uses `$input` for nothing else. Inject `SymfonyStyle $io` and drop `$input`, as the other commands do.
- `lib/MahoCLI/Commands/SysDirectoryRegionsImport.php:89`: `importRegions()` is public, but `__invoke()` is its only caller and only forwards its parameters. Make it private or move the body into `__invoke()`.
- `lib/MahoCLI/Commands/Install.php:48,74`: `$dbPass` and `$adminPassword` are now parameters of `__invoke()` without `#[\SensitiveParameter]`, while `handleForceInstall()` at line 163 marks the same value. Add the attribute to both.
- `lib/MahoCLI/Commands/CreateCommand.php:113`: the scaffold writes a PHP file without an SPDX header (same before this PR). Add the four-line docblock to the template.
- `lib/MahoCLI/Commands/DBConnect.php:15`, `Serve.php:17`, `TranslationsMissing.php:18`: the `OutputInterface` import is no longer used. Delete the three lines.
- `lib/MahoCLI/Commands/FrontendLayoutDebug.php:44`: `SymfonyStyle $io` comes before `OutputInterface $output`, the reverse of every other command that takes both. Swap them.
- `lib/MahoCLI/Commands/AccessibilityScan.php:48`, `FrontendThemeExport.php:67`, `SysDirectoryRegionsImport.php:69`, `CronUnlock.php:36`: a `(string)` cast survives on a typed parameter. A default of `''` removes the cast, at the cost of `[default: ""]` in the help text.
- `lib/MahoCLI/Commands/PaypalWebhookSimulate.php:36` and `QueueWork.php:41,53`: two changes the PR body does not list. The `--resource-id` default moved from the option definition into the body, so the help screen no longer prints a frozen timestamp. The `--index` and `--sleep` defaults are now integers `0` and `1` instead of the strings. Both keep the run-time result. Add one line to the PR body.
- `tests/Backend/Unit/MahoCLI/Commands/ListApiResourcesTest.php:25`: the comment says "Stays on execute()", but the command has no `execute()` any more. Reword to "Keeps CommandTester::execute()".
- `AGENTS.md`: add two or three lines under "Other key systems" that show the `__invoke()` shape with one `#[Argument]` and one `#[Option]`, so the next command follows it.
- Test, new `tests/Backend/Unit/MahoCLI/CommandDefinitionTest.php`: no test reads the built definition of any command, so a renamed option or a lost shortcut turns nothing red. Build the definition of every class under `lib/MahoCLI/Commands/` and compare it with a committed fixture. Red: delete `name: 'db_host'` from `Install::__invoke()`, the key becomes `db-host` and no longer matches.
- Test, `tests/Backend/Integration/MahoCLI/CatalogPriceWebsiteOverridesTest.php`: no test proves that an `int` option rejects a non-numeric value. Red: `overridesAuditRun(['--limit' => 'abc'])` must throw `InvalidOptionException` with `The value "abc" is not valid for the "limit" option`.
- Test, `tests/Backend/Unit/MahoCLI/Commands/FrontendThemeCreateTest.php`: no test proves that `--parent` now requires a value. Red: `run(['--parent' => null])` must throw `InvalidOptionException` with `The "--parent" option requires a value.`
- Test, new `tests/Backend/Unit/MahoCLI/Commands/CreateCommandTest.php`: nothing runs the scaffold or loads the file it writes. Red: run `dev:create-command`, `php -l` the written file, load the class and run it; delete the file in `afterEach`.

**Rejected.**
- `lib/MahoCLI/Commands/ListApiResources.php:78`, "the SymfonyStyle table forces the test to stay on `execute()`, switch to `new Table($output)`": the call predates this PR, and the plain `Table` uses a different border style, so the fix changes visible output that the PR keeps unchanged.

**Checks.** php-cs-fixer, PHPStan level 6 and Rector report nothing on the 87 changed files under `lib/`. The three Rector hits and the PHPStan hits under `tests/` are outside both tool configs and exist on main. The 112 CLI unit tests pass locally. Two rounds. CI on the last commit: green, all 20 jobs pass on 70865bfbc6.

PR #1432, opened 2026-09-18
Shared one configurable Playwright runtime between the accessibility scanner and future browser-based scanners

### LINKED ISSUE #1416: Share one configurable Playwright runtime between the accessibility scanner and future browser-based scanners
## Context

`Maho_AccessibilityScan` installs its own Playwright and Chromium under `var/accessibility-scan/playwright/` on first use (`Maho_AccessibilityScan_Model_Runner::installPlaywright()`), sets `PLAYWRIGHT_BROWSERS_PATH` to `browsers/` next to it, and skips the install only when `node_modules/.package-lock.json` exists there.

While adding Node and the Chromium libraries to the official Docker image for 26.9 (MahoCommerce/docker-images), a few consequences of that design showed up. They will matter more once a second browser-based tool exists, such as a cookie scanner.

## Observations

- `playwright install chromium` downloads both the full Chromium and the headless shell. Measured on a fresh 26.9.0 install: 641 MB + 340 MB + ffmpeg, about 1 GB under `var/`, not the "roughly 300 MB" the CLI announces. A headless launch only ever uses the headless shell.
- The runtime lives under `var/`, which is often ephemeral in container deployments, so it is downloaded again after every redeploy unless that path is mounted.
- The location and the lock-file heuristic are internal to one module. A host or an image cannot provide a runtime without reproducing that exact tree, and a second module would download its own copy.
- The runtime resolves `playwright@^1` at install time, while `package.json` pins `playwright` 1.62 for the test suite, so the Chromium build a store runs depends on the install date.

## Proposal

One browser runtime helper shared by every scanner:

1. **Configurable location.** Keep the `var/` default for zero-config installs, but honour `PLAYWRIGHT_BROWSERS_PATH` when set and add a config/env setting for the install directory. That lets a host put the runtime on a persistent path, and lets an image ship one through two environment variables.
2. **Detect an existing install.** If a Playwright package satisfying the constraint is already present, globally or at the configured path, use it and skip `npm install`.
3. **Pin the version.** Use the same Playwright version the repository pins for its tests, instead of `^1`, so stores, images and docs agree on one Chromium build.
4. **Choose the browser per scanner.** Headless shell for scans that only read the DOM (accessibility), full Chromium where fidelity matters (a cookie scanner, where consent scripts may treat the shell differently). One flag in the helper; `install chromium-headless-shell` alone saves about 640 MB per store.
5. **One install command** (`accessibility:install` could become the first caller of a generic `browser:install`, or stay as an alias), so a deployment installs the runtime once.

Node itself needs no change: the scanner already finds `node` and `npm` on `PATH`, and the Docker image now provides them along with the Chromium shared libraries.

Happy to help with the implementation or with testing against the Docker image.

fballiano:
Closes #1416.

- New `Maho\Browser\Runtime` owns the Playwright install for every browser-based scanner. The runtime directory comes from `MAHO_BROWSER_RUNTIME_DIR`, then `system/browser/runtime_dir`, then `var/browser-runtime`. `PLAYWRIGHT_BROWSERS_PATH` is honoured, so a host or a Docker image can provide the whole runtime with two environment variables.
- Install detection is content-based (package version, scanner packages, browser build from `browsers.json`), not a lock-file heuristic.
- The Playwright version is the one `package-lock.json` pins for the test suite. The lock file is read from the Maho package directory, so a child project that holds Maho in `vendor/` works the same way as the repository checkout.
- A matching Playwright package in the repository or global `node_modules` is linked into the runtime instead of downloaded again.
- Each scanner declares its browser build and its npm packages through `Maho\Browser\ScannerInterface` and registers under `global/browser/scanners`. The runtime builds one `package.json` from every registered scanner and refuses a runtime directory whose `package.json` belongs to another project. The accessibility scanner uses `chromium-headless-shell`, which saves about 640 MB per store.
- Child processes run through Symfony Process, which the platform already uses elsewhere.
- Node and npm paths moved to System > Configuration > System > Browser Runtime. An upgrade script carries existing values over.
- `sys:playwright:install` replaces `accessibility:install`.

> [!NOTE]


fballiano:
**Shape.** The PR moves the Playwright runtime out of the accessibility scanner into one shared class, `Maho\Browser\Runtime`, that any scanner can use. Scanners register under `global/browser/scanners`, declare their browser and their npm packages, and the runtime keeps one directory, one manifest, and one browser download for all of them. The runtime pins the Playwright version from the lock file inside the Maho package, links an external Playwright copy when the version matches, and runs every child process through Symfony Process. The shape fits the platform: it removes a per-module installer, puts the shared job in `lib/Maho/`, and leaves no second way to install a browser.

**Fixed.**
- `lib/Maho/Browser/Runtime.php:47` read `package-lock.json` from the store root, which a child project with Maho in `vendor/` does not have. It now reads the lock file from the Maho package directory. 1 line changed, 1 comment added.
- `lib/Maho/Browser/Runtime.php:224` looked for `playwright-core/browsers.json` only beside the `playwright` package. A globally installed package keeps it under its own `node_modules`, so the runtime never counted as installed. Added the second lookup. 4 lines.
- `lib/Maho/Browser/Runtime.php:254` `install()` took the install lock and ran `npm root -g` on every command-line scan, including cron. It now returns early when the runtime is complete. 3 lines.
- `lib/Maho/Browser/Runtime.php:507` `install()` overwrote any `package.json` in a configured runtime directory and kept only its dependencies. It now refuses a manifest whose name is not `maho-browser-runtime`. 4 lines plus one CSV line.
- `app/code/core/Maho/AccessibilityScan/Helper/Data.php:158` repeated the browser choice and the package list that the runner already declares. The helper now asks the runner. `Runner::PACKAGES` is gone. 4 lines removed, 2 added.
- `lib/Maho/Browser/Runtime.php:262` read the dependency list back from a file under `var/`, which the web server can write. `install()` now builds the package set from the scanner registry. `readDependencies()` is gone.
- `lib/Maho/Browser/Runtime.php:370` was a hand-written process runner while the platform already uses `symfony/process`. `run()` and `getNodeVersion()` now use Symfony Process with the same environment, timeout and error messages. 80 lines removed.
- The legacy `var/accessibility-scan/playwright` adoption is gone. No store holds that directory.
- `tests/Backend/Integration/AccessibilityScan/ConfigMigrationTest.php:44` orders its rows by `path`, so the strict array comparison no longer depends on row order.
- `tests/Backend/Unit/Maho/Browser/RuntimeTest.php` covers the foreign manifest refusal.

**Advice.**
- `app/design/adminhtml/default/default/template/accessibilityscan/dashboard.phtml:14` builds a `Runtime` itself. Keep one helper method that returns the requirement issues, so the template does not know the library class.
- `app/code/core/Maho/AccessibilityScan/Helper/Data.php:158` builds a second `Runtime` while the runner it creates already holds one. A `isRuntimeInstalled()` on the runner removes the duplicate object.
- `app/code/core/Mage/Core/etc/system.xml:42` puts the node and npm paths in the System section. A backend user with that section can point them at any file, and a scan runs it as the web server user. The capability existed before, but the section is wider now. Consider `local.xml` or the environment only.
- `app/code/core/Maho/AccessibilityScan/etc/config.xml:51` adds an XML registry for scanners while the platform declares observers, cron jobs and routes with attributes. A wrong alias fails silently.
- `lib/MahoCLI/Commands/SysPlaywrightInstall.php:20` is named after Playwright while the class, the config group and the interface say "browser". `sys:browser:install` matches them.
- `lib/MahoCLI/Commands/SysPlaywrightInstall.php:78` reuses the `$browser` option variable for the loop value, so one name holds a string and an enum. Name the loop variable differently.
- `lib/MahoCLI/Commands/SysPlaywrightInstall.php:78` with `--force` and two registered browsers, the loop runs `npm install` once per browser. Not reachable today.
- `lib/MahoCLI/Commands/AccessibilityInstall.php` was removed with no alias. Update the website documentation in the same release.
- `app/code/core/Maho/AccessibilityScan/controllers/Adminhtml/Accessibilityscan/ScanController.php:90` the edited docblock leaves a broken line wrap.
- `app/code/core/Maho/AccessibilityScan/sql/accessibilityscan_setup/upgrade-1.0.0-1.1.0.php:31` tests the value three times. `trim((string) $value) !== ''` is enough.
- Tests, `tests/Backend/Unit/Maho/Browser/RuntimeTest.php`:
  - nested playwright-core: link a package that holds `node_modules/playwright-core/browsers.json` with revision 1700. Red: `expect($this->runtime->getBrowserRevision(Browser::HeadlessShell))->toBe('1700')`.
  - early exit: fixture with the pinned version, the scanner package and the shell build, then `install()`. Red: `expect(is_file($this->dir . '/package.json'))->toBeFalse()`.
  - `syncScript`: copy a source, change it, sync again. Red: `expect(file_get_contents($this->dir . '/scan.mjs'))->toBe('console.log(2)')` and `expect(glob($this->dir . '/scan.mjs.*.tmp'))->toBe([])`.
  - line 137 "does not report its own runtime dir" cannot fail when nothing matches. Point the runtime dir at the repository root and assert `findExternalPlaywright()` is not the repository package.
  - line 71 creates the real `var/browser-runtime`. Record whether it existed and remove it in `afterEach`. Move the node and npm path reset into `afterEach` too.
- `tests/Backend/Unit/Maho/AccessibilityScan/HelperTest.php`: a fixture without `@axe-core/playwright` must make `isRuntimeInstalled()` false. Red: `expect(Mage::helper('accessibilityscan')->isRuntimeInstalled())->toBeFalse()`.
- New tests for the command (`--browser firefox` returns `Command::INVALID`; a complete fixture prints "already installed") and for the dashboard block (`toContain('sys:playwright:install')` on an empty runtime dir).

**Rejected.**
- `app/locale/en_US/Maho_AccessibilityScan.csv` drops "Unable to create directory %s" while the helper still uses it. A row whose key equals its value is never loaded, so the line had no effect.
- `SysPlaywrightInstall.php:78` repeats `npm install` per browser under `--force`. Only one scanner is registered, so the loop runs once.

**Checks.** php-cs-fixer, PHPStan level 6 and Rector are clean on every changed PHP file, and the phtml fixer is clean on the template. The Runtime unit test passes locally (14 tests). Three review rounds, then two follow-up commits (`bfdbfd6bc7`, `eb577a362e`). CI on `5ddc173e04` was green on all 30 jobs, run https://github.com/MahoCommerce/maho/actions/runs/35358070704. CI on `eb577a362e` is running.

